PR #4, opened 2024-10-19
Run engine-dmn through GPT-4 for an example of generated javadoc

datakurre:
This is for discussion. Not for merge.

@javahippie @Noordsestern Regarding https://forum.operaton.org/t/proposal-improving-javadocs/25/1

I took the first project I found, which helps running a code base through GPT for JavaDoc, to get some idea if LLM is any good. I found https://github.com/fynnfluegge/doc-comments-ai, which can be automated to go through a large code base. As such, it calls GPT for every function without existing JavaDoc. It could probably be patched to also generate something for classes and also replace existing docs for function.

The initial impression is the well expected "meh".

* Costs are bearable. Probably less than ten euros for a code base of this size wit GPT-4. And it would take hours in time.
* It does touch code. Unfortunately. I will try later, if I can revert all changes to code lines, because doc comment lines should be distinguishable.
* As such, it also add noise with trivial comments for trivial function.

I didn't try out yet, if it introduced any syntax errors. At first, I should try to revert changes in code. If any of the generated docs look any good, it would be possible to try to iterate this by tweaking "the prompt" with a good example file for better results.


hauptmedia:
I think for review purposes it would be nice if there wouldn't be whitespace/formatting changes in the same commit. Can you control this?

datakurre:
I didn’t try very hard, but I could not get rid of the formatting changes with just “prompt engineering”. I assume that I can remove them automatically locally from the commit. I’ll maybe able to try it out later today.

datakurre:
Surprisingly, I have been unable to find a way with git to only stage and commit changes being part of a java doc comment (e.g. with a regex). I tried to automate interactive staging (`git add -p`) with minimal chunk, but it still leaks some unwanted changes.

Yet, for GPT-4-javadoc evaluation this might be good enough. So, let this be an example of, how GPT-generated docs could look. And with good ideas we could run a few iterations with modified LLM prompt.

javahippie:
Thanks for looking into this and investing your time! I think this can be helpful. Do you think there is a way to prevent the LLM from generating JavaDoc which is too trivial, e.g. for Getters an Setters? I think for most projects there is a sweet spot.

datakurre:
@javahippie Most probably. This has to be done in two phases anyway:

1) GPT phase, where one can “prompt engineer” = guess the right words to get LLM to produce better results

2) git phase, where one can drop all unwanted changes.

I am still looking for a way to automatically stage changes line by line with git. Its default chunking cannot always be splitter to narrow changes enough.

Would you think that engine-dmn is a good enough subset to iterate with? Or would you suggest something else?

hauptmedia:
By the way, I came across this project today: https://github.com/abinthomasonline/repo2txt With really large context windows coming to LLMs in the near future, it might be possible to provide the entire code as context.

kthoms:
@datakurre How to continue here? IMO the approach is promising, but I second @javahippie 's opinion. If you could suppress trivial Javadocs and carefully review the output then we should get a mergeable result.

kthoms:
@datakurre I tend to close the PR in a week when there is no action.

datakurre:
@kthoms I can close this. I used this as a platform for discussion. I'll open a new one, once I am able to apply the feedback.

I had not used LLM in this scale before, so this is also experiment to me. I wanted to see, how well they would work out of the box. As we see, the result was lukewarm.

A couple of open issues besides the quality in general:

* Skip updating trivial methods. Thanks to the naming convention, this could probably be applied already when selecting, which code to pass to LLM and which not to.

* Avoid changes outside docs. It does not seem to be trivial to instruct LLM to NOT touch indentation or code in general. I've also been trying to cope with this with git by simply not staging changes outside docs. But also this has challenges in that git does not allow line by line staging, but has its own chunking algorithm.

PR #34, opened 2024-10-23
Add .editorconfig

### LINKED ISSUE #33: Provide EditorConfig configuration
Contributors should keep the style of existing code. If they use an IDE that supports [EditorConfig](https://editorconfig.org/) then having a configuration will be helpful for them.

Provide a `.editorconfig` file in the project root.

The configuration should suite to the existing code as far as possible. This means, when auto-formatting code with a configured editor, the code should stay unchanged as far as possible.

kthoms:
Fixes #33

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+# General configuration for all files
+[*]
+charset = utf-8
+end_of_line = lf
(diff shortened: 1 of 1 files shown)


javahippie:
Thanks, I will try an check out this branch asap (and test it with some editors)

Issue #41, opened 2024-10-24
[build] Configure maven-enforcer-plugin


kthoms:
**Is your feature request related to a problem? Please describe.**
The enforcer plugin ensures that minimum requirements are met during the build, e.g. Maven and Java version.

When running 
```./mvnw versions:display-plugin-updates```
the build is complaining about the missing plugin configuration:

```
[ERROR] Project does not define required minimum version of Maven.
[ERROR] Update the pom.xml to contain maven-enforcer-plugin to
[ERROR] force the Maven version which is needed to build this project.
[ERROR] See https://maven.apache.org/enforcer/enforcer-rules/requireMavenVersion.html
[ERROR] Using the minimum version of Maven: null
```

**Describe the solution you'd like**
Configure the [maven-enforcer-plugin](https://maven.apache.org/enforcer/maven-enforcer-plugin/usage.html) in the root POM with settings:
- Minimal Maven version 3.9.0 (although earlier would fit)
- Java version 11 (that's current setting)
- groupId `org.operaton.*`

This should be enough for the beginning.

a-m-zill:
@kthoms i think we can close this. I have approved your PR and merged. Thx again.

PR #42, opened 2024-10-24
[#41] Add maven-enforcer-plugin

kthoms:


a-m-zill (approved):
Thx @kthoms that is good. 🙂

PR #48, opened 2024-10-25
Add modeller compatibility information into main README.md file.

a-m-zill:
I have included a note in the readme so that users have information about the compatibility of the BPMN models. As long as we don't have an Operaton modeller.


javahippie:
lgtm!

PR #68, opened 2024-10-28
[#67] Publish and archive surefire reports

### LINKED ISSUE #67: [build] Publish JUnit results in builds
When a PR causes test failures it is not on first sight which test is failing. You need to dig into the console output to look that up.
Further, it is not visible how many tests have been performed.

Add a step to the build workflow that publishes the JUnit test results.

This is how it could look like in the action using the [junit-report-action](https://github.com/marketplace/actions/junit-report-action):

**Describe alternatives you've considered**
- Upload surefire reports as artifacts would create a zip and attach them to the build. We could consider this additionally.
- More GitHub actions exist that are doing similar things, e.g. [test-summary action](https://github.com/test-summary/action)

kthoms:
- use `junit-report-action` to publish test results and add a check to PR builds
- archive surefire reports

Fixes #67

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,8 @@ on:
 
 permissions:
   contents: read
+  checks: write
+  id-token: write
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


javahippie:
Nice, I thought about this during lunchtime, a nice coincidence :D

kthoms:
There is a permission problem. I had this also in my test before, but then added permissions and it worked.
Bildschirmfoto 2024-10-28 um 14 45 09

Here the same branch is [built on my fork](https://github.com/kthoms/operaton/pull/5):
Bildschirmfoto 2024-10-28 um 14 46 35

kthoms:
I'm not sure how to continue here. The error I'm seeing should not occur with the permissions set.

Here a [link to the appropriate hint](https://github.com/mikepenz/action-junit-report/issues/23#issuecomment-1412597753).

And on my fork this does not appear.

kthoms:
We should see "2 successful checks" here.

kthoms:
Let's try it. If this not working, we'll go the next step. Does not look as if it harms, the build was still green.

javahippie:
I think it worked: https://github.com/operaton/operaton/actions/runs/11562223287

Issue #80, opened 2024-10-29
Sonarqube Analysis step failing


kthoms:
**Describe the bug**
With https://github.com/operaton/operaton/pull/74 Sonarqube analysis has been enabled.

```
Error:  Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:4.0.0.4121:sonar (default-cli) on project operaton-root: Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:4.0.0.4121:sonar (default-cli) on project operaton-root: Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator to check the permissions of the user the token belongs to
```

**To Reproduce**
[example log](https://github.com/operaton/operaton/actions/runs/11583527959/job/32248865826?pr=43) 

**Expected behavior**
Build is green

**Related**
#66

javahippie:
I'm assuming that a PR from a forked Repo is not allowed to access the necessary secrets? "Project not found" can be an authorization issue with Sonarqube. Checking...

javahippie:
The resources in that link won't help us as Sonarqube tells me that this project is too big for automated analysis. I'm afraid the only thing we can do is running the analysis on the main branch only and skipping it for merge requests, which is not ideal.

<img width="814" alt="grafik" src="https://github.com/user-attachments/assets/9a484dcf-8957-47e2-bdea-909d0e8066d1">

javahippie:
Taking the liberty to close the issue, adressed by #81, #82

PR #81, opened 2024-10-30
chore: Disabled Sonar analysis for PRs outside of operaton organisation [#80]

javahippie:


kthoms:
We should consider to enable for PR builds later again, wenn we want to enforce the quality gate.
Please also consider my PR to just allow this step to fail. When Sonar is not able to be executed then this does not harm the build.

javahippie:
I think we can add both for the moment, not failing the build seems reasonable to me

I don't see a way to re-enable the analysis for PR right now. The best we could do is having quality gates for the PRs from branches within the repo and not for external collaborators. To configure this we could check if "SONAR_TOKEN" is accessible in the workflow. If it is, then we run the analysis, if not we skip the whole step and save ~7 minutes of build time.

javahippie:
Added a condition, checking if the workflow is running inside our org

kthoms:
Rename PR title to "Disabled Sonar analysis for PRs outside of operaton organisation"?

PR #104, opened 2024-11-03
chore(parent): Bump Quarkus to Version 3.85

hauptmedia:
Related-to: https://github.com/camunda/camunda-bpm-platform/issues/4346

Backported commit e4c392c18f from the camunda-bpm-platform repository.
Original author: psavidis <69160690+psavidis@users.noreply.github.com>

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -18,7 +18,7 @@
   </description>
 
   <properties>
-    <version.quarkus>3.8.0</version.quarkus>
+    <version.quarkus>3.8.5</version.quarkus>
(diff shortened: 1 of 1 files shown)


kthoms:
I hoped that we get these version bumps by Dependabot.

javahippie:
Currently we have dependabot activated for security issues, only. There are a lot (!) of older dependencies in the repository, so my fear was that Dependabot would swamp us with Pull Requests if we enabled version updates. But it might be a good idea to just try it

PR #106, opened 2024-11-03
[#105] Extend ProcessEngineExtension to enable injection of engine services

### LINKED ISSUE #105: Enable ProcessEngineExtension to inject also the engine's services
The `ProcessEngineExtension` allows to inject the engine instance into fields of type `ProcessEngine` in the test class.
```
@ExtendWith(ProcessEngineExtension.class)</code>
class YourTest {
   ProcessEngine processEngine; // injected
}
```

Often tests need to access the various services or the engine's configuration. To access those, the `processEngine` instance needs to be accessed each time.

The existing test code base declares the various services in `AbstractProcessEngineTestCase` which would need subclassing and a dependency on JUnit 4.

Extend the extension to be able to inject services in the test's fields:
```
@ExtendWith(ProcessEngineExtension.class)</code>
class YourTest {
   ProcessEngine processEngine; // injected
   RuntimeService runtimeService;
}
```

**Additional context**
This would facilitate the transition to JUnit 5 (#27) for test classes that derive from `AbstractProcessEngineTestCase`.

kthoms:
This extends ProcessEngineExtension to be able to inject engine services into fields of the test class. By doing this, tests can be cleaner and just declare a fields of engine services that the test is using.

Fixes #105

### DIFF
--- a/test-utils/junit5-extension/src/test/java/org/operaton/bpm/engine/test/junit5/ProcessEngineExtensionServiceInjectionTest.java
+++ b/test-utils/junit5-extension/src/test/java/org/operaton/bpm/engine/test/junit5/ProcessEngineExtensionServiceInjectionTest.java
@@ -1,0 +1,52 @@
+package org.operaton.bpm.engine.test.junit5;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.operaton.bpm.engine.*;
+import org.operaton.bpm.engine.impl.cfg.ProcessEngineConfigurationImpl;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ExtendWith(ProcessEngineExtension.class)
+class ProcessEngineExtensionServiceInjectionTest {
+    ProcessEngine processEngine;
+    // this is to check that injection will set _all_ fields with the target type
+    ProcessEngine processEngine2;
+    ProcessEngineConfiguration processEngineConfiguration;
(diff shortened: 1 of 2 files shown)


test-utils/junit5-extension/src/test/java/org/operaton/bpm/engine/test/junit5/ProcessEngineExtensionServiceInjectionTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.operaton.bpm.engine.*;
+import org.operaton.bpm.engine.impl.cfg.ProcessEngineConfigurationImpl;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ExtendWith(ProcessEngineExtension.class)
+class ProcessEngineExtensionServiceInjectionTest {
+    ProcessEngine processEngine;
+    ProcessEngine processEngine2;
```
javahippie, line 13: Maybe add a comment why there is `processEngine` and `processEngine2`? I guess you are testing that a class which was referred twice has the same reference, but I stumbled over this and had to check / think for a short time about it

kthoms: Yes, the original logic was to set any field of that type. I think this is a corner case that should not be used. Maybe it is even better to enforce that only one field of that kind exists?
I have added the comment.

Issue #85, opened 2024-10-31
Incorporate new language features (Java 17)


javahippie:
**Is your feature request related to a problem? Please describe.**
Right now there are constructs and patterns in the code which can be solved more elegantly and concise with Java 17 language features. An overview can be found here: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&tags=java14,java15,java16,java17&id=operaton_operaton

**Describe the solution you'd like**
We should identify code which can be improved by the new features and commit some changes, one after another.

This is a good issue for people to get started with their contributions. Pull Requests would ideally only consider one module at the time, to keep the changes small enough for reviews

kthoms:
@javahippie There are no issues left by the Sonar query. Anything specific left?
There is potential to leverage text blocks, pattern matching, records and quite some more. But nothing concrete actionable ATM.
I would tend to close this one now and create more precise new issues.

javahippie:
@kthoms Would agree to that. I think text blocks are a low hanging fruit for better readability, while record only make sense for classes which are not publicly accessible and could be considered "API". New, separate tickets make a lot of sense here

kthoms:
Closing as done.

PR #119, opened 2024-11-04
chore: Fix instanceof check and cast with pattern matching

dirk-olmes:
Fixing all Sonar issues of type S6201 in module engine. Part of #85


javahippie:
Hi @dirk-olmes, thanks for your first contribution! I activated the pipeline-run for your pull requests (won't be necessary in the future, just a one-time-thing), so the whole test suite will run. So far it looks good to me!

PR #150, opened 2024-11-06
chore(quarkus): Bump quarkus version to 3.15

hauptmedia:
Related to https://github.com/camunda/camunda-bpm-platform/issues/4580

Backported commit 64274723bb from the camunda-bpm-platform repository.

### DIFF
--- a/parent/pom.xml
+++ b/parent/pom.xml
@@ -18,7 +18,7 @@
   </description>
 
   <properties>
-    <version.quarkus>3.8.5</version.quarkus>
+    <version.quarkus>3.15.0</version.quarkus>
     <version.spring.framework>5.3.39</version.spring.framework>
     <version.spring.framework6>6.1.13</version.spring.framework6>
     <version.spring-boot>3.3.5</version.spring-boot>
--- a/quarkus-extension/README.md
+++ b/quarkus-extension/README.md
@@ -26,18 +26,18 @@ can look like the following:
 
 ```properties
 # process engine configuration
-quarkus.operaton.cmmn-enabled=false
-quarkus.operaton.dmn-enabled=false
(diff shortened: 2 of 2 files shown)


hauptmedia:
This fails with the error messagea: `Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (default-compile) on project operaton-bpm-quarkus-engine: Fatal error compiling: java.lang.IllegalStateException: This module is configured to use @ConfigMapping annotations but we found a @ConfigRoot without a corresponding @ConfigMapping annotation in: org.operaton.bpm.quarkus.engine.extension.OperatonEngineConfig. Either add the annotation or add the -AlegacyConfigRoot=true argument to the annotation processor config in the pom.xml -> [Help 1]
`
Should we enable the legacyConfigRoot Option? What do you think @kthoms?

javahippie:
@hauptmedia There is a commit missing, the @ConfigMapping was added in the bump to quarkus 3.14 here:

https://github.com/camunda/camunda-bpm-platform/commit/816b58fd91e578d1ec7e4992300cdcd114ac3e5b

javahippie:
I'm not sure how much effort it is to merge those together, if you backport the referenced commit with the porting-tool, I can merge that one first and a rebase on this branch should resolve the issue

hauptmedia:
the porting tool cannot do it fully automatically, so i will just add the patch file manually to this branch. No big deal. I'm now quite experienced doing this stuff 😂😂

hauptmedia:
i skipped the 3.14 commit because i thought it's simple version bump that is superseeded by the next version bump. But that is not always the case. Every commit has to be check. I noted every commit I skipped, will check all again if I missed something

PR #168, opened 2024-11-08
Fix instanceof check and cast with pattern matching in juel. (#85)

dirk-olmes:
Fix instanceof check and cast with pattern matching in juel. (#85)


javahippie:
Thanks again! :)

javahippie:
On a different notice, I'm not sure why GH still does not list you as a contributor and marks you as a "First Time contributor" on this PR 🤔

dirk-olmes:
@javahippie no need to thank me again - more patches like these will come in soon.

Maybe I'm still a first time contributor because I recreated the branch after my first PR? It occurred to me only later that reusing the same branch for multiple PRs is possible.

Issue #176, opened 2024-11-08
Reduce build warnings


kthoms:
**Is your feature request related to a problem? Please describe.**
The build log contains a couple warnings, e.g.
```
Warning:   Parameter 'useDefaultManifestFile' (user property 'jar.useDefaultManifestFile') is deprecated: For version 3.0.0 this parameter is only defined here to break the build if you use it!
```
Not all of them will be resolvable, others will do.

**Describe the solution you'd like**
Try to solve as many build warnings as possible. Go through a typical build log for the `main` branch (see [Actions](https://github.com/operaton/operaton/actions)) and PRs.

It is expected that the build just works as before, but with less warnings.

kthoms:
I did not find a possibility yet to resolve the `useDefaultManifestFile` warnings.

The property is defined in the parent POM with `false`. Removing it will change it to `true`, although its default value is `false`. I did not see yet why it is set to true then.

Without having the configuration in the parent POM the build would break.

javahippie:
I have [a suspicion](https://github.com/operaton/operaton-release-parent/blob/02ffca5be1bdf54d37cc692a771c4855dcc1c1c4/pom.xml#L73)

kthoms:
I had a deeper look at the warnings from the assembly plugin like

```
[WARNING] Cannot include project artifact: org.operaton.bpm.tomcat:operaton-bpm-tomcat:pom:1.0.0-beta-5-SNAPSHOT; it doesn't have an associated file or directory.
```

These warnings can't be resolved. The only way would be filtering the logs. But this isn't worth the effort.

PR #178, opened 2024-11-08
chore(build): Define property 'surefire.argLine' (#176)

kthoms:
This change makes the JVM arguments used by surefire more flexible.

In `parent/pom.xml` a property `surefire.argLine` is defined. This property is now used in any places that require additional customization of the JVM args.

Often customization is required for adjusting memory settings. The memory settings are now defined with a default value in the property `surefire.memArgs` in the parent POM. 

Modules that need customization of memory now just override the `surefire.memArgs` instead of overriding surefire's `argLine` property. This way we make sure that the surefire default arg line is always taken.

The original intent was to get rid off the warning:
```OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended```

Therefore I wanted to append a setting to the default `surefire.argLine`. However, this did not work out (e.g. tried with `-Xshare:off` or `-XX:-PrintWarnings`).

Anyway, IMO this setting cleans it up a bit.

**Testing Instructions**

You can see that the settings are applied by looking at the effective POM, or by enabling debug output with `-X` in the execution.

Example: `engine-rest` sets different memory settings than default and appends `file.encoding`:
```
mvn help:effective-pom -f engine-rest/engine-rest | grep -H -n "<argLine>"
(standard input):2364:              <argLine>-Xmx2g -Duser.language=en -Duser.region=US -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/Users/karstenthoms/Development/projects/operaton/engine-rest/engine-rest/target/heap_dump.hprof -Dfile.encoding=ISO-8859-1</argLine>
(standard input):2607:                  <argLine>-Xmx2g -Duser.language=en -Duser.region=US -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/Users/karstenthoms/Development/projects/operaton/engine-rest/engine-rest/target/heap_dump.hprof -Dfile.encoding=ISO-8859-1</argLine>
(standard input):2705:                  <argLine>-Xmx2g -Duser.language=en -Duser.region=US -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/Users/karstenthoms/Development/projects/operaton/engine-rest/engine-rest/target/heap_dump.hprof -Dfile.encoding=ISO-8859-1</argLine>
(standard input):2795:                  <argLine>-Xmx2g -Duser.language=en -Duser.region=US -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/Users/karstenthoms/Development/projects/operaton/engine-rest/engine-rest/target/heap_dump.hprof -Dfile.encoding=ISO-8859-1</argLine>
```

### DIFF
--- a/distro/wildfly/subsystem/pom.xml
+++ b/distro/wildfly/subsystem/pom.xml
@@ -12,6 +12,10 @@
 
   <name>Operaton - Wildfly Subsystem</name>
 
+  <properties>
+    <surefire.memArgs>-Xmx512m</surefire.memArgs>
+  </properties>
+
   <dependencies>
 
     <dependency>
@@ -93,7 +97,6 @@
         <configuration>
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
           <enableAssertions>true</enableAssertions>
-          <argLine>${jacoco.argLine} -Xmx512m</argLine>
           <systemProperties>
             <property>
               <name>jboss.home</name>
--- a/distro/wildfly26/subsystem/pom.xml
+++ b/distro/wildfly26/subsystem/pom.xml
@@ -12,6 +12,10 @@
 
   <name>Operaton - Wildfly 26 Subsystem</name>
 
+  <properties>
+    <surefire.memArgs>-Xmx512m</surefire.memArgs>
+  </properties>
+
   <dependencies>
 
     <dependency>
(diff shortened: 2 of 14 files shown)


kthoms:
Not working as expected so far. Will continue later.

PR #183, opened 2024-11-09
chore(ci): make frontend build skippable

hauptmedia:
original commit message: reintroduce oracle jdk stages (#4610)

related to https://github.com/camunda/team-automation-platform/issues/52

Backported commit 47c0504b58 from the camunda-bpm-platform repository.
Original author: tasso94 <3015690+tasso94@users.noreply.github.com>

### DIFF
--- a/engine-rest/docs/pom.xml
+++ b/engine-rest/docs/pom.xml
@@ -53,6 +53,7 @@
         <artifactId>frontend-maven-plugin</artifactId>
         <configuration>
           <outputdir>${project.build.directory}</outputdir>
+          <skip>${skip.frontend.build}</skip>
         </configuration>
         <executions>
           <execution>


hauptmedia:
This was originally added in for this commit https://github.com/camunda/camunda-bpm-platform/commit/47c0504b58
and then used in the build pipeline. We don't have the Jenkin based build pipeline but I thought perhaps it might be useful in the future to be able to skip the frontend build

javahippie:
I'd prefer to think about this one, first. Maybe it would be a nice idea to have a Maven build profile which skips the Frontend builds entirely, including all dependencies and tests in qa, I'd prefer that over a flag.

hauptmedia:
No need to rush this one through – in fact, we could close it for now and start a discussion instead. There are no other commits depending on this. We (as a user of this project) have the use case to create a build without the embedded web apps, and it would be fantastic if we could achieve this through a Maven profile. Let’s explore options and see what might work best

javahippie:
On the other hand, there is no harm in including this now and adding something better later :)

PR #188, opened 2024-11-11
Fix instanceof check and cast with pattern matching in engine. (#85)

dirk-olmes:


kthoms (approved):
Thanks Dirk, lgtm;
Keep on going, looking forward for more PRs ;)

Issue #84, opened 2024-10-30
Tests not executable from IDE


kthoms:
**Describe the bug**
It is not possible to run unit tests from within the IDE. I'm using IntelliJ, but have seen this with Eclipse, too.

When the IDE compiles the projects before launching tests, it can't resolve classes from package `org.operaton.bpm.impl.juel.jakarta.el`.

**To Reproduce**
1. Open the project in IntelliJ
2. Open a unit test, e.g. `MBeanServiceContainerTest`
3. Press Play Button

Alternatively a similar issue can be produced running Open Rewrite from Maven CLI:
```
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.activeRecipes=org.openrewrite.java.RemoveUnusedImports -Drewrite.exportDatatables=true
```

This will fail on module `operaton-engine`:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:testCompile (default-testCompile) on project operaton-engine: Compilation failure: Compilation failure: 
[ERROR] /Users/kthoms/Development/git/github/operaton/engine/src/test/java/org/operaton/bpm/application/impl/el/CalledProcessApplication.java:[25,45] package org.operaton.bpm.impl.juel.jakarta.el does not exist
[ERROR] /Users/kthoms/Development/git/github/operaton/engine/src/test/java/org/operaton/bpm/application/impl/el/CalledProcessApplication.java:[39,13] cannot find symbol
[ERROR]   symbol:   class ELResolver
[ERROR]   location: class org.operaton.bpm.application.impl.el.CalledProcessApplication
[ERROR] /Users/kthoms/Development/git/github/operaton/engine/src/test/java/org/operaton/bpm/application/impl/el/CallingProcessApplication.java:[37,13] cannot find symbol
[ERROR]   symbol:   class ELResolver
[ERROR]   location: class org.operaton.bpm.application.impl.el.CallingProcessApplication
[ERROR] /Users/kthoms/Development/git/github/operaton/engine/src/test/java/org/operaton/bpm/engine/test/standalone/el/ExpressionBeanAccessTest.java:[26,45] package org.operaton.bpm.impl.juel.jakarta.el does not exist
[ERROR] /Users/kthoms/Development/git/github/operaton/engine/src/test/java/org/operaton/bpm/engine/test/standalone/el/ExpressionBeanAccessTest.java:[69,43] cannot find symbol
[...]

PR #195, opened 2024-11-13
chore(deps): Remove jakarta-el shading (#84)

kthoms:
This change removes shading of the `jakarta-el` library in module `operaton-juel`.

The shading of `jakarta-el` caused problems in IDEs, since the shaded classes are required at compile time, but are only produced at build time. Also, the IDE does not resolve references against the shaded build result, but against the module sources.


kthoms:
Test executed within IDEA:

Bildschirmfoto 2024-11-13 um 17 13 46

javahippie:
Thanks a lot, that's a big change with a big effect!

Just to be clear about the ramifications: This means, that the next Operaton release would be incompatible with older Java EE / Jakarta EE versions and Spring 5 / Spring Boot 2, right? (Not that it's a bad thing, just to know what to put into migration guides, release notes...)

kthoms:
Yes, this would be the consequence. The question arising is of course with which App-Servers, Spring and Java version Operaton should be compatible with. This is up to you, what direction Operaton should go.

IMHO Operaton should not keep compatibility with ancient API. Who should maintain that? Likely there is not enough development power, and likely not enough need.

Maybe the issue should be labeled as breaking.

For the case that this here would be merged, then it is definitely something for the release documentation and migration path.

javahippie:
I would agree to this. Is it okay for you to start a short discussion on the forum? Then others have the chance to chime in

hauptmedia:
I copy/pasted the whole disussion to chatgpt and let it create a forum post from it => https://forum.operaton.org/t/dropping-legacy-java-ee-and-spring-5-spring-boot-2-support-in-operation/67

javahippie:
I have not seen any objections to this, yet. My proposal would be to merge this after the weekend to give people more time to read and address this issue

javahippie (approved):
There were no objects from the community, I'd be okay with merging this today

PR #197, opened 2024-11-14
chore(testing) JUnit5 migration: submodules /model-api

kthoms:
This change migrates the test code base of
- `model/bpmn-model`
- `model/bcmn-model`
- `model/dmn-model`
- `model/xml-model`
to JUnit 5.

Changes:
- Performed initial migration with Open Rewrite
- Replaced Hamcrest by AssertJ with Open Rewrite
- Manual migration, especially of used JUnit 4 rules. Some tweaking was required for `TestModelTest` and its subclasses, since this `@Parametrized` test used instance setup before each test execution with parameters, but JUnit 5 does not support `@BeforeEach` with parameters.
- Translated parametrized tests
- Dependency management

### DIFF
--- a/model-api/bpmn-model/src/test/java/org/operaton/bpm/model/bpmn/DefinitionsTest.java
+++ b/model-api/bpmn-model/src/test/java/org/operaton/bpm/model/bpmn/DefinitionsTest.java
@@ -15,14 +15,13 @@
  * limitations under the License.
  */
 package org.operaton.bpm.model.bpmn;
-import org.junit.jupiter.api.Assertions;
+
 import org.junit.jupiter.api.Test;
-import org.operaton.bpm.model.bpmn.instance.*;
 import org.operaton.bpm.model.bpmn.instance.Process;
+import org.operaton.bpm.model.bpmn.instance.*;
 import org.operaton.bpm.model.bpmn.util.BpmnModelResource;
 import org.operaton.bpm.model.xml.ModelParseException;
 import org.operaton.bpm.model.xml.ModelReferenceException;
-import org.operaton.bpm.model.xml.ModelValidationException;
 import org.operaton.bpm.model.xml.impl.util.IoUtil;
 
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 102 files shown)


model-api/bpmn-model/src/test/java/org/operaton/bpm/model/bpmn/DefinitionsTest.java
```diff
@@ -138,12 +133,7 @@ public void shouldNotAffectComments() throws IOException {
     definitions.getImports().add(importElement);
 
     // validate model
-    try {
-      Bpmn.validateModel(bpmnModelInstance);
-    }
-    catch (ModelValidationException e) {
-      Assert.fail();
-    }
+    Bpmn.validateModel(bpmnModelInstance);
```
javahippie, line 136: Does JUnit 5 fail explicitly upon thrown exceptions? This case here is handled differently from the case in line 113 and I don't quite understand why

model-api/bpmn-model/src/test/java/org/operaton/bpm/model/bpmn/DefinitionsTest.java
```diff
@@ -226,12 +213,7 @@ public void shouldAddMessageAndMessageEventDefinition() {
     startEvent.getEventDefinitions().add(anotherMessageEventDefinition);
 
     // validate model
-    try {
-      Bpmn.validateModel(bpmnModelInstance);
-    }
-    catch (ModelValidationException e) {
-      Assert.fail();
-    }
+    Bpmn.validateModel(bpmnModelInstance);
```
javahippie, line 216: Same here as in line 136

model-api/bpmn-model/src/test/java/org/operaton/bpm/model/bpmn/DefinitionsTest.java
kthoms: I have changed, because I did not see a sense on catching and failing. There is a small difference, but in the end both cases would make the test fail.

ModelValidationException is a RuntimeException. If it occurs, it would be catched by the test engine and make the test fail. That is the intention.

But yes, L133 I have overseen that and I will make it the same.

To simulate that I have made the validation fail:
Screenshot 307

This is how it would be handled before:
Screenshot 308

And this is now:
Screenshot 309

model-api/bpmn-model/src/test/java/org/operaton/bpm/model/bpmn/DefinitionsTest.java
kthoms: I have done this refactoring intentionally.

Issue #27, opened 2024-10-22
Migrate test code to JUnit 5 / AssertJ


kthoms:
**Is your feature request related to a problem? Please describe.**
The test code base is still on JUnit 4. Most modern frameworks use JUnit 5 nowadays.

**Describe the solution you'd like**
The test code should be migrated to JUnit 5. Test assertions should use AssertJ, especially to replace Hamcrest.

Open Rewrite could be helpful here, it supports [refactoring from JUnit 4 to 5](https://docs.openrewrite.org/running-recipes/popular-recipe-guides/migrate-from-junit-4-to-junit-5). The migration supports partially the effort.

However, things like JUnit 4 _rules_ can't be automatically migrated. They are used quite much in the code and it will be manual effort to rewrite these test classes.

Due to the vast amount of test code it might be necessary to do this step by step. In this case JUnit 5 and 4 code have to coexist and work together in the build. This will likely require using the vintage engine.

**Proceeding**

1. Create a branch for the module to migrate
2. Run the JUnit 4 migration recipe
```
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-testing-frameworks:RELEASE -Drewrite.activeRecipes=org.openrewrite.java.testing.junit5.JUnit4to5Migration,org.openrewrite.java.testing.cleanup.RemoveTestPrefix,org.openrewrite.java.testing.cleanup.TestsShouldNotBePublic,org.openrewrite.java.testing.assertj.JUnitToAssertj,org.openrewrite.java.testing.assertj.SimplifyAssertJAssertion,org.openrewrite.staticanalysis.UnnecessaryThrows,org.openrewrite.staticanalysis.UnnecessaryExplicitTypeArguments -Drewrite.exportDatatables=true verify -DskipTests -Dskip.frontend.build=true -f <MODULEPATH>
```
3. Manual migration: The migration recipe only works to a certain degree. It is a good basis, but will leave the code uncompilable in most cases.
4. Run ```./mvnw -Popenrewrite rewrite:run -f <MODULEPATH>```
5. Clean up code (use Sonarlint plugin)

**Current Status**
Table created with Gist [list-modules-with-tests.sh](https://gist.github.com/kthoms/e563b76b317c326ab6cccd7902a64128)

```
| status    | directory                                                                                | # JUnit 4       | # JUnit 5       | # Hamcrest      |
[...]

PR #208, opened 2024-11-16
refactor(tests): JUnit5 Migration: Migrate Hamcrest to AssertJ (#27)

kthoms:
This change reduces most usages of Hamcrest and replaces it by AssertJ. The goal is to use AssertJ all over the test base.

The initial transition was done with Open Rewrite. Changes to each module have been done with separate commits. 

After that Hamcrest 1.3 was replaced by 2.2 and manual changes have been done on code that came directly along by the dependency change.

Finally code has been polished.

This is an intermediate step and as preparation of further JUnit 5 migrations, as the Open Rewrite migration recipes would do similar. By this the change later will be smaller.

Related to #27

### DIFF
--- a/commons/typed-values/pom.xml
+++ b/commons/typed-values/pom.xml
@@ -39,6 +39,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+    </dependency>
+
+    <dependency>
       <groupId>org.hamcrest</groupId>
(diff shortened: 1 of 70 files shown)


hauptmedia:
Hi @kthoms, 👋

Would it be possible to use the squash option when merging your PRs into the main branch?

This helps keep the commit history in the main branch clean and easy to follow for those tracking changes. Of course, anyone interested in the individual commits can still dive into the details right here in the PR. 😊

Thanks so much for your contribution!

Best,
Julian

PR #211, opened 2024-11-16
chore: Resolve some Sonar findings

kthoms:
- use `String#replace()` instead of `String#replaceAll` where no regexp replacement is needed
- use `instanceof` with variable assignment
- changed some string concatenation findings


kthoms:
Will open later again

Issue #220, opened 2024-11-18
chore(build): Add OpenRewrite plugin to root POM


kthoms:
**Is your feature request related to a problem? Please describe.**
When using Open Rewrite for cleanup actions I usually run the Maven CLI, recipe by recipe, module by module. Running the CLI on root level usually fails.

E.g. running:
```
mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.activeRecipes=org.openrewrite.java.RemoveUnusedImports -Drewrite.exportDatatables=true
```

fails with
```
[INFO] --- compiler:3.13.0:compile (default-compile) @ operaton-engine-feel-juel ---
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 29 source files with javac [debug target 17] to target/classes
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /Users/kthoms/Development/git/github/operaton/engine-dmn/feel-juel/src/main/java/org/operaton/bpm/dmn/feel/impl/juel/FeelEngineFactoryImpl.java:[23,45] package org.operaton.bpm.impl.juel.jakarta.el does not exist
[ERROR] /Users/kthoms/Development/git/github/operaton/engine-dmn/feel-juel/src/main/java/org/operaton/bpm/dmn/feel/impl/juel/FeelEngineFactoryImpl.java:[24,45] package org.operaton.bpm.impl.juel.jakarta.el does not exist
[ERROR] /Users/kthoms/Development/git/github/operaton/engine-dmn/feel-juel/src/main/java/org/operaton/bpm/dmn/feel/impl/juel/FeelEngineFactoryImpl.java:[91,13] cannot find symbol
````

This is due to steps that require phases beyond compile, here it is related to #84.

With fix for #84 it would fail with
```
[ERROR] Failed to execute goal org.eclipse.transformer:transformer-maven-plugin:1.0.0:transform (transform-java-sources) on project operaton-engine-spring-6: Exception transforming directory: Exception loading transformer rules: /Users/kthoms/Development/git/github/operaton/engine/target/classes (Is a directory) -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.eclipse.transformer:transformer-maven-plugin:1.0.0:transform (transform-java-sources) on project operaton-engine-spring-6: Exception transforming directory
```

**Describe the solution you'd like**
[...]

Issue #88, opened 2024-10-31
Perform code cleanups


kthoms:
**Is your feature request related to a problem? Please describe.**
The code base has grown and contains patterns that can be safely refactored. The code style should be modern and clean.

**Describe the solution you'd like**

- Use [Open Rewrite recipes](https://docs.openrewrite.org/recipes/java) and apply them on the code base. One module per Pull Request and applied recipe.
- Look up build warnings of the IDE for patterns. Resolve warnings per identified pattern (e.g. unnecessary type qualifiers)

**Instructions**

1. Look up potential cleanups by looking at Sonar issues
  - Go to [Sonarqube](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton)
  - Look at "Rules"
  - Select one where might be potential for an available OpenRewrite recipe
1. Search for a matching [Open Rewrite recipe](https://docs.openrewrite.org/recipes/staticanalysis/) 
1. Create a branch on your fork, e.g. `cleanup/remove-unused-imports`
1. Perform the cleanup action
  - Copy the command line from option "Maven command line", e.g. from "[Add @Serial annotation to serialVersionUID](https://docs.openrewrite.org/recipes/staticanalysis/addserialannotationtoserialversionuid)"
    ```
    mvn -U org.openrewrite.maven:rewrite-maven-plugin:run -Drewrite.recipeArtifactCoordinates=org.openrewrite.recipe:rewrite-static-analysis:RELEASE -Drewrite.activeRecipes=org.openrewrite.staticanalysis.AddSerialAnnotationToSerialVersionUID -Drewrite.exportDatatables=true
    ```
  -  modify the command line: optionally use `./mvnw` instead of `mvn` if Maven >=3.9 is not installed. Append `verify -DskipTests -Dskip.frontend.build` to the command line
  - execute the command, this should take a while and finally make modifications
1. Verify that the modifications do not break code, execute `./mvnw verify -DskipTests -Dskip.frontend.build`
1. Fix issues if any
1. Commit with pattern:
  ```
refactor: <short description of the cleanup>

<description_of_the_change_from_recipe>

Performed cleanup with OpenRewrite recipe <qualified_recipe_identifier>.

related to #88
  ```
 

![Screenshot 339](https://github.com/user-attachments/assets/2848c94d-b43c-4d2a-bff3-90f2af055080)

javahippie:
Do you think there is an overlap with #85 here?

[...]

PR #221, opened 2024-11-18
chore: Remove unused imports

kthoms:
Performed cleanup with Open Rewrite

related to #220, #88


kthoms:
Seems that the rewrite recipe is a bit too aggressive, it removed `import org.xmlunit.util.Predicate;`
I'll file a bug.

Issue #201, opened 2024-11-15
Set up nightly build


kthoms:
**Is your feature request related to a problem? Please describe.**
Currently there is only the regular build. It builds and tests everything, but does not produce useful output. Also it did not detect that `install` is failing.

**Describe the solution you'd like**
Set up a scheduled build job that runs nightly. It should:

- Execute `./mvnw clean install` with a build-local, not cached target m2 repository
- Build and archive assemblies
- Produce code reports and attach as build artifact. For example:
  - [Dependency Reports](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-report-mojo.html), dependency tree
  - Code Metrics: LOC, number of files per module (e.g. with [sloc-maven-plugin](https://github.com/OrhanKupusoglu/sloc-maven-plugin))
  - License Checks
  - Javadocs
  - API Diff
  - [Available Dependency Updates](https://www.mojohaus.org/versions/versions-maven-plugin/display-dependency-updates-mojo.html)
  - [Available Dependency Updates](https://www.mojohaus.org/versions/versions-maven-plugin/display-dependency-updates-mojo.html)
- [Available Maven Plugin Updates](https://www.mojohaus.org/versions/versions-maven-plugin/display-plugin-updates-mojo.html)
- [Available Property Updates](https://www.mojohaus.org/versions/versions-maven-plugin/display-property-updates-mojo.html)
- ...

PR #236, opened 2024-11-20
Set up nightly build

kthoms:
- provide `nightly-build.yml`:
  - build with goals `clean install`
  - use local uncached repository
  - archive reports
  - archive distros
- Remove matrix build from main build pipeline
- Reports
  - dependency analysis
  - dependency updates
  - plugin updates
  - sloc

relates to https://github.com/operaton/operaton/issues/201

### DIFF
--- a/.github/workflows/nighly-build.yml
+++ b/.github/workflows/nighly-build.yml
@@ -1,0 +1,97 @@
+name: Nightly Build
+
+on:
+  schedule:
+    - cron: "0 3 * * *"     # Runs at 3:00 AM UTC daily
+  workflow_dispatch:        # Allows manual trigger
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Nightly Build and Report Generation
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-latest]     # If needed multi-OS add macos-13, windows-latest
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set Up JDK 17
+        uses: actions/setup-java@v3
+        with:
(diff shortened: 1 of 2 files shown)


kthoms:
@javahippie I'm not able to trigger the build for testing from the "Actions" page when selecting the nightly build workflow.
Am I missing some right?

There should by "Run workflow" button.

javahippie:
> @javahippie I'm not able to trigger the build for testing from the "Actions" page when selecting the nightly build workflow. Am I missing some right?
> 
> There should by "Run workflow" button (see [docs](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow)).

Let me check the repository settings, I don't see it, either, might be a security / auth issue

javahippie:
If I switch to the fitting branch and check the workflow, it complains that the workflow file is invalid:

```
Invalid workflow file: .github/workflows/nighly-build.yml#L20
You have an error in your yaml syntax on line 20
```

https://github.com/operaton/operaton/actions/runs/11928166253

I think this might be the root cause?

hanshuebner:
The "Run Workflow" button for `workflow_dispatch` workflows will only appear after the workflow file has landed in the main branch.

kthoms:
@javahippie I'll check the workflow again

@hanshuebner Thanks for the pointer. But how to test such workflow changes on branches? I would like to make sure that it does what it should before going to main.

hanshuebner:
Am Mi., 20. Nov. 2024 um 23:48 Uhr schrieb Karsten Thoms <
***@***.***>:

> @hanshuebner <https://github.com/hanshuebner> Thanks for the pointer. But
> how to test such workflow changes on branches? I would like to make sure
> that it does what it should before going to main.
>
I think you need to push a dummy workflow with workflow_dispatch enabled to
main before you make changes to it on the branch, but I'm not sure.  I've
resorted to making workflow changes on the main branch mostly.

kthoms:
I'll merge this now as a starting point. Potential fixes / enhancements will come soon.

PR #243, opened 2024-11-21
fix(build,juel): Use managed version of jakarta-el and use compile scope for juel

### LINKED ISSUE #242: JUEL dependency on jakarta.el-api must have runtime scope
**Describe the bug**
Building of the _connect_ modules only fails with NoClassDefFoundError at test execution.

**To Reproduce**
Execute
```
./mvnw -am -amd -pl org.operaton.connect:operaton-connect-root
```

This will result in 
```
[ERROR]   ConnectProcessEnginePluginTest.<clinit>:48 » NoClassDefFound jakarta/el/ExpressionFactory

java.lang.NoClassDefFoundError: org/operaton/bpm/impl/juel/jakarta/el/ELContext
	at org.operaton.bpm.dmn.engine.impl.DefaultDmnEngineConfiguration.initElProvider(DefaultDmnEngineConfiguration.java:182)
	at org.operaton.bpm.dmn.engine.impl.DefaultDmnEngineConfiguration.init(DefaultDmnEngineConfiguration.java:100)
	at org.operaton.bpm.dmn.engine.impl.DefaultDmnEngineConfiguration.buildEngine(DefaultDmnEngineConfiguration.java:90)
	at org.operaton.bpm.dmn.engine.test.junit5.DmnEngineExtension.initializeDmnEngine(DmnEngineExtension.java:86)
	at org.operaton.bpm.dmn.engine.test.junit5.DmnEngineExtension.injectIntoTestInstance(DmnEngineExtension.java:130)
	at org.operaton.bpm.dmn.engine.test.junit5.DmnEngineExtension.postProcessTestInstance(DmnEngineExtension.java:101)
```

**Expected behavior**
It must be possible to build the connect modules without introducing an explicit dependency on JakartaEL in the connect modules. This should come transitive through JUEL.

kthoms:
Resolves a NoClassDefFoundError for JakartaEL classes at test runtime for modules transitvely depending on juel.

fixes #242

### DIFF
--- a/juel/pom.xml
+++ b/juel/pom.xml
@@ -20,15 +20,25 @@
     <!-- We shade artifacts into the jar, so we need to generate
     a dependency BOM for the license book -->
     <skip-third-party-bom>false</skip-third-party-bom>
-    <!-- Using Jakarta Expression Language 4.0 for Java 8 compatibility -->
-    <version.jakarta.el>4.0.0</version.jakarta.el>
   </properties>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.operaton.bpm</groupId>
+        <artifactId>operaton-core-internal-dependencies</artifactId>
+        <version>${project.version}</version>
+        <scope>import</scope>
+        <type>pom</type>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
 
   <dependencies>
     <dependency>
       <groupId>jakarta.el</groupId>
       <artifactId>jakarta.el-api</artifactId>
-      <version>${version.jakarta.el}</version>
+      <scope>compile</scope>
     </dependency>
   </dependencies>
 


internal-dependencies/pom.xml
```diff
@@ -80,6 +80,7 @@
         <artifactId>servlet-api</artifactId>
         <version>2.5</version>
       </dependency>
+      <!-- Using Jakarta Expression Language 4.0 for Java 8 compatibility -->
```
javahippie, line 83: I think this comment (although only copied / moved) might be confusing as we do not support Java 8 anymore

kthoms: I was thinking about it and thought to let it in because 4.0.0 is not the latest release. I would like to upgrade to latest Jakarta EL release with a later issue.

But brings me to the point to just remove now this comment and raise the issue for the upgrade.

kthoms: Issue raised: https://github.com/operaton/operaton/issues/252

kthoms: Comment removed

PR #245, opened 2024-11-21
Fix instanceof check and cast with pattern matching in engine. (#85)

dirk-olmes:


a-m-zill (approved):
Hello @dirk-olmes,
thx for your job.
It looks good to me. 👍🏼 

Best regards,
Andreas

PR #250, opened 2024-11-22
Fix instanceof check and cast with pattern matching in spin/dataformat-json-jackson. (#85)

dirk-olmes:
related to #85


kthoms:
```
JsonTreeJsonPathGroovyTest.shouldGetElementFromJsonPath » StackOverflow
JsonTreeEditListPropertyJavascriptTest.shouldFailInsertAfterWrongSearchObject »  Unexpected exception, expected<org.operaton.spin.json.SpinJsonPropertyException> but was<java.lang.Exception>
JsonTreeJsonPathJavascriptTest.shouldGetElementFromJsonPath »  java.lang.StackOverflowError
```
From the code I do not see on first sight why this has been caused, but you should find that out when debugging one of the failing tests.

kthoms (approved):
Thx, @dirk-olmes !

PR #234, opened 2024-11-20
chore(build): Fix surefire coverage configuration

kthoms:
The property `jacoco.argLine` did not contribute to `surefire.argLine` in the build. The jacoco goal `prepare-agent` was properly executed and set property `jacoco.argLine`, but at that point of time the `surefire.argLine` was already evaluated.

Fixed this by moving the surefire plugin configuration to the profiles `coverage` and `coverage-not-enabled` and specifying the `argLine` config arg once with `jacoco.argLine` and once without.

fixes #230

PR #255, opened 2024-11-23
fix(build): Fix missing coverage data

kthoms:


kthoms:
replaced by https://github.com/operaton/operaton/pull/234

PR #256, opened 2024-11-23
feat(engine): don't fail on change variables async when process instance does not exist

hauptmedia:
Related to https://github.com/camunda/camunda-bpm-platform/issues/4795

Backported commit 4b101ea9cbe from the camunda-bpm-platform repository.

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/batch/variables/BatchSetVariablesHandler.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/batch/variables/BatchSetVariablesHandler.java
@@ -51,7 +51,7 @@ public class BatchSetVariablesHandler extends AbstractBatchJobHandler<BatchConfi
 
     for (String processInstanceId : processInstanceIds) {
       commandContext.executeWithOperationLogPrevented(
-          new SetExecutionVariablesCmd(processInstanceId, variables, false, true));
+          new SetExecutionVariablesCmd(processInstanceId, variables, false, true, false));
     }
   }
 
--- a/engine/src/test/java/org/operaton/bpm/engine/test/util/ExecutableProcessUtil.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/test/util/ExecutableProcessUtil.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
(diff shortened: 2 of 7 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/batch/variables/BatchSetVariablesHandler.java
```diff
@@ -51,7 +51,7 @@ public void executeHandler(BatchConfiguration batchConfiguration,
 
     for (String processInstanceId : processInstanceIds) {
       commandContext.executeWithOperationLogPrevented(
-          new SetExecutionVariablesCmd(processInstanceId, variables, false, true));
+          new SetExecutionVariablesCmd(processInstanceId, variables, false, true, false));
```
javahippie, line 54: I wonder if it would be nice to introduce variables / constants here, because the trailing parameters `false`, `true`, `false` don't carry a lot of semantics. @hauptmedia @kthoms what do you think, would that make the code more readable, or are IDEs helping us here, anyways?

engine/src/test/java/org/operaton/bpm/engine/test/util/ExecutableProcessUtil.java
javahippie, line 1: Personally, I would not introduce a Util for creating test data, as this will couple several tests together, but I don't have really *strong* opinions on this one, would be fine for me to leave it in

engine/src/main/java/org/operaton/bpm/engine/impl/batch/variables/BatchSetVariablesHandler.java
hauptmedia: My understanding of this patch (without looking too much into the details) is that this a patch within a bigger effort to unify the behaviour for all batch operations until 7.24 LTS. So from a functional point of view it is important that we follow these changes. Refactoring the internal command structure seems to me another task where we should have a broader look how the command api can be improved in general.

javahippie: Works for me, thanks!

javahippie (approved):

engine/src/test/java/org/operaton/bpm/engine/test/util/ExecutableProcessUtil.java
hauptmedia: I also don't have a strong opinion on this. As long tests don't reference these kind of helper classes from other maven artifacts it should be ok to share common used definitions for multiple tests.

Issue #251, opened 2024-11-22
NullPointerException in test execution of operaton-external-task-client (flaky)


kthoms:
**Describe the bug**
I'm exeperiencing sometimes failing builds in the integration tests of module `operaton-external-task-client`.

```
[INFO] --- failsafe:3.5.1:integration-test (default) @ operaton-external-task-client ---
[INFO] Failsafe report directory: /Users/karstenthoms/Development/projects/operaton/clients/java/client/target/failsafe-reports
[INFO] Using auto detected provider org.apache.maven.surefire.junitplatform.JUnitPlatformProvider
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.operaton.bpm.client.variable.pa.PaSerializationIT
[ERROR] Tests run: 3, Failures: 0, Errors: 3, Skipped: 0, Time elapsed: 0.198 s <<< FAILURE! -- in org.operaton.bpm.client.variable.pa.PaSerializationIT
[ERROR] org.operaton.bpm.client.variable.pa.PaSerializationIT.shouldSelectSequenceFlowAndCompleteProcessInstance_XmlSerialization -- Time elapsed: 0.183 s <<< ERROR!
java.lang.RuntimeException: java.lang.NullPointerException: Cannot invoke "org.apache.hc.client5.http.impl.classic.CloseableHttpClient.execute(org.apache.hc.core5.http.ClassicHttpRequest, org.apache.hc.core5.http.io.HttpClientResponseHandler)" because "this.httpClient" is null
        at org.operaton.bpm.client.rule.EngineRule.startProcessInstance(EngineRule.java:304)
        at org.operaton.bpm.client.rule.EngineRule.startProcessInstanceByKey(EngineRule.java:227)
        at org.operaton.bpm.client.variable.pa.PaSerializationIT.setup(PaSerializationIT.java:91)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
        Suppressed: java.lang.NullPointerException: Cannot invoke "org.operaton.bpm.client.dto.ProcessInstanceDto.getId()" because "this.processInstance" is null
                at org.operaton.bpm.client.variable.pa.PaSerializationIT.tearDown(PaSerializationIT.java:99)
                ... 3 more
[...]

PR #260, opened 2024-11-24
bugfix(test): Changed the strategy for initializing JUnit Extensions …

javahippie:
Fixes [#251]

### DIFF
--- a/clients/java/client/src/it/java/org/operaton/bpm/client/variable/FileSerializationIT.java
+++ b/clients/java/client/src/it/java/org/operaton/bpm/client/variable/FileSerializationIT.java
@@ -18,7 +18,7 @@ package org.operaton.bpm.client.variable;
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.api.extension.RegisterExtension;
 import org.operaton.bpm.client.ExternalTaskClient;
 import org.operaton.bpm.client.dto.ProcessDefinitionDto;
 import org.operaton.bpm.client.dto.ProcessInstanceDto;
@@ -48,6 +48,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -59,8 +60,6 @@ import static org.operaton.bpm.client.util.ProcessModels.USER_TASK_ID;
 import static org.operaton.bpm.client.util.ProcessModels.createProcessWithExclusiveGateway;
 import static org.operaton.bpm.engine.variable.type.ValueType.FILE;
 
-@ExtendWith(EngineRule.class)
-@ExtendWith(ClientRule.class)
 public class FileSerializationIT {
 
   protected static final String VARIABLE_NAME_FILE = "fileVariable";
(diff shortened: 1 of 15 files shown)


javahippie:
Currently the error scnearios are failing

.github/workflows/build.yml
```diff
@@ -74,8 +74,9 @@ jobs:
             mkdir -p "$dir/target"
             touch "$dir/target/executeJacoco"
           fi
-        done
-        ./mvnw package
+        done 
+        ./mvnw fr.jcgay.maven.plugins:buildplan-maven-plugin:list
```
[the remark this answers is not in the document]
hauptmedia: why not, the whole builds takes 25 minutes, so 2-3 secs don't really matter ;)

javahippie:
Down to one error:

```
 Error:  Failures: 
Error:    FileSerializationIT.shouldFailWhenCompletingWihtDeferredFileValue:586 
Expecting value to be true but was false
```

.github/workflows/build.yml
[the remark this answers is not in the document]
a-m-zill: @javahippie i think also to leave it like it is, because this does not speed up so much 😁

clients/java/client/src/it/java/org/operaton/bpm/client/variable/FileSerializationIT.java
```diff
@@ -567,32 +569,29 @@ public void shouldFailWhenCompletingWihtDeferredFileValue() {
...
     // then
     List<VariableInstanceDto> variableInstances = engineRule.getVariablesByProcessInstanceIdAndVariableName(processInstance.getId(), null);
-    assertThat(variableInstances).hasSize(2);
 
     List<String> variableNames = new ArrayList<>();
     for (VariableInstanceDto variableInstance : variableInstances) {
       variableNames.add(variableInstance.getName());
     }
 
-    assertThat(variableNames).containsExactlyInAnyOrder(VARIABLE_NAME_FILE, ANOTHER_VARIABLE_NAME_FILE); // contains not "deferredFile"
```
javahippie, line 595: This nearly drove me mad. I'm suspecting, that due to the expected thrown error in the previous JUnit 4 tests, there was an error introduced in the test.

If we complete an external task service with two variables, of which one causes a serialization error, my expectation would be that the transaction is rolled back in the db and NONE of the two variables is in the database. I cannot understand the usecase which makes one of two variables and up in the database.

@a-m-zill can I get an opinion here?

.github/workflows/build.yml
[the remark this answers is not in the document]
kthoms: The buildplan will be downloadable from today's nightly build.

clients/java/client/src/it/java/org/operaton/bpm/client/variable/FileSerializationIT.java
a-m-zill: Hi @javahippie,
as we have talked about in our chat, i agree that this is not a right behavior to expect one variable saved and one not of two new variables in one transaction.
Further i suggest to check in the end of the test that the process instance is again or still on the external task. Because that it the position where it should be.

shiftyseaotter: Hi, the documentation for the ExpectedException rule says: 
"This rule does not perform any special magic to make execution continue as if the exception had not been thrown. So it is nearly always a mistake for a test method to have statements after the one that is expected to throw the exception"
So it is very likely that variables were never saved and the subsequent test assertion never happened, but the test still reported success because the expected exception was thrown as intended. So I think your expectations are correct and the test had unreachable false assertions

javahippie: @shiftyseaotter thanks for reading the docs and finding this! This might have implications on other tests which use / used `ExcpectedException`. @kthoms Something to keep in mind for upcoming refactorings

javahippie: @a-m-zill Added a check for the currently active topic, and refactored the code a little, happy about an approve if it works for you!

a-m-zill (approved):
Thx @javahippie, that LGTM 👍🏼

kthoms:
Thank you @javahippie !

PR #263, opened 2024-11-25
chore(tests): bump selenium to 4.26.0 and chromedriver to 131.0.6778.85

hauptmedia:
related to cibseven/cibseven#9

Backported commit f625b48 from the cibseven repository.
Original author: Serge Krot <serge.krot@cib.de>

### DIFF
--- a/distro/run/qa/pom.xml
+++ b/distro/run/qa/pom.xml
@@ -20,7 +20,6 @@
     <version.jersey-json>1.15</version.jersey-json>
     <version.jersey-apache-client>1.15</version.jersey-apache-client>
     <version.httpcomponents>4.5.10</version.httpcomponents>
-    <version.chromedriver>112.0.5615.49</version.chromedriver>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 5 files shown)


javahippie:
Did you run the integration tests locally for testing? I think these changes are not run in the default pipeline

kthoms:
Do we need now a different label or not?

hauptmedia:
I think a new label would be good for Github, so you can filter for it. For the release notes we don't have to differentiate if something was a backport

kthoms:
@hauptmedia Do you think this change is safe to merge? I don't see what speaks against it. But I can't test this ATM.
If you think this is OK, then remove the draft and confirm.

kthoms:
Thanks for the update.

Do you know what needs to be activated to run the integration tests? Then I would enable them in the build.

hauptmedia:
The integration tests are described here https://github.com/operaton/operaton/blob/main/qa/README.md

javahippie:
I'd suggest closing this, @hauptmedia, I think we will have our own dependency adjustments when the integration tests are up and running again

Issue #37, opened 2024-10-24
Set up automatic release process


kthoms:
**Is your feature request related to a problem? Please describe.**
The project needs a release process that performs several tasks

- Building the release artifacts
- Deploy Jar artifacts to Maven Central
- Deploy release artifacts to GitHub
- Create and deploy change logs
- Tag sources
- Prepare sources for next development increment
- Update documentation
- Announcements
- Increment versions in sources
   - pom.xml
   - package.json

**Describe the solution you'd like**
The release process should be automated as far as possible. It should be triggered by a GitHub workflow with required parameters.

I would propose to leverage [JReleaser](https://jreleaser.org/guide/latest/index.html).
The setup for GitHub actions is described [here](https://jreleaser.org/guide/latest/continuous-integration/github-actions.html).

The setup will require to store some secrets, e.g. for Maven Central and PGP, and a token that allows JReleaser to make changes to the GitHub project.

**Describe alternatives you've considered**
The release actions can be scripted with Shell scripts and integrated in a GitHub workflow. However, many repeatable tasks are already provided by a solution like JReleaser.

**Additional context**
- none -

javahippie:
I'm in favor of JReleaser, I never had a use case for it but I like the concept

kthoms:
I could work on it, but would need assistance setting up the secrets.

javahippie:
That's not an issue, I can assist with that. We could also work on this in a synchronuous session if you'd like

hauptmedia:
We also need to decide how we should handle this => https://github.com/camunda/camunda-bpm-platform/commit/396d635ce7

This should be automated by the release pipeline

kthoms:
@javahippie I would need to know what we would like to publish to Maven Central.
IMO the BOM `bom/operaton-only-bom/pom.xml` is a comprehensive list of modules to expect in Maven Central, the included BOMs recursively.
As distros "tomcat webapp" and "sqlscripts" come by default from the build, should be enough?

kthoms:
package.json has to be updated, see https://github.com/operaton/operaton/pull/269/files

kthoms:
For deployment to Maven Central there are some things to consider:

**Not all modules have to be deployed**
Deployment should be done only for a subset of modules. A large subset, though.
I'm considering currently this approach:
- By default (root POM) the maven-deploy-plugin is skipped by a property `deploy.skip`
[...]

PR #267, opened 2024-11-25
feat(build): Set up release workflow (#37)

kthoms:
Provide a `release` workflow and `jrelease.yaml` for automatic release builds.

relates to #37

### DIFF
--- a/.github/jreleaser/changelog.tpl
+++ b/.github/jreleaser/changelog.tpl
@@ -1,0 +1,10 @@
+# About this release
+
+## New and Noteworthy
+
+{{changelogNewAndNoteworthy}}
+
+## Changelog
+
+{{changelogChanges}}
+{{changelogContributors}}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,150 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_branch:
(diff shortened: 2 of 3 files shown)


kthoms:
Currently work in progress, a copy from the recipe repo.

The current changelog, it will have to be tweaked with labelers, categories and rewriters:
[CHANGELOG.md](https://github.com/user-attachments/files/17910266/CHANGELOG.md)

kthoms:
I would use this change as a starting point and would like to merge that. From then on, I could create a branch and operate on the branch to finalize the release workflow / configuration. In order to be able to do this, the workflow has once to be on master.

hauptmedia:
Sure, we really need to get this working to push the beta-2 release out. It has so many changes and improvements compared to beta-1 that the release of beta-2 should be a priority.

kthoms:
This should be a good starting point now.
@javahippie If you don't mind then I'll merge that.

kthoms:
> ```
> platform:
>   replacements:
>     aarch_64: aarch64
> `` 
> ```

Copied this from JReleaser's configuration. Not sure when this will become relevant, likely when we start shipping executable distributions, e.g. Operaton as executable for Mac via homebrew.

Quite sure that we do not need this ATM.

kthoms:
I'll work on the configuration anyway with the next steps. This is just the starting point.

PR #269, opened 2024-11-26
chore(webapps): adjusted version of the webapps to be 1.0.0-beta-2

hauptmedia:
The webapps have their own versioning, this should match the project's overall version. Should be merged before beta-2 release.

### DIFF
--- a/engine-rest/docs/package.json
+++ b/engine-rest/docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "docs",
-  "version": "1.0.0-beta-1",
+  "version": "1.0.0-beta-2",
   "private": true,
   "scripts": {
     "build": "node build"
--- a/webapps/frontend/package.json
+++ b/webapps/frontend/package.json
@@ -1,7 +1,7 @@
 {
   "name": "operaton-bpm-webapp",
   "private": true,
-  "version": "1.0.0-beta-1",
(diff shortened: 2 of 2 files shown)


hauptmedia:
Just found some more beta-1 references, will add them to the commit

hauptmedia:
What about these files? Must they stay at beta-1 for the build to run?

Bildschirmfoto 2024-11-26 um 14 05 34

javahippie:
My understanding is, that JReleaser will adjust all version numbers in POM files on its own. If JReleaser is not able to change versions in package.json on its own, we should find an automated way of doing this, I would like to avoid all manual steps necessary for a new release

hauptmedia:
I changed it only for the webapps/doc packages now for preparation of the beta-2 release

kthoms:
JReleaser won't change the versions. We have to do this in the release workflow. I'm changing the POM versions with the versions plugin. I will take care of updating them also. Noted in #37 .

kthoms:
@hauptmedia The `run` modules have not been catched up by the version bump. `operaton-bpm-run-root` is not hooked in the reactor. Likely this should be added to the root POM. Anyway, the version should be bumped.

kthoms:
This change is fine anyway. Let's deal in a next PR.

PR #277, opened 2024-11-27
Fix instanceof check and cast with pattern matching in commons. (#85)

CBorowski-dev:


kthoms (changes requested):
Could you try to avoid format changes? The javadocs should be untounched.

CBorowski-dev:
Oh, I used a different IDE and didn't notice that "auto-format on save" or a similar feature was activated. Will undo the unwanted changes.

kthoms (approved):
Looks good now, thx.

javahippie (approved):
lgtm, too!

PR #281, opened 2024-11-27
Make distros downloadable as separate artifacts

### LINKED ISSUE #280: Split "distros" in single downloadable artifacts
The Nightly Build produces a single zip "distros" containing ATM the Tomcat tar.gz and SQL Scripts distro together.

Screenshot 323

The single distros should be downloadable as separate build artifacts. And not only by the Nightly Build but also the regular build.

Change `build.yml` and `nightly_build.yml`.

kthoms:
This change makes the distributions "Operaton (Tomcat Bundle)" and "Operaton SQL Scripts" available as separately downloadable artifacts. Before, both distros were packaged into a single "distros" zip file.

fixes #280

### DIFF
--- a/.github/scripts/prepare-distros-archive.sh
+++ b/.github/scripts/prepare-distros-archive.sh
@@ -1,7 +1,0 @@
-#!/usr/bin/env bash
-TARGET_DIR=$1
-
-echo "📦 Collecting distros"
-mkdir -p $TARGET_DIR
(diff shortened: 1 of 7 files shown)


kthoms:
I want to put the Tomcat version into the artifact name. This version should be read in the "Build" job and passed on to the "Publish" job. This is not working as expected so far.
I could hard-code this, but want to avoid it.

PR #289, opened 2024-11-28
Fix instanceof check and cast with pattern matching in webapps.

CBorowski-dev:
relates to #85

### DIFF
--- a/.github/jreleaser/changelog.tpl
+++ b/.github/jreleaser/changelog.tpl
@@ -4,6 +4,13 @@
 
 {{changelogNewAndNoteworthy}}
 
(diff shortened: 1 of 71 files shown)


kthoms:
@CBorowski-dev The PR contains quite some commits for the small changeset. This is because you use the "merge" option with git. You should think about using "rebase" instead. This keeps the commits cleaner.

Not a problem yet, the commits are anyway squashed before merging to `main`.

kthoms (approved):

kthoms:
@CBorowski-dev Please add also "relates to #85" to the commit message body and PR description.

PR #295, opened 2024-11-29
Fix instanceof check and cast with pattern matching in juel. (#85)

CBorowski-dev:
relates to #85

### DIFF
--- a/.github/scripts/add_description.sh
+++ b/.github/scripts/add_description.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+# Search for all pom.xml files in the current directory and subdirectories
+find . -name "pom.xml" | while read pom_file; do
+  # Check if the <description> tag already exists
+  if ! grep -q "<description>" "$pom_file"; then
+    echo "Adding <description> tag to $pom_file..."
+    # Add the <description> tag before the closing </project> tag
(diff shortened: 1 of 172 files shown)


CBorowski-dev:
Fix instanceof check and cast with pattern matching in juel. (#85)

juel/src/main/java/org/operaton/bpm/impl/juel/NumberOperations.java
```diff
@@ -32,9 +32,7 @@ private final static boolean isDotEe(String value) {
 		int length = value.length();
 		for (int i = 0; i < length; i++) {
 			switch (value.charAt(i)) {
-				case '.':
-				case 'E':
-				case 'e': return true;
+				case '.', 'E', 'e': return true;
```
javahippie, line 35: Maybe refactor this switch with a set#contains check? Having a switch with one return statement reads a little funny to me

javahippie (commented):
Thanks! One minor comment, but it's not a strong opinion, and definitely not a dealbreaker for me.

One additional remark: When merging these PR, we often have to adapt the commit messages to match our commit message schema which is mentioned here: https://github.com/operaton/operaton/blob/main/CONTRIBUTING.md#commit-message-conventions

If you could follow that schema in your commit messages, it would save us some time.

PR #298, opened 2024-11-29
add <description> tag to pom.xml files

xProppi:
I added a <description> tag to all pom.xml files across the repository to meet the requirements for Maven Central, where strict checks can fail if this tag is missing. The description is set to ${project.name} by default, but it can be adjusted if necessary. I also left in the script which I used.

Let me know if any further modifications are needed.


kthoms (changes requested):
@xProppi Wonderful. Yes, a bit dummy to have this for now, but over the time should by more meaningful descriptions.

Could you just move the shell script to .github/scripts. Not a perfect location for scripts outside of the scope of the build itself, but I would like to keep the number of files in the project root as small as possible.

And as first time contributor, please just confirm you are fine that you contribute changes you do under APACHE-2.0 license.

kthoms:
And side note: We are preparing the release and automatic generation of a contributors list. If you want to be mentioned there by name, then you would need to update your GitHub profile. But no requirement here.

xProppi:
Thank you for your feedback! I’ve moved the script to the `.github/scripts` folder as suggested. Regarding the license, I confirm that I’m fine contributing under the APACHE-2.0 license. I also updated my GitHub profile, which was previously sparse since I primarily used GitLab up until now.

kthoms (approved):
Thanks a lot for providing your first contribution. Would love to get more from you.

PR #299, opened 2024-11-29
Add script to update the version in all project files

kthoms:
Private the scipt `set-project-version.sh`. This reads the current version from the root pom.xml and takes the version to set as argument.

It changes all versions and checks afterwards that no match is left.

### DIFF
--- a/.github/scripts/set-project-version.sh
+++ b/.github/scripts/set-project-version.sh
@@ -1,0 +1,61 @@
+#!/usr/bin/env bash
+
+if [ ! -f "mvnw" ]; then
+  echo "⚠️ Maven Wrapper not found. You must execute this script from the project root directory. Exiting..."
+  exit 1
+fi
+
+if [ -z "$1" ]; then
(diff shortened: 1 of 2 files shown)


.github/scripts/set-project-version.sh
```diff
@@ -0,0 +1,61 @@
...
+PACKAGE_JSON_FILES=(\
+  $(find . -name package.json) \
+  $(find . -name package-lock.json) \
+)
+
+for PACKAGE_JSON_FILE in ${PACKAGE_JSON_FILES[@]}; do
+  sed -i '' -e "s/\"version\": \"$CURRENT_VERSION\"/\"version\": \"$1\"/" $PACKAGE_JSON_FILE
+done
+
+echo "🔄 Updating version in pom.xml files that are not part of the reactor"
```
kthoms, line 30: **Review Note**: I know this is a hack. But not all modules are currently in the reactor, so they can't be matched by the versions-maven-plugin.
Should be fixed later.

PR #300, opened 2024-11-29
refactor(juel): Incorporate new language features

CBorowski-dev:
- Introduces comma-separated labels in switch cases
- instanceof checks and casts are replaced with instanceof with assign to a binding variable
- Code simplifications

related to #85

### DIFF
--- a/juel/src/main/java/org/operaton/bpm/impl/juel/NumberOperations.java
+++ b/juel/src/main/java/org/operaton/bpm/impl/juel/NumberOperations.java
@@ -29,7 +29,13 @@ public class NumberOperations {
 	private final static Long LONG_ZERO = 0L;
 
 	private final static boolean isDotEe(String value) {
-    return value.contains(".") || value.contains("E") || value.contains("e");
+		int length = value.length();
+		for (int i = 0; i < length; i++) {
+			switch (value.charAt(i)) {
+				case '.', 'E', 'e': return true;
+			}
+		}
+		return false;
 	}
 
 	private final static boolean isDotEe(Object value) {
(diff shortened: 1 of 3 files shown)


juel/src/main/java/org/operaton/bpm/impl/juel/NumberOperations.java
```diff
@@ -29,15 +29,7 @@ public class NumberOperations {
...
-		int length = value.length();
-		for (int i = 0; i < length; i++) {
-			switch (value.charAt(i)) {
-				case '.':
-				case 'E':
-				case 'e': return true;
-			}
-		}
-		return false;
+    return value.contains(".") || value.contains("E") || value.contains("e");
```
kthoms, line 32: I fear this implementation is inefficient. Although I expect that the evaluated strings are small, just consider the case (totally irreal, just to demonstrate):
Suppose you have a 2000 char string not containing any of these characters. This could scan index of a String match (not even for a single charater!) 3 times!
It looks stupid, but your original implementation with the switch would be efficient. Yes it looks awkward, but just let it as you had. It has 3 cases, although in a single line.

CBorowski-dev: It's not that simple. I tested the original code and the contains code with various worst case scenarios. For short strings (length 10-20) the original code is twice as fast and the time difference on my machine is ~3500ns. Interestingly, for longer strings (length 2000-4000), the contains code is "only" 20-30% slower. Optimizations, cache hits, etc. play an important role. On the other hand, the code readability is significantly better with the contains variant, imho.
Anyway, I put the original code back (for performance reasons) and merged the case lines.

kthoms (approved):
Thx for the update

juel/src/main/java/org/operaton/bpm/impl/juel/NumberOperations.java
kthoms: Since the string usually will be shorter and this code is potentially called often, performance beats style.
Thanks for measuring this, I learned again something.

PR #305, opened 2024-12-02
chore(engine): Fix instanceof check and cast with pattern matching in engine

geduldia:
related to operaton#85


engine/src/main/java/org/operaton/bpm/engine/impl/scripting/engine/JuelScriptEngine.java
```diff
@@ -178,11 +178,11 @@ private static Method getImportMethod() {
 
   public static void importFunctions(ScriptContext ctx, String namespace, Object obj) {
     Class< ? > clazz = null;
-    if (obj instanceof Class) {
-      clazz = (Class< ? >) obj;
-    } else if (obj instanceof String) {
+    if (obj instanceof Class classObj) {
```
kthoms, line 181:
```suggestion
    if (obj instanceof Class<?> classObj) {
```

kthoms (commented):
Just one minor

kthoms (approved):

Issue #203, opened 2024-11-15
Handle SQL Migration Versions


hauptmedia:
We need to decide how to manage SQL migration versioning.

Currently, the database version is set to 7.22.0 by Operaton. To ensure compatibility, we need to:

1.	Increase the database version to 7.23.0 after some release.
2.	Do this before merging any changes that rely on a modified data schema.

Reference:
See this commit for context: [396d635ce7](https://github.com/camunda/camunda-bpm-platform/commit/396d635ce7).

Note: This commit has not been backported yet

kthoms:
@hauptmedia Do we need to set it actually to 7.23.0, or might be 8.0.0 a valid increment, although the schema does not change? IMO it is confusing to have a version number related to Camunda here.

hauptmedia:
As long as we do not introduce database changes ourselfes we should stay with the C7 versioning scheme until 7.24. I already saw dabase changes incoming for 7.23.0. It's the only way - without doing bigger efforts - to seamlessly migrate all old c7 versions and the future 7.23/7.24 versions. Setting it to 8.0.0 now would mean we cannot differentiate if the user already applied the 7.23 or 7.24 database schema update

hauptmedia:
Looking at the SQL migration files [here](https://github.com/operaton/operaton/blob/31d20645021551da572ed458a1fddddedf10da15/engine/src/main/resources/org/operaton/bpm/engine/db/upgrade/postgres_engine_7.21_to_7.22.sql), I’ve noticed that there’s no check to verify if the column already exists before adding it.

This can cause issues for users upgrading from versions 7.23 or 7.24, as the migration will fail when attempting to add the column again with a version 8.0. To resolve this, we need to update the SQL scripts to ensure the column is added only if it doesn’t already exist.

This adjustment is straightforward for PostgreSQL, but we’ll also need to implement similar checks for other supported databases: MSSQL, MySQL, Oracle, DB2, and H2.

So my strategy is to delay this as much as possible and there is a good reason for us doing db changes that are Operaton only.

hauptmedia:
This issue is resolved with 86a8cf6

PR #309, opened 2024-12-02
chore(version): Update database version

hauptmedia:
* Add instance migration project

* Adjust create scripts

* Create upgrade scripts

* Add the new changeSets to liquibase

* add migration fixture for 7.23

* update old engine test

* remove telemetry fields from old engine cfg

* chore(database): remove telemetryReporterActivate from old-engine and rolling-update tests

* adjust old version for spin and connect

* exclude DatabaseNamingConsistencyTest in test-old-engine

* chore(database): activate oracle23 in old-engine stage in CI (#4677)

related to #203
related to camunda/camunda-bpm-platform#4556

Backported commit 396d635ce7 from the camunda-bpm-platform repository.
Original author: psavidis <69160690+psavidis@users.noreply.github.com>

### DIFF
--- a/engine/src/main/resources/org/operaton/bpm/engine/db/liquibase/operaton-changelog.xml
+++ b/engine/src/main/resources/org/operaton/bpm/engine/db/liquibase/operaton-changelog.xml
@@ -140,8 +140,16 @@
              stripComments="true"/>
   </changeSet>
 
-  <changeSet author="Camunda" id="7.22.0-tag">
-    <tagDatabase tag="7.22.0"/>
+  <changeSet author="Camunda" id="7.22-to-7.23">
+    <sqlFile path="upgrade/${db.name}_engine_7.22_to_7.23.sql"
+             encoding="UTF-8"
+             relativeToChangelogFile="true"
+             splitStatements="true"
+             stripComments="true"/>
+  </changeSet>
+
+  <changeSet author="Camunda" id="7.23.0-tag">
+    <tagDatabase tag="7.23.0"/>
   </changeSet>
 
 </databaseChangeLog>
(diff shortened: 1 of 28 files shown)


kthoms:
I can't judge that ATM and am not sure if this should go into the beta-2. I would feel more comfortable if this would go into the next beta.

hauptmedia:
Now that we have released the beta-2 we can merge this for beta-3. We also have this sql scripts artifacts now, I think until 7.24 EOL version we just need to advice anyone coming from a version >7.22 to manually check and execute the db migration scripts. These are only tiny changes adding a field here and there so it's absolutely doable.

engine/src/main/resources/org/operaton/bpm/engine/db/upgrade/h2_engine_7.22_to_7.23.sql
```diff
@@ -0,0 +1,19 @@
+--
+-- Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
```
kthoms, line 2: How do we deal with new files? When they are copied from existing it might be necessary to keep the original copyright notice, but somehow you should be incorporated.

Same for all other new files.

engine/src/main/resources/org/operaton/bpm/engine/db/liquibase/operaton-changelog.xml
```diff
@@ -140,8 +140,16 @@
              stripComments="true"/>
   </changeSet>
 
-  <changeSet author="Camunda" id="7.22.0-tag">
-    <tagDatabase tag="7.22.0"/>
+  <changeSet author="Camunda" id="7.22-to-7.23">
```
kthoms, line 143: author?

engine/src/main/resources/org/operaton/bpm/engine/db/upgrade/h2_engine_7.22_to_7.23.sql
hauptmedia: this PR is actually a backported commit, so in this case we would have to leave the copyright notice as it is

hauptmedia:
Added a forum thread for it: https://forum.operaton.org/t/deciding-on-the-database-versioning-for-operaton/105/2

javahippie:
If I interpret our discussion in the forum correctly, this can be merged as is, I think?

hauptmedia:
That is true, i just wanted to wait until today so that everybody had a chance to comment on this

PR #323, opened 2024-12-06
Fix instanceof check and cast with pattern matching (#85)

dirk-olmes:
related to #85

### DIFF
--- a/engine-cdi/core/src/main/java/org/operaton/bpm/engine/cdi/impl/annotation/StartProcessInterceptor.java
+++ b/engine-cdi/core/src/main/java/org/operaton/bpm/engine/cdi/impl/annotation/StartProcessInterceptor.java
@@ -63,7 +63,7 @@ public class StartProcessInterceptor implements Serializable {
       return result;
     } catch (InvocationTargetException e) {
       Throwable cause = e.getCause();
-      if (cause != null && cause instanceof Exception exception) {
+      if (cause instanceof Exception exception) {
         throw exception;
       } else {
         throw e;
--- a/engine-cdi/core/src/main/java/org/operaton/bpm/engine/cdi/impl/annotation/StartProcessInterceptor.java
+++ b/engine-cdi/core/src/main/java/org/operaton/bpm/engine/cdi/impl/annotation/StartProcessInterceptor.java
@@ -63,8 +63,8 @@ public class StartProcessInterceptor implements Serializable {
       return result;
     } catch (InvocationTargetException e) {
       Throwable cause = e.getCause();
-      if(cause != null && cause instanceof Exception) {
-        throw (Exception) cause;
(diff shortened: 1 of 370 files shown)


engine-dmn/engine/src/main/java/org/operaton/bpm/dmn/engine/impl/type/DateDataTypeTransformer.java
```diff
@@ -56,7 +56,7 @@ public TypedValue transform(Object value) throws IllegalArgumentException {
       Date date = transformString(stringValue);
       return Variables.dateValue(date);
 
-    } if (value instanceof ZonedDateTime zonedDateTimeValue) {
+    } else if (value instanceof ZonedDateTime zonedDateTimeValue) {
       Instant instant = (zonedDateTimeValue).toInstant();
```
kthoms, line 60:
```suggestion
      Instant instant = zonedDateTimeValue.toInstant();
```

kthoms (changes requested):

engine-cdi/core/src/main/java/org/operaton/bpm/engine/cdi/impl/annotation/StartProcessInterceptor.java
```diff
@@ -63,8 +63,8 @@ public Object invoke(InvocationContext ctx) throws Exception {
       return result;
     } catch (InvocationTargetException e) {
       Throwable cause = e.getCause();
-      if(cause != null && cause instanceof Exception) {
-        throw (Exception) cause;
+      if (cause != null && cause instanceof Exception exception) {
```
javahippie, line 66: Just a question on how to proceed here: A user commented on a commit some time earlier, that `instanceof` is null safe. If the passed parameter is null, it is not an instance of `Exception` and the check is `false`, skipping the body. Is this something we should already improve here?

javahippie: See https://docs.oracle.com/javase/specs/jls/se17/html/jls-15.html#jls-15.20.2

> At run time, the result of the type comparison operator is determined as follows:
>
 >   If the value of the RelationalExpression is the null reference ([§4.1](https://docs.oracle.com/javase/specs/jls/se17/html/jls-4.html#jls-4.1)), then the result is false.

kthoms: the null check is obsolete then

kthoms (approved):

kthoms:
Sorry @dirk-olmes I did not recognize that the findings have already been adressed for a long time. Just ping in a comment. Did you know that you could (re-)request a review when clicking on one of the reviewers in top right?

Issue #246, opened 2024-11-21
Revise disabled tests


kthoms:
**Is your feature request related to a problem? Please describe.**
There are currently 137 disabled tests in the code base.

**Describe the solution you'd like**
These tests should be revised. There are different possibilities here:
- The cause why the test has been disabled was resolved. Then the test can be enabled.
- The test itself is bogus. Then the test case should either be removed or fixed.
- The reason why this test was disabled is still a problem. Then open a bug report and leave a comment in the Disabled annotation with reference to the issue.

Raise pull requests. Could be also for single issues.

WoSchmo:
I searched for "@Disabled" in the whole project and only got 8 results. Has anyone already taken care of the other 131 or is there another way to disable tests?

javahippie:
Hey, i found the ticket here: https://jira.camunda.com/browse/CAM-9441

It was referenced in this commit: https://github.com/camunda/camunda-bpm-platform/commit/eade21a100a3cac43d9ab1e4f33c8ccb08f42e8a

Interesting, that the test was added and immediately disabled.

kthoms:
@WoSchmo Most tests are still JUnit 4, search for `@Ignore, too.

kthoms:
> Interesting, that the test was added and immediately disabled.

This could be to reproduce the bug by a test. Someone who would like to solve the issue could likely use the test.

javahippie:
I, too, think, that this was the plan. The ticket was then closed as "not relevant" and the tests were forgotten

kthoms:
Does not mean that it is irrelevant for us. But likely low prio.

kthoms:
Anyway, @WoSchmo, thanks for looking at these tests. If you feel comfortable to change something, just do it. If you need to discuss something, maybe it is best on our Slack channel. Likely not every discussion related to these tests is necessary to persist in the issue or forum. But feel free to communicate however you like.

PR #324, opened 2024-12-06
qa: enable tests, which don't fail anymore (#246)

WoSchmo:
related to #246


javahippie:
Thanks for the updates! Just to be sure, you were able to simple enable all but one test without any changes to production code? Not to be doubtful, I'm just wondering why they were disabled

javahippie:
Looks great, thanks for enabling those tests again!

PR #326, opened 2024-12-07
feat(engine): Add property for skipping locked jobs in the job executor query [#264]

### LINKED ISSUE #264: Add a configuration to skip locked rows in the select statement for the job executor
[In our forum](https://forum.operaton.org/t/job-execution-rejected/68/39?u=javahippie) we experimented with adding `FOR UPDATE SKIP LOCK` to the select statement for the job executor. This does not work for all supported databases, but sped up execution for many nodes running in parallel.

 We might consider adding an opt-in configuration flag to enable this behavior, so users running on a database which supports this could benefit from higher concurrency.

**Describe alternatives you've considered**
An alternative provided in the thread was to provide read offsets for different nodes so they don't all try to read on the most recent jobs. This works too, but might be dangerous if a node crashes and the range that was reserved for it is not read anymore.

javahippie:
Preview Feature for the next Milestone release, not to be a part of a final release, yet.

Fixes #264


javahippie:
Additional discussion here: https://forum.operaton.org/t/design-of-the-job-executor-improvement/109/2

kthoms:
I'm missing a test case, that at least checks that the SQL is containing `FOR UPDATE SKIP LOCKED`. Since h2 supports it maybe even an integration test with the feature enabled.

javahippie:
You're right and this is planned. This is still a draft and WIP

kthoms:
@javahippie This could be something for version 1.1? If so, assign the milestone. Branch `main` is for version 1.1 already and people could start testing.

andreybpanfilov:
We are also sporadically facing with issues described in [Job-execution-rejected](https://forum.operaton.org/t/job-execution-rejected/68), unfortunately, I'm afraid this approach won't work as expected. The are two main reasons:

1. `FOR UPDATE SKIP LOCKED` feature is extremely "naughty", and it's behaviour depends on both DB engine and execution plan, please check for example:
   - [How many row should lock when using SELECT .. FOR UPDATE LIMIT 1](https://stackoverflow.com/questions/76098834/how-many-row-should-lock-when-using-select-for-update-limit-1)
   - [LIMIT 1 not respected when used with FOR UPDATE SKIP LOCKED in Postgres 14](https://stackoverflow.com/questions/75755863/limit-1-not-respected-when-used-with-for-update-skip-locked-in-postgres-14)
   - [FOR UPDATE SKIP LOCKED with ROWNUM on ORACLE 12c](https://stackoverflow.com/questions/52121693/for-update-skip-locked-with-rownum-on-oracle-12c)
2. `JobManager` must not acquire `exclusive` tasks which are acquired or being executed by another instances of `JobManager`, in turn this means if you want `JobManager` to be capable to run `AcquireJobsCmd` command in parallel, you need also to recheck what tasks have been acquired by competing instances, and somehow do not consider or remove exclusive duplicates from execution queue (yes, original `camunda` code seems to be broken from this perspective) - I believe serial execution of `AcquireJobsCmd` would perform much better (and simpler)

javahippie:
Hi, @andreybpanfilov, thanks for your comments, I really appreciate your insights!

Your remarks make a lot of sense – this feature is finicky and can find edge cases which might not behave the way we like. I think your thoughts about serial execution of the AcquireJobsCmd. I will keep exploring if the FOR UPDATE SKIP LOCKED approach can be built in a robust way, but I'd also like to evaluate your proposal. I promise you that this feature won't be rushed, would at most(!) be an opt-in feature users need to actively enable and that we will scrap this idea if during testing we find out that it won't work.

javahippie:
@kthoms making progress on this. Queries, Indexes seem to be robust and I think I managed to navigate around all the caveats of skip locked ... for every vendor so far (minor verifications for Oracle, because Rownum is ... well, how it is).

One question arises: You can globally configure the job executor to work by priority first, not by due date. I have never seen this flag activated, and I think very few people use this.

For the locking to work properly (and not too broad) on MySQL and MariaDB, I am adding a new index. This index does not work, if the job executor works by priority. Without an index, concurrent nodes reading from the table could lock too many entries, leading for some of the nodes to possibly get empty results and run a job executor cycle without doing anything. There are now three options:

    Document that combining the skip locked behavior and priority ordering does not give you a performance boost on MySQL and MariaDB (weird, concerning for readers)
    Adding a proper index, so that the Job Executor can also properly lock for priority ordering. Might impact write performance, needs a benchmark, I believe
    Document that users preferring priority ordering should create the missing index themselves. In that case, the impact on anybody else is avoided, but they can optimize the engine for their preferences

It's too late in the evening for me to have a preference here, would be happy about a short discussion in the upcoming days

javahippie:
Ran the integration tests. Same flakey tests failed on multiple vendors (I think they were introduced when Thread.sleep() was replaced with awaitility. Will create an issue and fix it in a different PR.

What's now missing: Agreement on the behavior and possibly some configuration guards. If we want preview features, this would be a prime candidate

kthoms (approved):

PR #340, opened 2024-12-16
Chore/set project encoding

WoSchmo:
I had problems on my local machine because there are cyrillic characters in the source code as well as in generated code, which caused maven to constantly fail. Setting the encoding fixed it.


kthoms (approved):
Thanks, @WoSchmo. This resolves also build warnings of kind
```
Warning:  Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
```

PR #344, opened 2024-12-17
chore: Remove unnecessary throws declarations

kthoms:
related to #88


kthoms:
This is a large change set. I have looked at the individual changes already.

PR #345, opened 2024-12-18
skip maven install on windows machines

WoSchmo:
On my Windows machine I had problems with the install plugin in the changed project poms. This solves that.

### DIFF
--- a/distro/run/qa/integration-tests/pom.xml
+++ b/distro/run/qa/integration-tests/pom.xml
@@ -130,5 +130,32 @@
     </plugins>
   </build>
 
+  <profiles>
+    <profile>
+      <id>windows</id>
+      <activation>
+        <os>
+          <family>windows</family>
+        </os>
+      </activation>
+      <properties>
+        <os.type>win32</os.type>
+      </properties>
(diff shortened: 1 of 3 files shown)


javahippie:
Hi @WoSchmo thanks for the report and PR! 

Could you elaborate which errors you saw on Windows? The install phase is pretty late in the Maven Lifecycle, so I wonder if `mvn verify` would be enough as a workaround, and we should instead try and make the install plugin work on Windows, so developers are able to install their dev version into the local repository?

kthoms:
I think we should know about the root cause. Disabling install does not sound as a solution. We had cases where specific modules needed to be skipped, but this should be done now. Recently the "run" distro had been added which caused such a problem, but this got fixed with https://github.com/operaton/operaton/pull/343.

Please report a bug with the error that you are experiencing. And try on latest `main`, please.

WoSchmo:
OK, I tried it again with the current main and it works. Thank you @kthoms for the update. I will close the MR! Sorry for the confusion!

kthoms:
Good to hear that it is working!

PR #349, opened 2024-12-18
chore(build): Use JReleaser for nightly releases

### LINKED ISSUE #347: Use JReleaser in nightly builds
Operaton uses JReleaser already for publishing releases. The Nightly Build distributes artifacts as build artifacts attached to the build job. The artifacts are downloaded as zip, which itself contains a tar.gz. 

Use JReleaser also in the Nightly Build workflow. It should create a prelimary release with attached artifacts.
Maven Central deployment must not happen, since it does not accept snapshot builds. But JReleaser seems to consider this already.
Consider attaching build reports as part of the distributed files.

kthoms:
This change makes the Nightly Build workflow use JReleaser to produce a preliminary release. The Nightly Build is now more aligned to the Release workflow.

For manual runs, the Nightly Build allows to specify the Java versions to build against.

Project Reports are attached as zip file to the release artifacts.

jreleaser.yml:
- include 'run' distro as artifacts
- package project reports as zip and attach as release artifacts
- set previousTagName: 1.0.0-beta-2

nightly-build.yml:
- use jacoco-create-flag-files.sh
- add `release` job
- include 'distro-run'

release.yml:
- set JRELEASER_PRERELEASE_ENABLED: false
- include 'distro-run'

prepare-reports.sh:
- package reports as zip

jacoco-create-flag-files.sh:
- extract shell script from nightly-build.yml

build.yml:
- use jacoco-create-flag-files.sh

closes #347

### DIFF
--- a/.github/jreleaser/changelog.tpl
+++ b/.github/jreleaser/changelog.tpl
@@ -4,13 +4,6 @@
 
 {{changelogNewAndNoteworthy}}
 
-### Breaking Changes
-
-- Operaton removed the compatibility layer for Activiti. If you need to use Activiti models you will have to convert them (see the following [blog post](https://camunda.com/blog/2016/10/migrate-from-activiti-to-camunda/) for details).
-- The support for the `javax.el` expression language has been removed. Application servers that ship this library are
-incompatible with Operaton. If you are running Operaton on an application server, make sure that it supports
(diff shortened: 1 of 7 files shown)


kthoms:
JReleaser now produces an "early-access" release:
https://github.com/operaton/operaton/releases/tag/early-access

Screenshot 337

Issue #144, opened 2024-11-05
Resolve usage of deprecated API


kthoms:
**Is your feature request related to a problem? Please describe.**
The code base is full of usage of deprecated API. Often there is a recommendation which API should be used instead.
The code base should be cleaned up to avoid further usage of such API.

**Describe the solution you'd like**
Scan for deprecated API usage. This can be done by reviewing build warnings or use means of the IDE (e.g. IJ: Code/Inspect Code).

For each usage pattern raise a PR.

When the used deprecated API was in from the Operaton codebase and there is no usage anymore then extend the `@Deprecated` annotation with `forRemoval=true`.

PR #361, opened 2024-12-20
chore: resolve deprecation org.junit.Assert.assertThat

cmoellerherm:
- replaced usages by org.hamcrest.MatcherAssert.assertThat;

related to #144


kthoms (approved):
In general we try to get rid off Hamcrest in favor of AssertJ, but this would be another refactoring. As an intermediate step this is OK, and it is easier to replace that with AssertJ's `assertThat(...).isEqualTo(...)` with Open Rewrite in a next action.
You could do this on your branch, but I would be fine to merge this.

kthoms:
Thinking of it, let's take this for the moment and do Hamcrest -> AssertJ next.

PR #367, opened 2024-12-22
chore(engine): add debug logs to improve job acquisition job execution logging

hauptmedia:
* add test cases for the job-acquisition and job-execution logs

related https://github.com/camunda/camunda-bpm-platform/issues/2666

Backported commit 824b35f3f2 from the camunda-bpm-platform repository.
Original author: Palanisamy, Prakash <prakash.palanisamy@fmr.com>

### DIFF
--- a/engine-spring/core/src/main/java/org/operaton/bpm/engine/spring/components/jobexecutor/SpringJobExecutor.java
+++ b/engine-spring/core/src/main/java/org/operaton/bpm/engine/spring/components/jobexecutor/SpringJobExecutor.java
@@ -22,6 +22,7 @@ import java.util.concurrent.RejectedExecutionException;
 import org.operaton.bpm.engine.impl.ProcessEngineImpl;
 import org.operaton.bpm.engine.impl.jobexecutor.JobExecutor;
 import org.springframework.core.task.TaskExecutor;
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor;
 
 /**
  *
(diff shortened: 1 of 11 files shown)


engine-spring/core/src/main/java/org/operaton/bpm/engine/spring/components/jobexecutor/SpringJobExecutor.java
```diff
@@ -59,10 +60,16 @@ public void executeJobs(List<String> jobIds, ProcessEngineImpl processEngine) {
 	  try {
       taskExecutor.execute(getExecuteJobsRunnable(jobIds, processEngine));
     } catch (RejectedExecutionException e) {
-
       logRejectedExecution(processEngine, jobIds.size());
       rejectedJobsHandler.jobsRejected(jobIds, processEngine, this);
-    }
+		} finally {
+			if (taskExecutor instanceof ThreadPoolTaskExecutor) {
```
javahippie, line 66: We have refactored "instanceof" checks to use the "new" style with direct assignment, I'd prefer if we did this in the backports directly, to not import this old style into our repo again

javahippie (changes requested):

engine/src/main/java/org/operaton/bpm/engine/impl/jobexecutor/RuntimeContainerJobExecutor.java
```diff
@@ -63,6 +64,15 @@ public void executeJobs(List<String> jobIds, ProcessEngineImpl processEngine) {
       logRejectedExecution(processEngine, jobIds.size());
       rejectedJobsHandler.jobsRejected(jobIds, processEngine, this);
     }
+
+    if (executorService instanceof JmxManagedThreadPool) {
```
javahippie, line 68: Same here with the newer instanceof style, sorry for not catching it in the first review

kthoms (approved):

PR #369, opened 2024-12-22
fix(asyncInstanceModification): respect ancestor during startBefore/after cmd execution (#4829)

hauptmedia:
related to camunda/camunda-bpm-platform#4507

Backported commit 7bf5c46119 from the camunda-bpm-platform repository.
Original author: Helene <21290204+PHWaechtler@users.noreply.github.com>

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/json/ModificationCmdJsonConverter.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/json/ModificationCmdJsonConverter.java
@@ -16,6 +16,7 @@
  */
 package org.operaton.bpm.engine.impl.json;
 
+import com.google.gson.JsonObject;
 import org.operaton.bpm.engine.impl.cmd.AbstractProcessInstanceModificationCommand;
 import org.operaton.bpm.engine.impl.cmd.ActivityAfterInstantiationCmd;
 import org.operaton.bpm.engine.impl.cmd.ActivityBeforeInstantiationCmd;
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/json/ModificationCmdJsonConverter.java
```diff
@@ -38,33 +38,37 @@ public class ModificationCmdJsonConverter extends JsonObjectConverter<AbstractPr
...
-      JsonUtil.addField(json, CANCEL_CURRENT, cancellationCommand.isCancelCurrentActiveActivityInstances());
-    }
-    else if (command instanceof ActivityInstanceCancellationCmd cancellationCommand) {
-      JsonUtil.addField(json, CANCEL_ACTIVITY_INSTANCES, cancellationCommand.getActivityInstanceId());
-      JsonUtil.addField(json, PROCESS_INSTANCE, cancellationCommand.getProcessInstanceId());
-    }
-    else if (command instanceof TransitionInstanceCancellationCmd canellationCommand) {
-      JsonUtil.addField(json, CANCEL_TRANSITION_INSTANCES, canellationCommand.getTransitionInstanceId());
-      JsonUtil.addField(json, PROCESS_INSTANCE, canellationCommand.getProcessInstanceId());
+    if (command instanceof ActivityAfterInstantiationCmd) {
```
javahippie, line 47: We have refactored "instanceof" checks to use the "new" style with direct assignment, I'd prefer if we did this in the backports directly, to not import this old style into our repo again

javahippie (changes requested):
Please take a look at the `instanceof` checks

kthoms:
I have fixed the findings and cleaned up the PR

javahippie (approved):

PR #374, opened 2024-12-25
chore(engine): Resolve & document deprecations in TransitionImpl, ScopeImpl

kthoms:
Methods in `TransitionImpl` & `ScopeImpl` regarding execution listeners have been deprecated. Documented in Javadoc their replacement, resolved usages of the deprecated methods and marked those methods for removal.

related to #144

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/pvm/process/TransitionImpl.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/pvm/process/TransitionImpl.java
@@ -57,7 +57,10 @@ public class TransitionImpl extends CoreModelElement implements PvmTransition {
     destination.getIncomingTransitions().add(this);
   }
 
-  @Deprecated
+  /**
+   * @deprecated Use {@link #addListener(ExecutionListener.EVENTNAME_TAKE, executionListener)} instead
+   */
+  @Deprecated(forRemoval = true)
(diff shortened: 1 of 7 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/pvm/process/TransitionImpl.java
```diff
@@ -57,21 +57,30 @@ public void setDestination(ActivityImpl destination) {
...
   public void addExecutionListener(ExecutionListener executionListener) {
     super.addListener(ExecutionListener.EVENTNAME_TAKE, executionListener);
   }
 
+  /**
+   * @deprecated Use {@link #getListeners(ExecutionListener.EVENTNAME_TAKE)} instead
+   */
   @SuppressWarnings({ "rawtypes", "unchecked" })
   @Deprecated
   public List<ExecutionListener> getExecutionListeners() {
```
javahippie, line 73: What is the reason to not mark this method `forRemoval=true`?

kthoms: I missed that, good catch

kthoms: done

Issue #167, opened 2024-11-07
chore(deps): Revise unused dependencies


kthoms:
**Is your feature request related to a problem? Please describe.**
Modules have dependencies that are not used.

**Describe the solution you'd like**
Look for declared but unused dependencies.

**Additional context**
Run
```
./mvnw dependency:analyze-report 
```
to get reports about dependencies per module. Relevant is the section "Unused but Declared Dependencies"

![Bildschirmfoto 2024-11-07 um 13 59 01](https://github.com/user-attachments/assets/fbc88788-3fb3-4213-ba3c-6b87fb3fa9df)

WoSchmo:
I am working on it. Someone can assign it to me.

javahippie:
Done, thanks a lot!

javahippie:
A hint, be aware of the limitations of this plugin: https://maven.apache.org/plugins/maven-dependency-plugin/faq.html#unused

WoSchmo:
Thank you for the hint! I will take this into account. If I am done I would appreciate a thourough review of someone with camunda knowledge, because some deps don't show directly in the code like the spring boot actuator.

PR #403, opened 2025-01-06
chore: remove unused dependencies

WoSchmo:
I have removed several dependencies. Please take your time and review if those are not neccesary otherwise!

related to #167

### DIFF
--- a/clients/java/client/pom.xml
+++ b/clients/java/client/pom.xml
@@ -78,19 +78,8 @@
     </dependency>
 
     <dependency>
-      <groupId>org.slf4j</groupId>
-      <artifactId>jcl-over-slf4j</artifactId>
-      <scope>test</scope>
-    </dependency>
-
-    <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.junit.jupiter</groupId>
-      <artifactId>junit-jupiter</artifactId>
       <scope>test</scope>
     </dependency>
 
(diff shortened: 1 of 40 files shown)


kthoms:
Thanks @WoSchmo. I will need a while to check this. I'm a bit unsure regarding dependencies that are depending on distributions. The default build does not check all profiles and engine integration tests are not yet enabled.

On first sight, it looks ok.

kthoms:
This was a bit too aggressive. The "run" distro for example fails with
```
Caused by: java.io.FileNotFoundException: class path resource [org/springframework/web/servlet/config/annotation/WebMvcConfigurer.class] cannot be opened because it does not exist
```

To verify, build and run the distros and see if at least tasklist is working as expected.

clients/java/client/pom.xml
```diff
@@ -77,22 +77,11 @@
       <artifactId>jackson-databind</artifactId>
     </dependency>
 
-    <dependency>
```
kthoms, line 80: @WoSchmo Perhaps you could leave your branch locally and provide smaller PRs that remove just certain dependencies. 
It should be rather safe to remove unused test dependencies. 
Then the logging dependencies. 
The Apache Commons dependencies should be also rather safe.
Inter-project dependencies have to be tested one by one.

kthoms (changes requested):

clients/java/client/pom.xml
WoSchmo: Yes, I think it's best to remove this PR and start over again.

javahippie:
@kthoms @WoSchmo Just checking if the latest comment still stands and this PR should be closed without merging it?

WoSchmo:
Thank you for the reminder!

PR #405, opened 2025-01-07
refactor(engine): changes multiline strings to text blocks

CBorowski-dev:
- changes multiline strings to text blocks

related to #85


kthoms:
@CBorowski-dev the change is ok. You could consider to combine changes also for multiple modules. It is not necessary to have a single change per PR.

kthoms (approved):

PR #409, opened 2025-01-07
refactor: Use StringUtils#hasText(), ObjectUtils#isEmpty, String#isEmpty

### LINKED ISSUE #410: Deprecate `StringUtil#hasText()`
The method `org.operaton.bpm.engine.impl.util.StringUtil#hasText` should be deprecated in favor of `org.springframework.util.StringUtils#hasText(java.lang.String)`.
There is only 1 known usage in the code base.

kthoms:
resolve Sonar findings of type [java:7158](https://sonarcloud.io/project/issues?rules=java%3AS7158&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton)

- use StringUtils#hasText() when checking for non-null and not-empty
- use ObjectUtils#isEmpty() when checking for null or empty
- use String#isEmpty() when the checked string is guaranteed non-null and only empty/not-empty is checked

Deprecate `org.operaton.bpm.engine.impl.util.StringUtil#hasText(String)`

related to #88
closes #410

### DIFF
--- a/engine-cdi/core/src/main/java/org/operaton/bpm/engine/cdi/ProcessVariables.java
+++ b/engine-cdi/core/src/main/java/org/operaton/bpm/engine/cdi/ProcessVariables.java
@@ -50,7 +50,7 @@ public class ProcessVariables {
 
   protected String getVariableName(InjectionPoint ip) {
     String variableName = ip.getAnnotated().getAnnotation(ProcessVariable.class).value();
-    if (variableName.length() == 0) {
+    if (variableName.isEmpty()) {
       variableName = ip.getMember().getName();
     }
     return variableName;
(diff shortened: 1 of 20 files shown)


kthoms:
It is quite annoying that Sonar detects code as "newly" duplicated which has been there forever.

The code in question is actually OK that it is duplicated. `AbstractTransactionProcessEngineConfiguration` declares a method `initTransactionManager()` which is implemented by `JtaProcessEngineConfiguration` and `JakartaTransactionProcessEngineConfiguration` in the same way. But that is not a reason to move the implementation to the base class, or to introduce another common base class just to implement the method for both classes in one base class.

I think we should leave the code as is and accept the Quality Gate failure here.

javahippie (approved):

PR #411, opened 2025-01-08
security: Sanitize message parameters to avoid injection attacks

kthoms:
also: add isTraceEnabled() for completeness and use is<Level>Enabled() methods to avoid having unused methods in the code (isWarnEnabled & isErrorEnabled were not used).

### DIFF
--- a/commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
+++ b/commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
@@ -75,7 +75,7 @@ public abstract class BaseLogger {
   /** the project code of the logger */
   protected String projectCode;
 
-  /** the component Id of the logger. */
+  /** the component id of the logger. */
   protected String componentId;
 
   protected BaseLogger() {
@@ -153,9 +153,9 @@ public abstract class BaseLogger {
    * @param parameters      a list of optional parameters
    */
   protected void logTrace(String id, String messageTemplate, Object... parameters) {
-    if (delegateLogger.isTraceEnabled()) {
+    if (isTraceEnabled()) {
       String msg = formatMessageTemplate(id, messageTemplate);
-      delegateLogger.trace(msg, parameters);
+      delegateLogger.trace(msg, sanitizeParameters(parameters));
     }
   }
 
@@ -167,9 +167,9 @@ public abstract class BaseLogger {
    * @param parameters a list of optional parameters
    */
   protected void logDebug(String id, String messageTemplate, Object... parameters) {
-    if(delegateLogger.isDebugEnabled()) {
+    if(isDebugEnabled()) {
       String msg = formatMessageTemplate(id, messageTemplate);
-      delegateLogger.debug(msg, parameters);
+      delegateLogger.debug(msg, sanitizeParameters(parameters));
     }
   }
 
@@ -181,9 +181,9 @@ public abstract class BaseLogger {
    * @param parameters a list of optional parameters
    */
   protected void logInfo(String id, String messageTemplate, Object... parameters) {
-    if(delegateLogger.isInfoEnabled()) {
+    if (isInfoEnabled()) {
       String msg = formatMessageTemplate(id, messageTemplate);
-      delegateLogger.info(msg, parameters);
+      delegateLogger.info(msg, sanitizeParameters(parameters));
     }
   }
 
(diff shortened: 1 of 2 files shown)


kthoms:
Adresses [this Sonar security finding](https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=VULNERABILITY&id=operaton_operaton&open=AZQjDLyfoZL_XCvHvyO8)
Screenshot 352

commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
```diff
@@ -276,4 +281,25 @@ protected String exceptionMessage(String id, String messageTemplate, Object... p
...
+    }
+    Object[] sanitized = new Object[parameters.length];
+    for (int i = 0; i < parameters.length; i++) {
+      sanitized[i] = sanitize(parameters[i]);
+    }
+    return sanitized;
+  }
+
+  private Object sanitize(Object parameter) {
+    if (parameter instanceof String) {
```
javahippie, line 299: In new code we should use variable reassignement in instanceof itself instead of casting.

commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
```diff
@@ -276,4 +281,25 @@ protected String exceptionMessage(String id, String messageTemplate, Object... p
     }
   }
 
+  /**
+   *  Sanitize parameters to avoid injection attacks
+   */
+  private Object[] sanitizeParameters(Object... parameters) {
+    if (parameters == null) {
+      return new Object[0];
```
javahippie, line 289: This is modifying the behavior. Before this change, null was passed to the logger, now it is an empty array. While null should not be passed on to the logger, I don't see any reason why this change should be the responsibility of `sanitizeParameters`.

commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
```diff
@@ -276,4 +281,25 @@ protected String exceptionMessage(String id, String messageTemplate, Object... p
...
+    Object[] sanitized = new Object[parameters.length];
+    for (int i = 0; i < parameters.length; i++) {
+      sanitized[i] = sanitize(parameters[i]);
+    }
+    return sanitized;
+  }
+
+  private Object sanitize(Object parameter) {
+    if (parameter instanceof String) {
+      return ((String) parameter).replaceAll("[\n\r\t]", "_");
```
javahippie, line 300: How does sanitizing Newline, Carriage Return and Tab fix security issues? Which vulnerabilities are prevented by this?

javahippie (changes requested):

commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
kthoms: I was just following the [recommendation from Sonar](https://sonarcloud.io/project/issues?severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=VULNERABILITY&pullRequest=411&id=operaton_operaton&open=AZRHr8XIaj5AKEHLTXFF&tab=how_to_fix).

kthoms: Are you aware of some library function that could help here?

javahippie: I dug a little bit on Google about this. What is meant apparently, is that an attacker could smuggle newlines into our logs to lay a false trail while doing something malicious, so banning these characters would prevent this. The suggested fix by Sonar addresses this "log forging".

XSS or Log Injection with variables like log4shell should be handled by the logging framework, so this seems to be fine, and the fix would make sense.

javahippie: For future discussion (not on this issue, maybe): https://cheatsheetseries.owasp.org/cheatsheets/Logging_Cheat_Sheet.html

commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
kthoms: done

commons/logging/src/main/java/org/operaton/commons/logging/BaseLogger.java
kthoms: good catch. I have changed that to return `null` and added a test for this method. However, it leaves a warning that an empty array should be returned instead of null. Added a suppression for it.

javahippie (approved):
lgtm, can be merged (not merging right now, because checks for the previous merge are still ongoing)

PR #412, opened 2025-01-09
refactor(tests): Testing exceptions with only one possible invocation throwing

kthoms:
Refactor tests that check for exceptions to have only a single call in the calling code that could throw a RuntimeException.

Refactor to use `assertThatThrownBy()` for checking code for exceptions thrown.

Minor code cleanups: Removing unthrown exceptions, removed unused variable assignments etc.

related to #88, #27

### DIFF
--- a/commons/typed-values/src/test/java/org/operaton/bpm/engine/test/variables/FileValueTypeImplTest.java
+++ b/commons/typed-values/src/test/java/org/operaton/bpm/engine/test/variables/FileValueTypeImplTest.java
@@ -16,8 +16,6 @@
  */
 package org.operaton.bpm.engine.test.variables;
 
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Test;
 import org.operaton.bpm.engine.variable.Variables;
 import org.operaton.bpm.engine.variable.impl.type.FileValueTypeImpl;
 import org.operaton.bpm.engine.variable.value.FileValue;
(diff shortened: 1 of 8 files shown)


commons/typed-values/src/test/java/org/operaton/bpm/engine/test/variables/FileValueTypeImplTest.java
```diff
@@ -174,15 +178,17 @@ void createValueWithNullProperties() {
   @Test
   void cannotCreateFileWithoutName() {
     InputStream file = this.getClass().getClassLoader().getResourceAsStream("org/operaton/bpm/engine/test/variables/simpleFile.txt");
-    assertThrows(IllegalArgumentException.class, () ->
-      type.createValue(file, Collections.<String, Object>emptyMap()));
+    Map<String, Object> valueInfo = emptyMap();
```
javahippie, line 181: Maybe "emptyValueInfo" to highlight, that *this* is the variable which lets the test fail? Not a must, though, just an opinion

javahippie (approved):

commons/typed-values/src/test/java/org/operaton/bpm/engine/test/variables/FileValueTypeImplTest.java
kthoms: Thx, changed that.

PR #415, opened 2025-01-09
refactor(tests): Refactor ProcessInstanceAuthorizationTest

kthoms:
Resolve: Testing exceptions with only one possible invocation throwing

Refactor tests that check for exceptions to have only a single call in the calling code that could throw a RuntimeException.

Refactor to use `assertThatThrownBy()` for checking code for exceptions thrown.

Using Copilot to refactor code. Using prompt:
"Use assertThatThrownBy
 Preserve line comments
 Place „// when“ above assertThatThrownBy
 Place „// then“ before the first assertion
 Remember that the failure message is the first assertion
 Use hasMessageContaining as replacement for assertTextPresent
"

related to #88, #27


javahippie (approved):
lgtm

PR #421, opened 2025-01-11
chore(build): Prepare deployment of sql-scripts

kthoms:
In order to activate the sql-checks the jar with the sql scripts has to be deployed.

The versioning of the database differs from the project version. While in C7 this was in line, Operaton starts with version 1.0 while the database starts with 7.23.

To reflect this the version of the artifact `org.operaton.bpm.distro:operaton-sql-scripts` starts with version 7.23.0.

This leads to adjustments in the POM: The build must not consume Operaton artifacts (esp. 'engine' which holds the DB scripts) with the `project.version` (set to 7.23.0), but with the parent's version (currently 1.0.0-beta-3).

In order to make the 'check-sql' profile working, the version 7.22.0 of this artifact must be deployed first.


distro/sql-script/pom.xml
```diff
@@ -10,6 +10,7 @@
 
   <groupId>org.operaton.bpm.distro</groupId>
   <artifactId>operaton-sql-scripts</artifactId>
+  <version>7.23.0</version>
```
kthoms, line 13: this version must be updated by `set-project-version.sh`

PR #429, opened 2025-01-13
refactor: Make logger instances final

kthoms:
Logger fields that are named upper case have been made final.

Resolve Sonar findings of type: java:S3008

related to #88


javahippie:
I think this is fine, if somebody switched out the static loggers at runtime, I think there is no way to help them

PR #431, opened 2025-01-13
refactor: Replace this "Map.get()" and condition with a call to "Map.computeIfAbsent()"

kthoms:
Resolve Sonar findings of type: java:S3824

related to #88


kthoms:
The Sonar issue is about `BpmnParse#parseSequenceFlow` being a "brain" method. This is out of the scope of this PR to refactor that.

PR #434, opened 2025-01-14
refactor: Resolve deprecation of Class#newInstance

kthoms:
Replace Class#newInstance() by Class#getDeclaredConstructor().newInstance()

Catch additional exceptions and collapse exceptions catches with equal body.

related to #144


kthoms:
Code duplication in DTOs is normal

Issue #427, opened 2025-01-13
Local build fails on Windows 11: NullPointerException: Cannot invoke "String.replace(java.lang.CharSequence, java.lang.CharSequence)"


ArneDeutsch:
**Describe the bug**
When executing a full maven build locally on my **Windows 11** Laptop the build fails consistenly with test failures while building **Operaton - Run - Core**. The following exception is shown many times:

```
[ERROR] org.operaton.bpm.run.test.plugins.OperatonRunProcessEnginePluginsRegistrationTest.shouldInitializeRegisteredPlugins -- Time elapsed: 0.002 s <<< ERROR!
[...]

PR #453, opened 2025-01-22
Fix NPE when running integration tests locally.

ArneDeutsch:
This fixes https://github.com/operaton/operaton/issues/427 .

Analyse: The issue arises if _operaton.deploymentDir_ is _null_ because in _OperatonBpmRunDeploymentConfiguration.getNormalizedDeploymentDir_ the value is accessed without a check. In a regular installation the property is set in the _run.bat_/_run.sh_ scripts. But these are not used for the integration tests. Instead the deafult is taken from _OperatonBpmRunConfiguration.operatonDeploymentConfiguration_. Here the value is configured to be _null_.

```
  @Bean
  public OperatonBpmRunDeploymentConfiguration operatonDeploymentConfiguration(@Value("${operaton.deploymentDir:#{null}}") String deploymentDir) {
    return new OperatonBpmRunDeploymentConfiguration(deploymentDir);
  }

```

Hence _null_ is a valid value in test scenarious and code should check for null.


kthoms (approved):

kthoms:
Thank you, Arne!

PR #455, opened 2025-01-22
Fix Sonar finding by converting class to record.

ArneDeutsch:
Part of https://github.com/operaton/operaton/issues/85 .


model-api/xml-model/src/main/java/org/operaton/bpm/model/xml/impl/instance/ModelTypeInstanceContext.java
```diff
@@ -25,17 +25,8 @@
...
-  private final ModelInstanceImpl model;
-  private final DomElement domElement;
-  private final ModelElementTypeImpl modelType;
-
-  public ModelTypeInstanceContext(DomElement domElement, ModelInstanceImpl model, ModelElementTypeImpl modelType) {
-    this.domElement = domElement;
-    this.model = model;
-    this.modelType = modelType;
-  }
+public final record ModelTypeInstanceContext(DomElement domElement, ModelInstanceImpl model,
```
kthoms, line 28: final should be obsolete, records can't be extended

kthoms (changes requested):
final should be obsolete, records can't be extended

javahippie:
I know that this was done to ensure backwards compatibility, but getters on records are not a thing we should do, imho. The record also has accessor methods by default (which should be used in idiomatic java) and now we have both `domElement()` and `getDomElement()` on the classes public API. Is it possible to refactor to use the accessors without breaking anything?

ArneDeutsch (commented):
Remove "final", removed getters, call record methods instead.

ArneDeutsch:
Mixed up merge requests.

PR #457, opened 2025-01-23
Record class refactoring

ArneDeutsch:
Refactore class to record (with correct "final" and removed getters).


javahippie:
Thanks, @ArneDeutsch! 

Before merging, I adjusted the commit messages according to our conventions: https://github.com/operaton/operaton/blob/main/CONTRIBUTING.md#commit-message-conventions

Issue #32, opened 2024-10-23
Remove the whole concept of a "License Key"


javahippie:
**Is your feature request related to a problem? Please describe.**
Camunda provided a proprietary version which is based on the OSS version. There are references to this license key in the codebase, but we don't need it in a pure OSS project and don't want to confuse users

**Describe the solution you'd like**
We need to carefully remove the references in the data model, without breaking any API (Java or REST).

hauptmedia:
I think this was already removed in the webapps that might have called these functions. Are the any references to this functions besides from the REST Api?

ArneDeutsch:
When I do a full text search for "licenseKey" on the code base I find:

- A lot of java code in **operaton-engine** for the management of license keys. To my understanding this is not needed and should be deleted.
- Some java and template code in **operaton-engine-rest-core** and **operaton-engine-rest-openapi**. Do we really need to not touch this REST API?
- A tiny bit of JS code in **operaton-webapp-root**. Semms to be administrator functionality to handle license keys. Remove?

hauptmedia:
I think it should be deleted everywhere. The concept of the license key was only used for the Enterprise Edition which was based on the Community Edition. WDYT @javahippie @a-m-zill

javahippie:
I think that the license key HTTP methods were mostly called by the Cockpit, because in no other context there was a difference between "Community Edition" and "Enterprise Edition". If there are still references to this in the cockpit, they are leftovers and can be removed imho.

It is part of the /telemetry endpoint which was already deprecated by Camunda. Inside this response, the license key is part of the internals DTO. I agree that this is to be considered a breaking change, but imho this will not impact any existing Camunda installations after the migration: https://docs.camunda.org/rest/camunda-bpm-platform/7.23-SNAPSHOT/#tag/Telemetry/operation/getTelemetryData

ArneDeutsch:
I will prepare a commit.

ArneDeutsch:
I find code like this and wonder if it is save to remove it, because existing Databases might have an issue. I have no idea what the "byte array table" is.

**PurgeDatabaseAndCacheCmd.java**
```
          // allow License Key in byte array table
          if (tableNameWithoutPrefix.equals("ACT_GE_BYTEARRAY") && commandContext.getResourceManager().findLicenseKeyResource() != null) {
            if (count != 1) {
[...]

PR #458, opened 2025-01-23
Remove all license key code.

ArneDeutsch:
Remove all code related to license key management. Also remove all tests for this code.

https://github.com/operaton/operaton/issues/32


engine/src/main/resources/org/operaton/bpm/engine/impl/mapping/entity/Resource.xml
```diff
@@ -58,11 +58,8 @@
     delete from ${prefix}ACT_GE_BYTEARRAY where DEPLOYMENT_ID_ = #{id}
   </delete>
 
-  <delete id="purgeTablePreserveLicenseKey" parameterType="string">
+  <delete id="purgeTable" parameterType="string">
```
a-m-zill, line 61: Here is a complete new functionality now. This will purge the whole ACT_GE_ByteArray Table. Is this something which we want?

@javahippie can you say what do you think here?

javahippie: I think this functionality purged the whole table before, too, but left the rows which stored a license key in place, didn't it?

a-m-zill: Ok, got it. So we can let it like it is now. :)

javahippie (approved):
lgtm!

kthoms (approved):

PR #466, opened 2025-01-27
build: Upgrade openapi-generator-maven-plugin to 7.11.0.

ArneDeutsch:
Upgrade openapi-generator-maven-plugin to 7.11.0, the latest avaiable version. Also upgrade dependencies accordingly.

related to #131

Issue #354, opened 2024-12-19
Upgrade & consolidate Maven Plugins


kthoms:
**Is your feature request related to a problem? Please describe.**
Some Maven plugins have updates. Some are used in multiple versions.

Status | Group Id | Artifact Id | Current Version | Latest Subincremental | Latest Incremental | Latest Minor | Latest Major | Dependency status
-- | -- | -- | -- | -- | -- | -- | -- | --
  | com.googlecode.maven-download-plugin | download-maven-plugin | 1.11.1 |   |   | 1.13.0 |   |  
  | io.quarkus | quarkus-extension-maven-plugin | 3.15.0 |   | 3.15.2 | 3.17.4 |   |  
  | org.apache.felix | maven-bundle-plugin | 5.1.2 |   | 5.1.9 |   | 6.0.0 |  
  | org.apache.maven.plugins | maven-failsafe-plugin | 3.5.1 |   | 3.5.2 |   |   |  
  | org.apache.maven.plugins | maven-javadoc-plugin | 3.10.1 |   |   | 3.11.2 |   |  
  | org.apache.maven.plugins | maven-release-plugin | 3.0.1 |   |   | 3.1.1 |   |  
  | org.apache.maven.plugins | maven-resources-plugin | 2.6 |   |   | 2.7 | 3.3.1 |  
  | org.apache.maven.plugins | maven-resources-plugin | 3.3.1 |   |   |   |   |  
  | org.apache.maven.plugins | maven-surefire-plugin | 3.5.1 |   | 3.5.2 |   |   |  
  | org.codehaus.cargo | cargo-maven3-plugin | 1.10.15 |   | 1.10.16 |   |   |  
  | org.codehaus.mojo | license-maven-plugin | 2.4.0 |   |   | 2.5.0 |   |  
  | org.jboss.shrinkwrap.resolver | shrinkwrap-resolver-maven-plugin | ${version.shrinkwrap.resolvers} |   |   |   | 3.3.3 |  
  | org.jboss.shrinkwrap.resolver | shrinkwrap-resolver-maven-plugin | 2.2.2 |   | 2.2.7 |   | 3.3.3 |  
  | org.liquibase | liquibase-maven-plugin | 4.29.2 |   |   | 4.30.0 |   |  
  | org.springframework.boot | spring-boot-maven-plugin | 3.3.5 |   | 3.3.6 | 3.4.0 |   |  
[...]

PR #467, opened 2025-01-29
chore: Update maven plugins

ArneDeutsch:
- Update maven plugin versions
- Add filter for PDF and PNG files to avoid exceptions in new maven-resource-plugin (can not filter binary files)
- Adapt test code that relies on dependencies of old maven plugin versions
- Also include changes from https://github.com/operaton/operaton/pull/466 (needed, toherwise integration tests fail)

realtes to: #354


kthoms:
Did a clean build locally and checked distros Tomcat & Operaton. Both working.

Many thanks @ArneDeutsch !

PR #475, opened 2025-01-31
chore: suppress warnings for intended fallthrough to default in switch

WoSchmo:


kthoms (commented):
Nitpick: Maybe declare the suppression on class level?
Also an option: I could configure these issues just as accepted in Sonar.
WDYT? Change or leave as is?

WoSchmo:
For me both would be fine. I personally like it if it is made clear in the code, that the fallthrough is intended. But if you want, you can configure it via Sonar.

Off-topic: Here are in my opinion several issues which might be resolved via Sonar:
https://sonarcloud.io/project/issues?impactSeverities=MEDIUM&impactSoftwareQualities=MAINTAINABILITY&rules=java%3AS125&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton&open=AZLaQv1d8ZwKcH34-Ygb

kthoms (approved):

kthoms:
> Off-topic: Here are in my opinion several issues which might be resolved via Sonar:

I went through the query you gave and resolved some of those issues. A couple have to stay open.

PR #489, opened 2025-02-03
chore: improves findings reported by SonarQube

jmaster1985:
https://sonarcloud.io/project/issues?fileUuids=AZLaQmot8ZwKcH34-S4N&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParse.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParse.java
@@ -214,9 +214,11 @@ public class BpmnParse extends Parse {
   public static final String PROPERTYNAME_VARIABLE_MAPPING_DELEGATE_EXPRESSION = "variableMappingDelegateExpression";
   public static final String PROPERTYNAME_RESOURCE = "resource";
   public static final String PROPERTYNAME_LANGUAGE = "language";
+  public static final String PROPERTYNAME_DEFAULT = "default";
   public static final String TYPE = "type";
 
   public static final String TRUE = "true";
+  public static final String FALSE = "false";
   public static final String INTERRUPTING = "isInterrupting";
 
   public static final String CONDITIONAL_EVENT_DEFINITION = "conditionalEventDefinition";
@@ -227,6 +229,7 @@ public class BpmnParse extends Parse {
   public static final String MESSAGE_EVENT_DEFINITION = "messageEventDefinition";
   public static final String ERROR_EVENT_DEFINITION = "errorEventDefinition";
   public static final String CANCEL_EVENT_DEFINITION = "cancelEventDefinition";
+  public static final String TERMINATE_EVENT_DEFINITION = "terminateEventDefinition";
   public static final String LINK_EVENT_DEFINITION = "linkEventDefinition";
   public static final String CONDITION_EXPRESSION = "conditionExpression";
   public static final String CONDITION = "condition";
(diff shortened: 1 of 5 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParse.java
```diff
@@ -3272,6 +3285,8 @@ public void parseBoundaryEvents(Element parentElement, ScopeImpl flowScope) {
       if (attachedActivity == null) {
         addError("Invalid reference in boundary event. Make sure that the referenced activity is defined in the same scope as the boundary event",
             boundaryEventElement);
+
+        continue;
```
jmaster1985, line 3289: question: continue with the loop, terminate with an exception or return from the method call? in any case, this should be handled - otherwise the program can get terminated abruptly with a NullPointerException, because of accessing this null-object down below (please see line 3300) @kthoms

engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParse.java
```diff
@@ -3526,33 +3541,35 @@ protected EventSubscriptionDeclaration parseSignalEventDefinition(Element signal
     if (signalRef == null) {
       addError("signalEventDefinition does not have required property 'signalRef'", signalEventDefinitionElement, signalElementId);
       return null;
+    }
+
+    SignalDefinition signalDefinition = signals.get(resolveName(signalRef));
+    if (signalDefinition == null) {
+      addError("Could not find signal with id '" + signalRef + "'", signalEventDefinitionElement, signalElementId);
+      return null;
```
jmaster1985, line 3549: note: in the previous revision, `if (signalDefinition == null)` just logged the error and continued the execution. where at line 3556, that null object has been accessed with a method call. this could lead to a NullPointerException. in this revision, I'm returning `null`, just like the condition above with `if (signalRef == null)`. not sure whether this is the 100% correct approach. @kthoms

engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParse.java
kthoms: Hard to decide. This changes the behavior.
Line 3300 is guarded by `if (compensateEventDefinition == null)`, so it could be that `attachedActivity` is `null`, but the NPE not raised.
It looks fishy, indeed, but without having a test or an error that actually shows that this is a problem, I would leave it untouched.

engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParse.java
kthoms: Here, I would say this is fine. When `signalDefinition` is `null`, it would cause the NPE. So returning `null` looks fine to me here.

engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParse.java
kthoms: Please remove it. The other changes are fine.

kthoms: I have removed this change

kthoms (approved):

PR #501, opened 2025-02-07
chore: Fix blocker sonar issues in test code

ArneDeutsch:
Add assertions to tests lacking them where they make sense. Add `@Test(expected = None.class)` to indicate when only the absense of an exception is required.

relates to: #88


kthoms:
@ArneDeutsch Please be careful when referencing annotations in PR/issue descriptions. This tags users. The users `@test` and `@override` are often accidently linked.
You have to use backticks.

kthoms (approved):

PR #507, opened 2025-02-11
Migrate integration tests

javahippie:
1. Enables Postgres Test via the Maven Docker plugin
2. Makes sure that integration tests using postgres are executed by the failsafe plugin, not surefire (because the integration test phase has test setup


engine/pom.xml
```diff
@@ -327,6 +329,11 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
```
javahippie, line 332: @kthoms Supporting the Nashorn Engine during testing is fine for me, but adding it as a bundled dependency might be unnecessary for most of the users?

kthoms:
I tried 
```
./mvnw -Pengine-integration,wildfly,postgres -f qa
```

this fails on my machine with
```
Caused by: java.lang.IllegalArgumentException: No <dockerHost> given, no DOCKER_HOST environment variable, no read/writable '/var/run/docker.sock' or '//./pipe/docker_engine' and no external provider like Docker machine configured
    at io.fabric8.maven.docker.access.DockerConnectionDetector.detectConnectionParameter (DockerConnectionDetector.java:87)
    at io.fabric8.maven.docker.service.DockerAccessFactory.createDockerAccess (DockerAccessFactory.java:34)
    at io.fabric8.maven.docker.AbstractDockerMojo.execute (AbstractDockerMojo.java:299)
```

Here is to say that I do not have Docker Desktop installed, but use Rancher Desktop as Docker replacement. Can be that fabric8 is not prepared for it.

Issue #216, opened 2024-11-17
JUnit 5 parameter resolver for BeforeEach/AfterEach methods


kthoms:
**Is your feature request related to a problem? Please describe.**
Parametrized JUnit 4 tests (annotated with `@RunWith(Parameterized.class)`) take their parameters as part of constructor arguments. In these tests (e.g. `ContainerAuthenticationFilterTest`) these parameters are assigned to fields which are used by test methods.

With JUnit 5 it is not possible to have a `@BeforeEach`/`@AfterEach` method that takes parameters. This makes migrating these tests problematic. Each parametrized test method would need to do the setup as first action.

**Describe the solution you'd like**
Write a parameter resolver that is capable of injecting parameters that are passed to `@ParametrizedTest` methods.

We are not alone with this problem. The issue is tracked in https://github.com/junit-team/junit5/issues/878 and there are partial solutions in referenced issues, e.g. https://github.com/junit-team/junit5/issues/3157

It looks as there is some chance that this will come up in one of the next JU5 releases, but for now it looks as if we have to work around.

This might be placed in a new module below `test-utils`.

**Alternatives**
Place in module `operaton-commons-testing`. But this has already a JU4 dependency and mixing might not be the best idea.

ArneDeutsch:
I've implemented an approach that maintains most of the test and only requires to change imports and annotations. Migration with that is quite easy because constructor, static parameter method and test code can stay unchanged.

kthoms:
This is a very pragmatic approach that helps us with the desired migration. Thanks @ArneDeutsch for the contribution.

PR #508, opened 2025-02-12
feat(test): New ParameterizedTestExtension

ArneDeutsch:
Add an extension for JUnit 5 to help migration of JUnit 4 Tests like ContainerAuthenticationFilterTest that are annotated with @RunWith(Parameterized.class). The extension implements the same mechanism for JUnit 5 such that the required code changes are minimal.

For the review please take into account the license header. I started a discussion at: 
https://forum.operaton.org/t/license-header-for-new-files/195

relates to #216, #27

### DIFF
--- a/test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
+++ b/test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright and/or licensed under one or more contributor license agreements.
+ * See the NOTICE file distributed with this work for additional information regarding
+ * copyright ownership. This file is licensed to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


kthoms:
This is a really nice addition. Actually when writing the issue I was more thinking of a different contract, namely that the arguments that are passed to parametrized test methods are being able to be passed to setup methods:

```
@BeforeEach
void setUp (String myArg) {}

@ParametrizedTest
void myTestMethod (String myArg) {}
```

However, your solution is also quite nice and would make the migration of existing tests much easier. So let's go for your way now.

There is a major problem, though. You placed the classes in the `junit5-extensions` module. This is a natural choice, but there was a reason why the issue was mentioning "a new module":
The `junit5-extensions` module depends on `operaton-engine` and that module contains the most tests that would leverage this extension. This would introduce a circular dependency.

`test-utils` seems legit as a container for such a extension.

Maybe we have to think about renaming `junit5-extension` module and rename it to `engine-junit5-extension`? Or create a new module `junit5-support`?
Or we move the existing engine-dependent extensions to the engine module itself into a separate source folder that could be used by the engine module and used by others, too? I.e. packaged separately with a different classifier?

I need to think about this. It should be possible to use these classes also from the engine module's tests.

kthoms:
I had an offline discussion with @ArneDeutsch about the topic into which module this extension would fit. At the end of the discussion we agreed that it is not part of this PR. For now, `junit5-extensions` is the right choice. It might be that other content has to be moved. This is now tracked by #512 .

PR #509, opened 2025-02-12
fix(webapps/assembly): Fix selectPDStatistics Query

hauptmedia:
Context: The query selectPDStatistics calculates the total processInstances wrong.
Used-by: ProcessDefinitionRestService#getStatisticsCount

Why: The group by incidentType part of the query should not be used for counting the total process instances.

Tests: The commit adds a test which expects the correct statistics returned when process instances related to incidents with different types are used.

Co-authored-by: Petros Savvidis <petros.savvidis@camunda.com>
Related to https://jira.camunda.com/browse/SUPPORT-24279

Backported commit 586037c29a from the camunda-bpm-platform repository.
Original author: Miklas Boskamp <20189772+mboskamp@users.noreply.github.com>

### DIFF
--- a/webapps/assembly/src/main/resources/org/operaton/bpm/cockpit/plugin/base/queries/processDefinition.xml
+++ b/webapps/assembly/src/main/resources/org/operaton/bpm/cockpit/plugin/base/queries/processDefinition.xml
@@ -97,7 +97,7 @@
 
               <!-- incidents -->
               left outer join
-                  <!-- Sum all incidents grouped by process definition and incident type-->
+                  <!-- Sum all incidents grouped by process definition -->
                   (
                     select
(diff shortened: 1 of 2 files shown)


webapps/assembly/src/main/resources/org/operaton/bpm/cockpit/plugin/base/queries/processDefinition.xml
```diff
@@ -101,7 +101,6 @@
                   (
```
kthoms, line 101: change comment, remove "and incident type"

kthoms (changes requested):

kthoms (approved):

Issue #512, opened 2025-02-13
Rework JUnit4/JUnit5 Testing Module Setup


ArneDeutsch:
Migrating JUnit 4 Tests to JUnit 5 Tests, especially in the engine project, require changes in our project layout. Karsten and I discussed this a bit and I would like to get additional input on how to solve this. Here are the problems:

- The "operaton-.engine" project contains JUnit 4 Tests that use JUnit 4 Extensions what extend JUnit 4 with functionality e.g. to start and stop the engine. This is a circular dependency, hence the JUnit 4 Extensions can not easily be extracted into another project.
- Users of Operaton might depend on these JUnit 4 extensions, hence they can not be removed
- We want to migrate tests in "operaton-.engine" to JUnit 5. This requires replacement extensions for the existing ones, targeting JUnit 5.
- Now we have extensions for JUnit 4 and JUnit 5 inside the "operator-engine" project, causing unneeded dependencies and confusion.

Maybe we can tackle this on maven level, introducing "junit4" and "junit5" classifier for the "operaton-.engine" project, placing the extensions in different source folders.

Apart from this we should also decide where to place JUnit4/JUnit5 extensions that are NOT dependent on the engine. But are needed e.g. to ease migration.

Please provide input on how we should tackle this. Better solutions then the one outlined above?

kthoms:
related to #27

ArneDeutsch:
Done. Ticket can be closed.

PR #525, opened 2025-02-21
chore(build): Refactor test project setup for JUnit 5 migration.

ArneDeutsch:
- Introduce junit4 and junit5 maven classifier for operaton-engine such that dependend projects can use junit4 as well as junit5 for tests.
- Engine dependent Rules/Extensions are now in junit4/jun it5 classified jar.
- Reverse dependency between operaton-engine and operaton-bpm-junit5. The later now contains engine independent junit5 extensions and can be used by other projects without depending on engine.
- Adapt all modules to use new maven modules/classifiers.

related to #512


engine-plugins/connect-plugin/pom.xml
```diff
@@ -33,10 +33,16 @@
     <dependency>
       <groupId>org.operaton.bpm</groupId>
       <artifactId>operaton-bpm-junit5</artifactId>
-      <version>${project.version}</version>
       <scope>test</scope>
     </dependency>
```
kthoms, line 37: These dependencies can be removed now

ArneDeutsch: Removed dependencies.

ArneDeutsch: Added junit-vontage to ensure Junit 4 Tests are also executed on operaton-engine.

test-utils/junit5-extension/pom.xml
```diff
@@ -38,17 +38,22 @@
...
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.operaton.bpm</groupId>
-      <artifactId>operaton-engine</artifactId>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-params</artifactId>
       <scope>compile</scope>
     </dependency>
+    <dependency>
```
kthoms, line 53: Why is this added?

kthoms: Move of DMN extensions and tests will be done via separate issue.

kthoms:
Thank you @ArneDeutsch ! This is a great change!

Issue #511, opened 2025-02-12
Resolve Sonar findings


kthoms:
**Is your feature request related to a problem? Please describe.**
Operaton aims to provide a high quality. We use Sonarqube to perform static code analysis on the code base. The findings Sonar is showing need to be resolved. Not all findings need to be resolved by code changes. 

We have to be careful about potential API changes, which have to be avoided, because we must not break users of C7 CE. For example, Sonar might report to reduce the visibility of fields, or perform renamings to comply to conventions. Within test code this is feasible since test code is not considered API. But in main code we have to be careful about this.

Furthermore, refactorings that change the code implementation should be thoroughly tested. The test coverage is already ok, but when the changed code is not covered new tests might be needed.

**Describe the solution you'd like**
Visit the [open findings on Sonarqube](https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton).

What you want to resolve does not matter, there are plenty open findings. Be it per class, package, or finding type. For the latter, select the "Rule" drop down and browse for rules.

![Image](https://github.com/user-attachments/assets/0243ef1c-f0aa-45e9-b1ae-247239b4a46b)

Best is to group changes that solve the same finding rule, or per class/package per commit. Make PRs of moderate size, or larger when they contain trivial repeating changes.

Overall, help us to reduce the open issues on Sonar.

antoinecampbell:
@kthoms I think it would be worthwhile to disable some of these issues as an example this [issue](https://sonarcloud.io/project/issues?rules=java%3AS1192&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton&open=AZLaQwRl8ZwKcH34-YpK).
It's marked as critical, but realistically this would not be fixed.

antoinecampbell:
Here's another issue that should be ignored: https://sonarcloud.io/project/issues?fileUuids=AZLaQmot8ZwKcH34-S1S&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton&open=AZLaQoGD8ZwKcH34-VOg. To address this issue it would break the convention that is established for these custom defined expression functions.

kthoms:
@biswa1shaw Anything you could work on? There are hundreds of issues left, just pick some.

kthoms:
Removing stale label, this is an ongoing maintenance effort.

LimpBizgit:
[...]

PR #526, opened 2025-02-21
chore: add assertions to tests

WoSchmo:
related to #511

### DIFF
--- a/engine/src/test/java/org/operaton/bpm/engine/test/bpmn/property/PropertyTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/test/bpmn/property/PropertyTest.java
@@ -32,8 +32,8 @@ public class PropertyTest extends PluggableProcessEngineTest {
 
     // DO NOT DELETE: WILL BE REACTIVATED SOON!
     // http://jira.codehaus.org/browse/ACT-88
-    
-    
+
+
 //    // Start the process -> waits in usertask
 //    Map<String, Object> vars = new HashMap<String, Object>();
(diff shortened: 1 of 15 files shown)


engine/src/test/java/org/operaton/bpm/engine/test/bpmn/property/PropertyTest.java
```diff
@@ -32,40 +32,8 @@ public void testUserTaskSrcProperty() {
 
     // DO NOT DELETE: WILL BE REACTIVATED SOON!
     // http://jira.codehaus.org/browse/ACT-88
-
```
kthoms, line 35: Please do not throw this away.
Instead you might want to activate the code with separate PR. Would be also fine to mark them ignored when they fail. But deleting this feels wrong here.

kthoms (changes requested):
Just small revert, rest looks fine!

engine/src/test/java/org/operaton/bpm/engine/test/bpmn/property/PropertyTest.java
WoSchmo: Oops, that was not intended. Thank you for pointing me out! I tried according to your suggestion, but there are references to variables, which are not declared, so let's leave it this way.

kthoms: Yes, I did not mean to do this with this PR. This could be done later.

PR #534, opened 2025-02-23
Move internal-dependencies (fixes #533)

### LINKED ISSUE #533: Move internal-dependencies to /bom
The internal-dependencies contain also a BOM like the other BOMs in /bom. Therefore this BOM should be moved to /bom.

Current structure
```
/bom
  /operaton-bom
      pom.xml
  /operaton-only-bom
      pom.xml
...
/internal-dependencies
  pom.xml
```

New structure
```
/bom
  /operaton-bom
      pom.xml
  /operaton-only-bom
      pom.xml
  /internal-dependencies
    pom.xml
...
```

Note that `internal-dependencies` is used quite often. Usually the restructuring should not matter as the pom.xml won't be referenced by paths, but by artifact coordinates that do not change.

FlamieCyrex:
- moved internal-dependencies BOM to .bom\
- removed duplicate dependency declarations in dependencyManagement
- ensured correct resolution of version.operaton.dbscheme from parent BOM
- fixed duplicate maven-dependency-plugin in integration-test-operaton-run profile"

### DIFF
--- a/webapps/assembly/pom.xml
+++ b/webapps/assembly/pom.xml
@@ -63,7 +63,7 @@
       <groupId>org.operaton.bpm</groupId>
       <artifactId>operaton-core-internal-dependencies</artifactId>
       <version>${project.version}</version>
-      <scope>import</scope>
+      <scope>compile</scope>
       <type>pom</type>
     </dependency>
 
--- a/webapps/assembly/pom.xml
+++ b/webapps/assembly/pom.xml
@@ -59,15 +59,6 @@
       </exclusions>
     </dependency>
 
-    <dependency>
-      <groupId>org.operaton.bpm</groupId>
-      <artifactId>operaton-core-internal-dependencies</artifactId>
-      <version>${project.version}</version>
-      <scope>compile</scope>
-      <type>pom</type>
-    </dependency>
-
-
     <!-- rest api -->
     <dependency>
       <groupId>org.operaton.bpm</groupId>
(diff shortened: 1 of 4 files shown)


webapps/assembly/pom.xml
```diff
@@ -63,7 +63,7 @@
       <groupId>org.operaton.bpm</groupId>
       <artifactId>operaton-core-internal-dependencies</artifactId>
       <version>${project.version}</version>
-      <scope>import</scope>
+      <scope>compile</scope>
```
kthoms, line 66: I'm wondering why this changed. I'll check if this dependency is necessary at all. Usually, internal-dependencies should be part of `dependencyManagement` with `import` scope. Looks fishy here. I'll check,

kthoms:
Thank you @FlamieCyrex  for your PR. I will have a look at it.

A formal question because this is your first contribution (and hopefully not the last): You agree that you are contributing this and all future contributions under Apache 2.0 license?

webapps/assembly/pom.xml
FlamieCyrex: @kthoms, i` ve changed it because it`s was`nt included in depencyManagement, so when i was checking it i thought that this is independent dependency.
I think, it`s my trouble.
So it`s needed to relocate in dependencyManagement?

FlamieCyrex:
@kthoms Yes, I agree to contribute under the Apache 2.0 license

webapps/assembly/pom.xml
kthoms: I meant the dependency itself looked bogus. Having an import scope BOM dependency in `<dependencies>` does not make sense. It has no effect.
Also, using `compile` scope for a BOM pom.xml also does not make sense, the internal-dependencies BOM only provides managed dependency versions, nothing to compile against.

I have removed this dependency and rebased your branch against a now hopefully fixed main branch.

If the build succeeds I will merge your change.

PR #536, opened 2025-02-24
chore: Fix two sonar findings with only one finding in code base.

ArneDeutsch:
- OperatonExtensionsTest: "Annotate this method with JUnit5 '@org.junit.jupiter.api.BeforeEach' or rename it to avoid confusion."
- DefaultAuthorizationProvider: "Refactor this method to not always return the same value."

relates to #511


ArneDeutsch:
I think the build failure is not related to the change. Can we re-run the build?

kthoms:
Yes, there was a build issue. I have rebased your branch.

kthoms (approved):

Issue #537, opened 2025-02-24
Resolve Sonar Reliability findings


kthoms:
**Is your feature request related to a problem? Please describe.**
Currently Operaton reaches "D" level in the category "Reliability".

![Image](https://github.com/user-attachments/assets/6dd63377-a425-4c3d-856c-4dc0485eb781)

**Describe the solution you'd like**

Go through the [findings list on SonarCloud](https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton).

Resolve the issues and raise PRs. If the issue can/should not be solved by code changes, please describe why these issues should be marked "accepted".

PR #544, opened 2025-02-27
chore: fix wrong bitnumber range check (#537)

WoSchmo:
@kthoms: Please resolve the following sonar findings manually, as they are false negatives: https://sonarcloud.io/project/issues?impactSeverities=HIGH&impactSoftwareQualities=RELIABILITY&rules=javabugs%3AS6466&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/util/BitMaskUtil.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/util/BitMaskUtil.java
@@ -98,8 +98,8 @@ public class BitMaskUtil {
     return MASKS[bitNumber - 1];
   }
 
-  private static void ensureBitRange(final int bitNumber) {
-    if(bitNumber <= 0 && bitNumber > 8) {
+  static void ensureBitRange(final int bitNumber) {
+    if(bitNumber <= 0 || bitNumber > 8) {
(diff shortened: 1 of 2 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/util/BitMaskUtil.java
```diff
@@ -99,7 +99,7 @@ public static int getMaskForBit(int bitNumber) {
   }
 
   private static void ensureBitRange(final int bitNumber) {
```
kthoms, line 101: Please do me a favor and add a unit test for BitMaskUtil. The fix is valid, but I think the functionality of this util class is critical enough to require a test.

It would be enough for this change to just add positive & negative test for method `ensureBitRange`. It is currently private. Either test it via one of the public methods. Or just remove "private" to make it package protected. It is totally valid to make private methods package-protected for the sake of covering the method by a unit test in the same package.

I'll check the mentioned Sonar issues now. Thanks for the link.

kthoms (changes requested):

kthoms (approved):
Thanksfor providing the test

PR #556, opened 2025-02-28
chore: Rewriting thread behavior to best practices.

ArneDeutsch:
We call 'await' in a loop until we run out of time or receive the signal. We exit immediately when receiving the signal and calculate if there is remaining time otherwise.

The 'await' can terminate either by a call from 'signal' or by timing out. Both can happen and are probably ok, but we should at least log the information that a timeout occurred while thread is still running.

If the thread is interrupted we catch the exception and signal the thread to exit gracefully.

relates to: #537


kthoms:
For this I would really like to have a unit test. I know that this is challenging.

kthoms (changes requested):
Test required

ArneDeutsch:
Added tests. Written against old implementation, same result with new implementation.

kthoms (approved):
Thanks for adding the test

PR #558, opened 2025-02-28
refactor: Replace stream.collect(Collectors.toList()) by stream.toList()

kthoms:
Resolves findings: https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&tags=java14%2Cjava15%2Cjava16%2Cjava17&id=operaton_operaton&open=AZLdYgMyewFQZUw2mxrH

 related to #85


engine/src/test/java/org/operaton/bpm/engine/test/api/mgmt/ManagementServiceAsyncOperationsTest.java
```diff
@@ -576,7 +576,7 @@ public void shouldSetInvocationsPerBatchTypeForJobsByProcessInstanceIds() {
   }
 
   protected List<String> getAllJobIds() {
-    return getAllJobs().stream().map(Job::getId).collect(Collectors.toList());
+    return getAllJobs().stream().map(Job::getId).toList();
```
javahippie, line 579: Stream#toList produces an unmodifiable list, making a test setup which adds IDs later complicated (or in this case, breaks), see line 189

Issue #566, opened 2025-03-04
Enable partial JUnit 5 migration for engine module


kthoms:
**Is your feature request related to a problem? Please describe.**
The engine module contains the largest amount of tests. Migrating them all at once will be rather challenging. Better is to migrate chunks of tests. This will ease the migration and review process.

**Describe the solution you'd like**
Create a separate test source folder for tests that are migrated to JUnit 5 in the engine module. 
Move some test classes and migrate them to JUnit 5.
Assure that the build executes both, JUnit 4 & JUnit 5 based tests.

**Acceptance Criteria**
Some engine tests have been moved to a separate test source folder.
These test classes have no dependency on JUnit 4.
The build executes still the same number of tests.

ArneDeutsch:
I experimented with this. It is easily possible to create a seperate source folder and have JUnit4 and JUnit5 tests seperated by source folder. This has it's strength but there is also a drawback.

Pro: We can see quite well what is migrated and what is left.
Contra: The review process is not so nice. There is no nice diff before/after, because the files have been moved.

It is also possible to have JUnit4 and JUnit5 Tests mixed in one source folder. This also works without flaws. The Pro/Contra is then reversed. It is harder to see the progress of the migration (fulltext search for "import org.junit" or similar would be needed) but the review of migrated tests would be a lot more easy (because of nice diff).

What do you think? What is better from your (everyone is asked) perspective?

kthoms:
Thinking about it, I would actually prefer to have the tests in one source folder. Review is important here. Searching for imports is fine for the gradual migration.

kthoms:
This ticket is closed, because partial migration in the same source folder is preferred.

PR #567, opened 2025-03-04
chore(test): Migrate first tests in engine from JUnit4 to JUnit5.

ArneDeutsch:
- Extended ParameteriedTestExtension to support `@Parameter(X)` on fields.
- Added unit test to cover new functionality.
- Migrated first four tests in engine project to JUnit5.

reöates to #566

### DIFF
--- a/test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
+++ b/test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
@@ -22,6 +22,7 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 import java.lang.reflect.Constructor;
+import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
@@ -38,6 +39,7 @@ import org.junit.jupiter.api.extension.ParameterResolutionException;
 import org.junit.jupiter.api.extension.ParameterResolver;
 import org.junit.jupiter.api.extension.TestInstanceFactory;
 import org.junit.jupiter.api.extension.TestInstanceFactoryContext;
+import org.junit.jupiter.api.extension.TestInstancePostProcessor;
 import org.junit.jupiter.api.extension.TestInstantiationException;
 import org.junit.jupiter.api.extension.TestTemplateInvocationContext;
 import org.junit.jupiter.api.extension.TestTemplateInvocationContextProvider;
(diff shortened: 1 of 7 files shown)


test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
```diff
@@ -158,22 +166,42 @@ public Object resolveParameter(ParameterContext parameterContext,
...
+							for (Field field : testInstance.getClass().getDeclaredFields()) {
+								Parameter parameterAnnotation = field.getAnnotation(Parameter.class);
+								if (parameterAnnotation != null) {
+									int index = parameterAnnotation.value();
+									if (index < 0 || index >= parameters.length) {
+										throw new ExtensionConfigurationException("Index " + index
+												+ " is out of bounds for the provided parameters array.");
+									}
+									field.setAccessible(true);
+									field.set(testInstance, parameters[index]);
```
kthoms, line 201: do you think the field accessibility should be reset after injecting the field? Likely does not make any difference, though, just feels more correct.

test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
```diff
@@ -77,6 +80,12 @@
  */
 public class ParameterizedTestExtension implements TestTemplateInvocationContextProvider {
 
+	@Retention(RetentionPolicy.RUNTIME)
```
kthoms, line 83: Nice extension of `ParameterizedTestExtension`

When you are on it: Please make the javadocs proper HTML with `<ol><li>...` for lists , `<pre>` for code blocks.
And add something about the parameters.

kthoms (changes requested):

test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
ArneDeutsch: Hmmm. Don't feel the same way. It would be in the same way more correct to add a AfterEach method to your test classes that nulls the fields all the time when you have an BeforeEach that fills the fields. I would not do add this code as it not really adds value.

test-utils/junit5-extension/src/main/java/org/operaton/bpm/engine/test/junit5/ParameterizedTestExtension.java
ArneDeutsch: Will do.

kthoms (approved):

PR #571, opened 2025-03-05
chore(test): Migrate some engine JUnit4 tests to JUnit5.

ArneDeutsch:
relates to #27


engine/src/test/java/org/operaton/bpm/container/impl/jmx/deployment/MockInitialContextRule.java
kthoms, line 1: Move to junit4 instead of deleting?

engine/src/test/java/org/operaton/bpm/engine/impl/batch/BatchInvocationsPerJobByBatchTypeConfigTest.java
```diff
@@ -39,32 +39,23 @@ public class BatchInvocationsPerJobByBatchTypeConfigTest {
...
-  @After
+  @AfterEach
   public void teardown() {
     processEngine.close();
   }
 
   @Test
   public void shouldSetInvocationsPerJobByBatchType() {
     // when
+    initEngine();
```
kthoms, line 58: Could initialization be done by ProcessEngineExtension?

kthoms (changes requested):

engine/src/test/java/org/operaton/bpm/container/impl/jmx/deployment/MockInitialContextRule.java
ArneDeutsch: Was only used in this one specific test case and dosen't seem to valuable ... but maybe someone has picked it up and used it for some other test case. Ok. will restore it, just in case.

kthoms (approved):

PR #610, opened 2025-03-15
chore: Resolve Sonar findings

antoinecampbell:
- Rename a field that had the same name as an inherited method
- Cleanup warning

related to #511


kthoms (approved):

antoinecampbell:
@kthoms I think we need to re-run the build. The test in question is not failing locally `SequentialJobAcquisitionTest`, based on the waits within the test this may be a flaky test.

PR #612, opened 2025-03-16
chore: Resolve Sonar findings

antoinecampbell:
- Rename a static field that had the same name as class member
- Mark unserializable field as transient

related to #511

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/persistence/entity/TaskEntity.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/persistence/entity/TaskEntity.java
@@ -181,30 +181,30 @@ public class TaskEntity extends AbstractVariableScope implements Task, DelegateT
   protected transient List<VariableInstanceLifecycleListener<VariableInstanceEntity>> customLifecycleListeners;
 
   // name references of tracked properties
-  public static final String PROPERTY_ASSIGNEE = "assignee";
-  public static final String PROPERTY_DELEGATION = "delegation";
-  public static final String PROPERTY_DELETE = "delete";
-  public static final String PROPERTY_DESCRIPTION = "description";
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/persistence/entity/TaskEntity.java
```diff
@@ -178,33 +178,33 @@ public class TaskEntity extends AbstractVariableScope implements Task, DelegateT
...
-  public static final String DELETE = "delete";
-  public static final String DESCRIPTION = "description";
-  public static final String DUE_DATE = "dueDate";
-  public static final String FOLLOW_UP_DATE = "followUpDate";
-  public static final String NAME = "name";
-  public static final String OWNER = "owner";
-  public static final String PARENT_TASK = "parentTask";
-  public static final String PRIORITY = "priority";
-  public static final String CASE_INSTANCE_ID = "caseInstanceId";
+  public static final String PROPERTY_ASSIGNEE = "assignee";
```
kthoms, line 184: Unfortunately these fields are public. Even when the likelyness of any client using them is low, I would consider them as API for now.
It will be possible to rename them if we would have an OpenRewrite recipe to migrate usages. Please raise an issue to https://github.com/operaton/migrate-from-camunda-recipe describing the renaming, or revert.

kthoms (changes requested):

engine/src/main/java/org/operaton/bpm/engine/impl/persistence/entity/TaskEntity.java
antoinecampbell: I've reverted the renaming of these fields

kthoms (approved):

Issue #565, opened 2025-03-03
Resolve dependency on Hamcrest


kthoms:
**Is your feature request related to a problem? Please describe.**
Some test code still relies on the Hamcrest library. 

**Describe the solution you'd like**
For a cleaner, fluent testing Operaton is using AssertJ. Refactor the code so that Hamcrest is no longer used by test code.

To identify related code
- search for "`import org.hamcrest`" in Java files
- search for "`<groupId>org.hamcrest</groupId>`" in pom.xml files

Resolve usages and clean up dependencies.

WoSchmo:
There is already an import to 
<dependency>
      <groupId>com.konghq</groupId>
      <artifactId>unirest-java</artifactId>
      <version>2.3.10</version>
      <classifier>standalone</classifier>
      <scope>provided</scope>
    </dependency>
which provides JSONObject & Co.

Alternatively there is
<dependency>
    <groupId>org.glassfish.jersey.media</groupId>
    <artifactId>jersey-media-json-jackson</artifactId>
    <version>2.35</version>
</dependency>

cachescrubber:
I started looking into this.

In **engine-rest-core** [REST Assured](https://github.com/rest-assured) is used, which depends on hamcrest. AFAIK  rest-assured has no asserts support. Migrating the test base to another framework would be quite an effort.

kthoms:
Let's leave the state as is for now.

Sine **engine-rest-core** is using REST Assured, that's ok to have that indirect dependency. Same is for JUnit 4 anyway.

The modules
* qa/test-db-instance-migration
* qa/test-db-rolling-update
can be handled once we activate those tests for Operaton. It is not clear yet when and if we do this.

Thanks @cachescrubber for this valuable change.

PR #619, opened 2025-03-20
Resolve dependency on Hamcrest

cachescrubber:
Work on #565 (Resolve dependency on Hamcrest)


engine/src/main/junit-shared/org/operaton/bpm/engine/test/api/authorization/util/MissingAuthorizationMatcher.java
kthoms, line 1: why is this removed?

kthoms (commented):
@cachescrubber Looks good, just one question

engine/src/main/junit-shared/org/operaton/bpm/engine/test/api/authorization/util/MissingAuthorizationMatcher.java
cachescrubber: Te class directly implements a hamcrest Type. The only usage was from `org.operaton.bpm.engine.test.api.authorization.util.AuthorizationScenarioInstance#assertAuthorizationException`. The code needed had no hamcrest dependencies and has been inlined to `AuthorizationScenarioInstance`.

kthoms: Thanks for the explanation. Then I do not have further questions.

kthoms (approved):
Thanks a lot for this contribution

Issue #604, opened 2025-03-15
Merge engine-rest jakarta modules


kthoms:
Merge modules

- `engine-rest/assembly-jakarta` with `engine-rest/assembly`
- `engine-res/engine-rest-jakarta` with `engine-rest/engine-rest? 

See #600 for a detaiiled description of the task.

cachescrubber:
Sorry for the confusion, I missed #637 is already in progress.

kthoms:
Also sorry for the confusion. Great that you have been working on it, too. Now it is done.

PR #640, opened 2025-03-24
Merge engine rest jakarta (follow-up)

cachescrubber:
Follow-Up on #637. Related to #604 

Yesterday I created a merge request to migrate the engine-rest-jakarta module. I did oversee that it was a WIP by @kthoms already. This PR is what is left over or what I did different. Basically the **resteasy** build profile and associated code.

Best,
Lars


kthoms (approved):
Thanks a lot for this additional changes. Could easily happen ATM that something is overseen. So this change is warmly welcome.

Side note: I try to concentrate any versioning of libraries into internal-dependencies BOM. It should not be necessary to specify version attributes for dependencies, even not with version properties. I will further clean this up later.

distro/webjar/pom.xml
```diff
@@ -32,17 +32,6 @@
     </dependencies>
   </dependencyManagement>
 
-  <!-- add dependency so this is run after the war was build -->
```
kthoms, line 35: I think this has to stay. It was a self-reference, but actually I changed it to refer to the webapp (not webjar!).
I was struggling with the build when I also tried to remove this dependency. I fear it has to stay.
You may not have recognized this since you might have already installed libraries in ~/.m2/repository/org/operaton

kthoms (changes requested):

distro/webjar/pom.xml
kthoms: I tried again with a clean build and it succeeded. Strange. Exactly this dependency caused yesterday troubles. Now it worked with a clean build.
Let's try for now. When we see that it gets unstable then we may need to add the dependency again.

cachescrubber:
I just reverted the webinar pom comment - probably a second to late :-)

PR #653, opened 2025-03-26
chore(test): Convert all tests in filter and form packages.

ArneDeutsch:
relates to #27

### DIFF
--- a/engine/src/test/java/org/operaton/bpm/engine/test/api/filter/FilterTaskQueryTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/test/api/filter/FilterTaskQueryTest.java
@@ -16,7 +16,9 @@
  */
 package org.operaton.bpm.engine.test.api.filter;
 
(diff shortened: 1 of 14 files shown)


engine/src/test/java/org/operaton/bpm/engine/test/api/filter/FilterTaskQueryTest.java
```diff
@@ -63,7 +74,20 @@
 /**
  * @author Sebastian Menski
  */
-public class FilterTaskQueryTest extends PluggableProcessEngineTest {
+public class FilterTaskQueryTest {
+  
+  @RegisterExtension
+  protected static ProcessEngineExtension engine = new ProcessEngineExtension();
```
kthoms, line 80: The ProcessEngineExtension should always be created via the builder when instantiating for ' @RegisterExtension`. Please verify other usages of the default constructor.

kthoms (changes requested):

kthoms (approved):

PR #664, opened 2025-03-27
dependencies: Remove Kong Unirest (and SetupTests main Class)

### LINKED ISSUE #659: Resolve or update com.konghq:unirest-java dependency
The POM `qa/integration-tests-webapps/integration-tests/pom.xml` declares a dependency on `m.konghq:unirest-java:jar:standalone:2.3.10`.

Either upgrade the library to the latest stable version, or refactor the code to use a different REST client. The normal library used in the code base is Resteasy, this would be the first choice. Plain Java HTTP is also ok.

**Additional context**
This version is from July 2019 and has a transitive dependency on org.json:json in a version that Dependabot raises a security issue for. Since this is only in test code, the security issue is not relevant.

javahippie:
Fixes #659


kthoms (approved):
LGTM

PR #698, opened 2025-04-07
fix(spin): make jsonPath use Double nodes as default number type

hauptmedia:
Related to https://github.com/camunda/camunda-bpm-platform/issues/2575


javahippie (approved):
lgtm, thanks!

PR #701, opened 2025-04-07
chore: bump Tomcat version to 10.1.39

hauptmedia:
Bumping Tomcat version to 10.1.39. We have currently two places where we manage the tomcat version, one is in the engine-rest pom with a comment "pin tomcat version for engine-rest". We need to investigate if this must stay pinned to 10.1.35 (and for what reason and possibly open an issue for it).

I opened the PR to see if the whole test pipelines runs through when we increase the at both places to 10.1.39

### DIFF
--- a/engine-rest/engine-rest/pom.xml
+++ b/engine-rest/engine-rest/pom.xml
@@ -12,7 +12,7 @@
 
   <properties>
     <!-- pin tomcat version for engine-rest -->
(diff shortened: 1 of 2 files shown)


kthoms:
Triggered an [integration build](https://github.com/operaton/operaton/actions/runs/14318779003)

kthoms (approved):
Integration tests passed for
- tomcat,h2
- tomcat,postgresql

PR #710, opened 2025-04-08
feat(core): expose process def key in process instance API (#4847)

hauptmedia:
related to camunda/camunda-bpm-platform#4049

Signed-off-by: Shenoy, Nandan <Nandan.Shenoy@fmr.com>
Co-authored-by: <yanavasileva@users.noreply.github.com>

Backported commit efdc8c800d from the camunda-bpm-platform repository.
Original author: yanavasileva <yanavasileva@users.noreply.github.com>


kthoms:
I plan to enable the rolling DB update tests for Beta-5 and expect that the required field is added with the DB upgrade scripts. Once the tests are running we can proceed here.

hauptmedia:
we have an incomplete database schema for C7 7.23, this field is missing to be C7 7.23 compatible

PR #717, opened 2025-04-10
feat(rest): make LockAndFetchHandler blocking queue's capacity configurable

hauptmedia:
Related to https://github.com/camunda/camunda-bpm-platform/issues/4885

Backported commit 6154d43341 from the camunda-bpm-platform repository.

### DIFF
--- a/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/impl/FetchAndLockHandlerImpl.java
+++ b/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/impl/FetchAndLockHandlerImpl.java
@@ -51,14 +51,16 @@ public class FetchAndLockHandlerImpl implements Runnable, FetchAndLockHandler {
   private static final Logger LOG = Logger.getLogger(FetchAndLockHandlerImpl.class.getName());
 
   protected static final String UNIQUE_WORKER_REQUEST_PARAM_NAME = "fetch-and-lock-unique-worker-request";
+  protected static final String BLOCKING_QUEUE_CAPACITY_PARAM_NAME = "fetch-and-lock-queue-capacity";
 
   protected static final long PENDING_REQUEST_FETCH_INTERVAL = 30L * 1000;
   protected static final long MAX_BACK_OFF_TIME = Long.MAX_VALUE;
   protected static final long MAX_REQUEST_TIMEOUT = 1800000; // 30 minutes
+  protected static final int DEFAULT_BLOCKING_QUEUE_CAPACITY = 200;
 
   protected SingleConsumerCondition condition;
 
-  protected BlockingQueue<FetchAndLockRequest> queue = new ArrayBlockingQueue<>(200);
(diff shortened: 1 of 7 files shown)


kthoms (commented):
Fine for me.

@javahippie Ok to add this feature with 1.0.0-beta-4, or add in beta-5?

engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/impl/FetchAndLockHandlerImpl.java
```diff
@@ -357,6 +366,27 @@ protected void parseUniqueWorkerRequestParam(String uniqueWorkerRequestParam) {
...
+    queue = new ArrayBlockingQueue<>(capacity);
+  }
+
+  private static int parseBlockingQueueCapacityParam(String queueSizeRequestParam) {
+    int capacity = DEFAULT_BLOCKING_QUEUE_CAPACITY;
+    if (queueSizeRequestParam != null) {
+      try {
+        final int parsedCapacity = Integer.parseInt(queueSizeRequestParam);
+        if (parsedCapacity <= 0) {
+          throw new NumberFormatException("Parameter " + BLOCKING_QUEUE_CAPACITY_PARAM_NAME + " has to be greater than zero");
```
javahippie, line 380: I'm not the biggest fan of throwing something and catching it immediately in the same block. I'd argue this is not really a NumberFormatException either, it is used as a vehicle for program flow here

kthoms: Yes, saw that too. It is to use the same catch as when parseInt would fail. I was thinking more of an IllegalArgumentException here.
Maybe the flow could be slightly improved here.

engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/impl/FetchAndLockHandlerImpl.java
```diff
@@ -357,6 +366,27 @@ protected void parseUniqueWorkerRequestParam(String uniqueWorkerRequestParam) {
...
+  private static int parseBlockingQueueCapacityParam(String queueSizeRequestParam) {
+    int capacity = DEFAULT_BLOCKING_QUEUE_CAPACITY;
+    if (queueSizeRequestParam != null) {
+      try {
+        final int parsedCapacity = Integer.parseInt(queueSizeRequestParam);
+        if (parsedCapacity <= 0) {
+          throw new IllegalArgumentException("Parameter " + BLOCKING_QUEUE_CAPACITY_PARAM_NAME + " has to be greater than zero");
+        }
+        capacity = parsedCapacity;
+      } catch (IllegalArgumentException e) {
```
javahippie, line 383: After this change, a `NumberParseException` thrown by `Integer#parseInt` is no longer handled, right?

PR #735, opened 2025-04-19
chore: delete old db test fixtures

### LINKED ISSUE #503: Drop old test fixtures
The module `qa/test-db-instance-migration` contains test scenarios for upgrading the Operaton database version from one minor release to the next. It contains C7 versions from 7.2 to 7.23. These versions simply do not exist in Operaton.

Image

We should agree on which is the first version which we actually include (7.23?) and then drop the others.

kthoms:
delete fixtures for database versions lower than 7.23

closes #503


kthoms:
@javahippie Is it safe to delete these resources?

hauptmedia:
I think it's a general question how our migration guide will look like. Do we suggest to upgrade to C7 7.24 CE first before upgrading to Operaton (which I would prefer to minimize testing efforts) or do we support upgrading from any C7 version to Operaton?

javahippie:
I agree with @hauptmedia here, maybe we should open a Forum topic on the matter? 
For us, life would be easier if we only "allow" migration from Camunda 7.24CE to Operation 1.0.0. On the other hand we mirror all SQL files since 7.22, so a migration might be possible

hauptmedia:
Many users will have to make some minor changes when upgrading from 7.xx to 7.24CE already and i fear that they mix up the changes needed to upgrade to 7.24 with the changes needed to make to migrate to Operaton and have an overall frustrating experience.

For me it would be much clearer to say: check that your application works with 7.24 first before you migrate it to Operaton. And yes let's move further discussion to the forum.

PR #742, opened 2025-04-22
resolve potential ArrayIndexOutOfBoundsException

WoSchmo:
resolves: https://sonarcloud.io/project/issues?impactSeverities=HIGH&impactSoftwareQualities=RELIABILITY&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton&open=AZOSJL45y4PHewdj1OX3


kthoms:
Looks good. Could you add a unit test that provides different values (<0, 0, 1, 60)?

Matter of taste: I prefer early exits, i.e. check the invariant first and throw the exception at the beginning. This would avoid an indentation level for the normal case. Optional.

kthoms (approved):

PR #759, opened 2025-04-25
feat(queryAPI): add filter for instances with jobs retrying in java API

hauptmedia:
related to camunda/camunda-bpm-platform#4904

Backported commit 22c50da015 from the camunda-bpm-platform repository.
Original author: Helene <21290204+PHWaechtler@users.noreply.github.com>

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/history/HistoricProcessInstanceQuery.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/history/HistoricProcessInstanceQuery.java
@@ -39,12 +39,16 @@ public interface HistoricProcessInstanceQuery extends Query<HistoricProcessInsta
   /**
    * Only select historic process instances with the given process instance.
    * {@link ProcessInstance} ids and {@link HistoricProcessInstance} ids match.
+   *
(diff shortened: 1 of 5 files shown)


hauptmedia:
When this PR gets merged there will be also two consecutive PRs that adds this query option to the REST API and webapps

engine/src/main/java/org/operaton/bpm/engine/history/HistoricProcessInstanceQuery.java
```diff
@@ -162,6 +162,13 @@ public interface HistoricProcessInstanceQuery extends Query<HistoricProcessInsta
    */
   HistoricProcessInstanceQuery incidentMessageLike(String incidentMessageLike);
 
+  /**
+   * Only select historic process instances which are associated with jobs that have exceptions and retries left.
+   *
+   * @return HistoricProcessInstanceQuery
+   */
+  HistoricProcessInstanceQuery withJobsRetrying();
```
kthoms, line 170: for backward compatibility adding methods to a public API interface should be done with a _default method_.

kthoms:
Resolved my findings and rebased the code

PR #768, opened 2025-04-27
chore: bump GraalVM version to 21.3.14

hauptmedia:


WoSchmo:
Should be obsolete because of: https://github.com/operaton/operaton/pull/774

hauptmedia:
Hi @WoSchmo,

the 21.3.14 version is the most up to date version of the "GraalVM Enterprise Edition" that is being used. I also thought of updating it further. Thank you for doing that. As the beta-4 release is near and we already have a lot of changes there, my suggestion would be to bump to 21.3.14 in beta-4 and further update it in beta-5.

WDYT?

Cheers,

Julian

WoSchmo:
@kthoms asked me to take care of that ticket. As it is probably a relatively crucial part of operaton and we are "skipping" several major versions, it might be best, to wait a little.

Issue #500, opened 2025-02-07
IdGeneratorDataSourceTest is silently broken


ArneDeutsch:
**Describe the bug**
The test org.operaton.bpm.engine.test.api.cfg.IdGeneratorDataSourceTest is green but actually doesn't test anything. The thrown exceptions are suppressed and only printed to the console, but the test does not fail.

**To Reproduce**
1. Go to 'org.operaton.bpm.engine.test.api.cfg.IdGeneratorDataSourceTest'
2. Execute the unit test (e.g. from eclipse)
3. Observe the console and notice the printed exceptions

**Expected behavior**
The test should actually verify the multithreading behaviour when executing IdGenerator from multiple threads and fail if the behaviour is not acceptable. Exceptions might not be silently ignored.

**Additional context**
Observed exception:

```
Exception in thread "Thread-7" org.operaton.bpm.engine.ProcessEngineException: ENGINE-09008 Exception while instantiating class 'org.operaton.bpm.engine.test.db.IdGeneratorDataSourceDoNothing': ENGINE-09017 Cannot load class 'org.operaton.bpm.engine.test.db.IdGeneratorDataSourceDoNothing': org.operaton.bpm.engine.test.db.IdGeneratorDataSourceDoNothing
	at org.operaton.bpm.engine.impl.util.EngineUtilLogger.exceptionWhileInstantiatingClass(EngineUtilLogger.java:88)
	at org.operaton.bpm.engine.impl.util.ClassDelegateUtil.instantiateDelegate(ClassDelegateUtil.java:56)
	at org.operaton.bpm.engine.impl.bpmn.behavior.ClassDelegateActivityBehavior.getActivityBehaviorInstance(ClassDelegateActivityBehavior.java:102)
	at org.operaton.bpm.engine.impl.bpmn.behavior.ClassDelegateActivityBehavior.lambda$0(ClassDelegateActivityBehavior.java:65)
	at org.operaton.bpm.engine.impl.bpmn.behavior.AbstractBpmnActivityBehavior.executeWithErrorPropagation(AbstractBpmnActivityBehavior.java:90)
	at org.operaton.bpm.engine.impl.bpmn.behavior.ClassDelegateActivityBehavior.execute(ClassDelegateActivityBehavior.java:64)
	at org.operaton.bpm.engine.impl.pvm.runtime.operation.PvmAtomicOperationActivityExecute$2.callback(PvmAtomicOperationActivityExecute.java:63)
[...]

PR #778, opened 2025-05-01
fix: failing IdGeneratorDataSourceTest

shiftyseaotter:
fix(engine): failing IdGeneratorDataSourceTest

- fixes siliently failing IdGeneratorDataSourceTest
- adds assertion to prevent it from happing again

relates to #500


kthoms (approved):
Thank you @shiftyseaotter

PR #781, opened 2025-05-02
chore: improve deprecation warnings (#511)

Kungi:
Further work on #511


engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/BusinessProcess.java
```diff
@@ -481,9 +481,9 @@ public VariableMap getAndClearCachedVariableMap() {
    * Get the map of cached variables and clear the internal variable cache.
    *
    * @return the map of cached variables
-   * @deprecated use {@link #getAndClearCachedVariableMap()} instead
+   * @deprecated Use {@link #getAndClearCachedVariableMap()} instead.
    */
-  @Deprecated(forRemoval = true)
+  @Deprecated(forRemoval = true, since = "1.0.0-beta-1")
```
kthoms, line 486: General question: I was also thinking about adding "since". I would not use the "beta" suffix. The actual release since when it is deprecated is 1.0.0, and deprecations I would expect only on a minor version level. For me, "since 1.0" would be the right message. Maybe you raise that topic for discussion in the forum?

engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/impl/event/CdiExecutionListener.java
```diff
@@ -21,6 +21,6 @@
  *
  * @author Sebastian Menski
  */
-@Deprecated
+@Deprecated(forRemoval = true, since = "1.0.0-beta-1")
```
kthoms, line 24: General: For deprecated classes/methods that have no dedicated replacement I would not mark them for removal. Not everything that is deprecated has to be removed, although likely it will.

To mark new removals this would have to be decided on case-by-case level. Please remove "forRemoval" for such cases. In other words, only add "forRemoval" when the deprecation javadoc states what to use instead.

kthoms (changes requested):

kthoms:
I started a discussion in the forum: https://forum.operaton.org/t/deprecation-strategy/237

engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/BusinessProcess.java
kthoms: It was decided that it must be `since = "1.0"`, see forum thread

kthoms:
I have worked through your changes and mainly 
- set `since = "1.0"` for deprecated API
- remove `forRemoval = true` when a replacement is unclear

Kungi (commented):
I reviewed all changes on a file by file basis and prettied up some things.

Issue #821, opened 2025-05-22
Remove obsolete modifiers from interface members


kthoms:
**What needs to be done? Please describe.**

`public` `static` `final` are implicit for interface members. Search for interfaces in the code base which have members with such modifiers and remove the modifiers.

Example:
```java
public interface ProcessEngineService {
  ...
  public ProcessEngine getDefaultProcessEngine();
  ...
```

change to 
```java
public interface ProcessEngineService {
  ...
 ProcessEngine getDefaultProcessEngine();
  ...
```
-----------------------

```java
public interface HistoricActivityInstanceQueryProperty {
  ...
  public static final QueryProperty HISTORIC_ACTIVITY_INSTANCE_ID = new QueryPropertyImpl("ID_");
  ...
```

change to 
```java
public interface HistoricActivityInstanceQueryProperty {
  ...
  QueryProperty HISTORIC_ACTIVITY_INSTANCE_ID = new QueryPropertyImpl("ID_");
  ...
```

Consider to provide PRs in chunks.

**Additional context**

Search for an OpenRewrite recipe (did not find one on first sight). If there is none, consider to report this to OpenRewrite as a feature request.

zeguilherme99:
Hi! 👋 I'd love to contribute by working on this issue.  
Is it okay if I take it?

zeguilherme99:
@ungerts Thanks, for the tip.

ungerts:
You're right—the static identifier isn't defined on a member of the interface. That said, I’d still recommend removing it, as it doesn’t seem necessary.

kthoms:
> and remove redundant modifiers in inner interfaces?

A nested interface is implicitly static. So yes, removal is ok here.

kthoms:
@zeguilherme99 Thanks for taking this. I have assigned the task to you. Please just indicate once that you are fine to contribute this and following according to the Apache 2.0 license. We don't have a formal CLA yet.

kthoms:
Thank you @zeguilherme99 and a warm welcome as new contributor ❤️

kthoms:
22 Remaining
![Image](https://github.com/user-attachments/assets/e2329744-39d3-4c1e-89a4-3b1749ac85fb)

zeguilherme99:
Sorry, I'll resolve that.

kthoms:
@zeguilherme99 This did not mean that you have to do it. It was more looking what is left to close. But great if you could take a look +1

kthoms:
[...]

PR #839, opened 2025-05-26
chore(juel): Unnecessary Modifier

zeguilherme99:
- Remove obsolete modifiers from interface members

related to #821


juel/src/main/java/org/operaton/bpm/impl/juel/ExpressionNode.java
```diff
@@ -30,18 +30,18 @@ public interface ExpressionNode extends Node {
 	/**
 	 * @return <code>true</code> if this node represents literal text
 	 */
-	public boolean isLiteralText();
+    boolean isLiteralText();
```
kthoms, line 33: inconsistent formatting

kthoms:
I've fixed the findings

kthoms (approved):

PR #841, opened 2025-05-26
chore(engine): Unnecessary Modifier

zeguilherme99:
- Remove obsolete modifiers from interface members

related to #821


kthoms (changes requested):
IMO the visibility of enum constructors should not be weakened. Leave private instead of package protected.

kthoms:
I have fixed the findings

Issue #852, opened 2025-06-02
License URL Update


kthoms:
**What needs to be done? Please describe.**

Several files contain HTTP instead of HTTPS URLs.

Replace 
```
http://www.apache.org/licenses/LICENSE-2.0
http://www.apache.org/licenses/LICENSE-2.0.txt
```

by 
```
https://www.apache.org/licenses/LICENSE-2.0
```

Skip files that are not directly related to Operaton owned sources, e.g. `mvnw`. In doubt just change and wait for review feedback.

Waize:
Hi! 
I've used this PR to set up the project and replaced the links with fulltext replacement in IntelliJ. 

I've saw some whitespace changes in Java files, which looks fine for me. 
I appreciate your feedback. :)

kthoms:
Thank you for solving this issue.

PR #867, opened 2025-06-06
chore: replace http license-addresses with https

Waize:
related to #852


kthoms:
@Waize Thanks for providing the patch. I'll need a bit time to review it, since I'm on vacation right now.

A formal question: As first time contributor I need to ask if you are fine to contribute this and subsequent patches under Apache 2.0 license. This especially means that no code is copied from incompliant sources.

Just state that this is fine for you.

Waize:
Hi, 

This is absolutely fine. 

Have a nice vacation!

kthoms:
@Waize the script (?) that you used for refactoring touched more than expected. It also removes empty lines and trailing whitespace. This makes it hard to see if there are accidential changes.

I have performed this command:
```
git log -p -n 1 | grep -E "^[-+] " | grep -v "LICENSE-2.0" > change.txt
```

I would expect an empty result. First this prints out all changes, but then it considers only additions and removals (lines starting with -/+). From that I expect only change in licenses, and excluding them should be empty.

Would it be possible to adjust your script? And if it is a script print it here?

kthoms:
@Waize Hold on before taking any action. I stepped through some hundred file changes already and they look good so far. My fear was that due to the vast amount of changes I might oversee something and that it would be better to separate the whitespace changes.
I will step through the changes further. It is just a very huge changeset. So far it looks as if the change could be committed.

Waize:
Hi @kthoms , 

I can explain what happened: 
I replaced the URLs with IntelliJ (Ultimate ) "replace in files" . I double checked, that your code settings are correctly set. 
IntelliJ is doing some formatting actions if you touch a file (one could turn that behaviour off). I really like that feature because it ensure the same code style for every class. 
Sadly it leads on those large changes to whitespace changes on many files, depending on the creation date and code style adjustments over time.

In summary those whitespace changes will appear sooner or later, but I totally understand the fuzziness they create in this PR. 
Maybe we can see this as the baseline that every touched file is now formatted the way the Operaton codestyles describe them to be. :) 


Side note: after merging this PR, the other PR for removing Optimize might get merge conflicts. I will resolve them at notice.

kthoms:
Yes, I also think that those whitespace changes would have to be done anyway. I want to reach a clean and harmonized code style. This naturally has huge implications on changesets. I would like to have version 1.0 in a state that this happens less.

Understanding how such huge changes have been done is important, so that I know where to double-check maybe.

I think we are good to go now with this. Many thanks for this contribution!

kthoms (approved):

PR #889, opened 2025-06-14
chore(build): Re-order POMs with sortpom-maven-plugin

kthoms:
This change configures and uses the sortpom-maven-plugin to sort Maven pom.xml files.

Some dependencies and properties are configured to be ignored by the plugin. Added a profile `code-cleanup` that executes the plugin.


kthoms:
Superseded by #914

PR #890, opened 2025-06-15
Skip nightly builds with no recent changes

### LINKED ISSUE #596: Run scheduled Nightly Build only when main has changed
The Nightly Build workflow is scheduled for 3 AM and is executed then unconditionally.

The NIghtly Build does not need to be executed when nothing has changed during the day on the `main` branch.

Configure `.github/workflows/nighly-build.yml` for scheduled builds that changes from the last 24 hours are checked. When no commit on `main` has happened, the build can abort.

It should be visible easily (e.g. in the build log, but maybe already on the Actions page) that the Nightly Build was aborted because nothing has changed.

AlexanderSkrock:
Aims to fix #596 

I decided to introduce a separate job for change detection whose output can be used to skip other jobs. While we only evalute the result for scheduled jobs, it could be handy for initial testing that the checks are performed for non-scheduled jobs as well.

### DIFF
--- a/.github/workflows/integration-build.yml
+++ b/.github/workflows/integration-build.yml
@@ -91,6 +91,7 @@ jobs:
           else
             echo "🔎✔️ Commits within the last 24 hours detected."
             echo "commits=true" >> $GITHUB_OUTPUT
+          fi
 
   build:
     name: Build
--- a/.github/workflows/integration-build.yml
+++ b/.github/workflows/integration-build.yml
@@ -71,6 +71,27 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  detect-recent-changes:
+    name: Detect recent changes
+    runs-on: ubuntu-24.04
+    outputs:
+      recent_changes: ${{steps.check-commits.outputs.commits}}
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Check for recent commits
+        id: check-commits
+        run: |
+          COMMITS=$(git rev-list --since="24 hours ago" HEAD)
+          if [ -z "$COMMITS" ]; then
+            echo "No commits in the last 24 hours."
+            echo "commits=false" >> $GITHUB_OUTPUT
+          else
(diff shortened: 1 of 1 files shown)


kthoms:
Hey @AlexanderSkrock , thanks for contributing this. It looks that this here should do the job that is wanted.

As a first time contributor, could you please just state that you are fine to contribute this and (hopefully) following changes under Apache 2.0 license?

I think you can remove the draft state, I would like to merge it.

AlexanderSkrock:
I happily declare to contribute under Apache 2.0 license regarding this and future pull request. Should I state it somewhere else or signing a CLA or alike? I did not find something within the contribution guidelines. :D

In additional to the legal topic, I just finalized my pull request. Also I tested it against a private repo of mine and was able to prove the check as well as the output mapping-

In case of no changes, the overview looks like this:
image
The details view shows the skipped step, but not its reason, sadly I do not know of an option to realize this. The log of the newly introduced detection job clearly states the reason, tho.
image

For the sake of showing all states, here's a screenshot for detected changes as well:
image

kthoms:
We haven't set up a CLA so far. Until we have that I am just asking for consense of each first-time contributor.

Thanks a lot for stepping in, this is what I have intended.

Before merging, maybe you could make the build appear cancelled? I'm wondering what is better: To see a green build that actually did nothing, or see cancelled builds. WDYT?

AlexanderSkrock:
Do you have an idea how I could achieve this? From within a workflow run it seems currently (almost) impossible to cancel via the api. Also, I read that github blocks this action using the standard `$GITHUB_TOKEN`. Introducing a controlling workflow that only defines the schedule and then calls the `.github/workflows/integration-build.yml` conditionally, but this would still result in a green nighty build while not triggering a standard integration workflow.

kthoms (approved):

kthoms:
Then let's not spend energy on this (now). I was just wondering if this could be an improvement, but I don't know either if this is possible with reasonable effort. I know from Jenkins that you could achieve that in a workflow, but GitHub I did not research yet.

So again, many thanks for your contribution! If you have yourself ideas to improve the build, just open issues or direct PRs.

PR #893, opened 2025-06-16
build(deps): Upgrade to Spring 6.2.8

kthoms:
Resolves a vulnerability in spring-web.

See https://github.com/operaton/operaton/security/dependabot/317


kthoms:
Integratin build is green

Issue #798, opened 2025-05-13
Remove Telemetry code


kthoms:
**What needs to be done? Please describe.**

Operaton does not support Telemetry. Most related API is already marked deprecated. The task is to clean up the Operaton code base from Telemetry related code.

ArneDeutsch:
To my understanding this involves:
- engine
  - Remove all classes inside `org.operaton.bpm.engine.telemetry`
  - Remove all classes inside `org.operaton.bpm.engine.impl.telemetry.dto`
  - Remove methods that return these instances, modify the ones that use the instances (e.g. `ManagementService.toggleTelemetry`, `ManagementService.isTelemetryEnabled` and `ManagementService.getTelemetryData` ... the first two are actually deprecated, the last one is not ... wonder if this is by accident!?)
  - Remove SQL entries related to telemetry from SQL scripts in `operaton\engine\src\main\resources\org\operaton\bpm\engine\db\liquibase\**` (e.g. `insert into ACT_GE_PROPERTY values ('telemetry.lock', '0', 1);`)
  - Remove/modify test code that is related to Telemetry (e.g. `ManagementAuthorizationTest.shouldGetTelemetryDataAsOperatonAdmin`)
- engine-rest
  - Remove `org.operaton.bpm.engine.rest.dto.TelemetryConfigurationDto`, `org.operaton.bpm.engine.rest.TelemetryRestService`, `org.operaton.bpm.engine.rest.impl.TelemetryRestServiceImpl` and methods that return these types.
  - Remove `TelemetryRestServiceTest`
- engine-rest-openapi
  - Remove the templates in `operaton\engine-rest\engine-rest-openapi\src\main\templates\models\org\operaton\bpm\engine\rest\dto\telemetry` and `operaton\engine-rest\engine-rest-openapi\src\main\templates\paths\telemetry`
  - Modify other templates that return/use the deleted templates
- webapp
  - Modify `ProcessEnginesFilter.setWebappInTelemetry`, `ServletContextUtil.isTelemetryDataSentAlready` and related code to remove telemetry there
- wildfly-subsystem
  - Remove/Mopdify `ManagedJtaProcessEngineConfiguration.initTelemetryData`
- qa-simple
  - Modify `SimpleApplicationIT` to use a different class for the test

Do you agree that this is more or less the task? Or did I get something wrong?

kthoms:
[...]

PR #904, opened 2025-06-17
WIP

ArneDeutsch:
relates to #798

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/diagnostics/DiagnosticsCollector.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/diagnostics/DiagnosticsCollector.java
@@ -1,0 +1,76 @@
+package org.operaton.bpm.engine.impl.diagnostics;
+
+import static org.operaton.bpm.engine.management.Metrics.ACTIVTY_INSTANCE_START;
+import static org.operaton.bpm.engine.management.Metrics.EXECUTED_DECISION_ELEMENTS;
+import static org.operaton.bpm.engine.management.Metrics.EXECUTED_DECISION_INSTANCES;
+import static org.operaton.bpm.engine.management.Metrics.ROOT_PROCESS_INSTANCE_START;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import org.operaton.bpm.engine.impl.metrics.Meter;
+import org.operaton.bpm.engine.impl.metrics.MetricsRegistry;
+import org.operaton.bpm.engine.impl.metrics.util.MetricsUtil;
+
+/**
+ * Collects metrics and command counts for diagnostic purposes.
+ */
+public class DiagnosticsCollector {
+
+  protected static final Set<String> METRICS_TO_REPORT = Set.of(
+      ROOT_PROCESS_INSTANCE_START,
+      EXECUTED_DECISION_INSTANCES,
+      EXECUTED_DECISION_ELEMENTS,
+      ACTIVTY_INSTANCE_START);
+
+  protected DiagnosticsRegistry diagnosticsRegistry;
+  protected MetricsRegistry metricsRegistry;
+
+  public DiagnosticsCollector(DiagnosticsRegistry diagnosticsRegistry,
+                              MetricsRegistry metricsRegistry) {
+    this.diagnosticsRegistry = diagnosticsRegistry;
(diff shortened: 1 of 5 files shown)


ArneDeutsch:
This PR is for discussion. It needs to be reviewed if the stuff that is removed by this is actually not be needed any more.

engine/src/main/java/org/operaton/bpm/engine/impl/diagnostics/DiagnosticsCollector.java
javahippie, line 1: Not to be removed, diagnostics is not telemetry:

https://docs.operaton.org/docs/documentation/user-guide/process-engine/diagnostics-data/

(Will not add this comment to other files in a "diagnostics" package

ArneDeutsch: Question: the DiagnosticsCollector uses classes form `org.operaton.bpm.engine.impl.telemetry.dto`. These classes are to be deleted. Right? For example the field `staticData` (Type: `org.operaton.bpm.engine.impl.telemetry.dto.TelemetryDataImpl`) will be deleted:

```
public class DiagnosticsCollector {
  protected TelemetryDataImpl staticData;
```

ArneDeutsch:
From the documentation I can see the following:

```
ProcessEngine processEngine = ...;
ManagementService managementService = processEngine.getManagementService();

TelemetryData telemetryData = managementService.getTelemetryData();
Internals productInternals = telemetryData.getProduct().getInternals();

String jdkVendor = productInternals.getJdk().getVendor();
String jdkVersion = productInternals.getJdk().getVersion();
```

Here TelemetryData is from the package `org.operaton.bpm.engine.telemetry`. I thought this should be removed ... but maybe I was wrong.

In `ManagementService` there are three related methods. I have removed all of them. Now I start to realize ... I guess the third one has to remain while the first to should go away. Right?

```
  /**
   * @deprecated The sending telemetry feature is removed.
   * Please remove the method references as they are no longer needed.
   */
  void toggleTelemetry(boolean enabled);

  /**
   * @deprecated The sending telemetry feature is removed.
   * Please remove the method references as they are no longer needed.
   * @return always <code>false</code>
   */
  Boolean isTelemetryEnabled();

  /**
   * This method returns the current state of the collected telemetry data. For a
   * detailed explanation on what is included, read the
   * <a href="https://docs.operaton.org/manual/latest/introduction/telemetry/#collected-data">Telemetry</a>
   * page in the Operaton docs.
   *
   * The engine collects information about multiple aspects of the installation.
   *
   * @see <a href="https://docs.operaton.org/manual/latest/introduction/telemetry/#collected-data">Operaton Documentation: Telemetry</a>
   */
  TelemetryData getTelemetryData();
```

In result the entire package `org.operaton.bpm.engine.telemetry` would stay as it is.

ArneDeutsch:
I also wonder if we should actually REMOVE the telemetry related code at all. For example we have marked a related method to be removed

```
  /**
   * @deprecated The sending telemetry feature is removed.
   * Please remove the method references as they are no longer needed.
   */
  void toggleTelemetry(boolean enabled);
```

In this case it does not return a type from the telemetry package, but most probably uses them internally. Should we delete the method right away? Should we remain with the method, but delete the contained functionality (no-op-method)? Or is deletion of the functionliaty actually not wanted, but in fact we just want to mark the stuff as deprecated!?

javahippie:
The issue is, that while Camunda move the non-telemetry code which can be used for internal diagnostics to a new package, they had to keep all exitsting interfaces (both REST and Java) in place to not break the API. It's entirely possible that there are methods which are no longer needed, but I agree that we will not get to a point where it is safe to delete those classes entirely (at least not for version 1.x). Maybe it's safe to assume, that Camunda removed everything that was safe to delete already with the commit we backported?

javahippie:
Any further thoughts on this? I think either way if we decided to remove parts of the telemetry API, this would be a candidate for a release later than 1.0, as this would be a breaking change.

ArneDeutsch:
Thinking about this I would postpone further removal of the telemetry stuff. It is marked as deprecated already and we are not exactly sure yet what to remove. Better to keep a bit of leagacy than removing to much. Suggest to close this PR and mark the ticket as post-1.0.

javahippie:
Agree, the ticket is already marked as `1.x-backlog`, we will take another look at it, then

PR #910, opened 2025-06-17
refactor: Make JobHandler#onDelete default method

kthoms:
Remove empty implementations

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/jobexecutor/AsyncContinuationJobHandler.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/jobexecutor/AsyncContinuationJobHandler.java
@@ -26,7 +26,6 @@ import org.operaton.bpm.engine.impl.context.Context;
 import org.operaton.bpm.engine.impl.interceptor.CommandContext;
 import org.operaton.bpm.engine.impl.jobexecutor.AsyncContinuationJobHandler.AsyncContinuationConfiguration;
 import org.operaton.bpm.engine.impl.persistence.entity.ExecutionEntity;
-import org.operaton.bpm.engine.impl.persistence.entity.JobEntity;
 import org.operaton.bpm.engine.impl.pvm.PvmActivity;
 import org.operaton.bpm.engine.impl.pvm.process.TransitionImpl;
 import org.operaton.bpm.engine.impl.pvm.runtime.LegacyBehavior;
@@ -166,8 +165,4 @@ public class AsyncContinuationJobHandler implements JobHandler<AsyncContinuation
 
   }
 
-  @Override
-  public void onDelete(AsyncContinuationConfiguration configuration, JobEntity jobEntity) {
-    // do nothing
-  }
 }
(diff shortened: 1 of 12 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/jobexecutor/AsyncContinuationJobHandler.java
```diff
@@ -117,8 +116,8 @@ protected String[] tokenizeJobConfiguration(String jobConfiguration) {
 
     if (jobConfiguration != null ) {
       String[] configParts = jobConfiguration.split("\\$");
-      if (configuration.length > 2) {
-        throw new ProcessEngineException("Illegal async continuation job handler configuration: '" + jobConfiguration + "': exprecting one part or two parts seperated by '$'.");
+      if (configParts.length > 2) {
```
javahippie: Did this fix a bug on the side? It looks like the empty array initialized with a size of two was always checked instead of the interpreted `configParts`. If this does fix a bug, I'd be more content with an additional unit test

engine/src/test/java/org/operaton/bpm/engine/test/jobexecutor/TweetNestedCommandExceptionHandler.java
```diff
@@ -41,13 +40,8 @@ public String getType() {
...
   @Override
   public void execute(JobHandlerConfiguration configuration, ExecutionEntity execution, CommandContext commandContext, String tenantId) {
-    Context.getProcessEngineConfiguration().getCommandExecutorTxRequired().execute(new Command<Void>() {
-
-      @Override
-      public Void execute(CommandContext commandContext) {
-        throw new RuntimeException("nested command exception");
-      }
-
+    Context.getProcessEngineConfiguration().getCommandExecutorTxRequired().execute((Command<Void>) commandContext1 -> {
```
javahippie: I think it's good that the local `commandContext` was renamed to not shadow the method parameter, but not the biggest fan of numbering variable names. This is a minor nitpick, though

engine/src/main/java/org/operaton/bpm/engine/impl/jobexecutor/AsyncContinuationJobHandler.java
kthoms: Yes, this fixes a bug. I'll add a test for it.

engine/src/test/java/org/operaton/bpm/engine/test/jobexecutor/TweetNestedCommandExceptionHandler.java
kthoms: I'll rename that.

engine/src/main/java/org/operaton/bpm/engine/impl/jobexecutor/AsyncContinuationJobHandler.java
kthoms: Actually, I think it is better to provide a separate PR for it. Otherwise this fix would be hidden by the squash commit.

javahippie (approved):

Issue #901, opened 2025-06-17
Make CmmnTransformListener methods to empty default methods


kthoms:
**What needs to be done? Please describe.**

`CmmnTransformListener` defines methods that have to be implemented by each implementation. The implementations will use just a subset to actually override the methods. By making the methods to empty default methods implementors only need to override methods that they actually fill.

Make all methods to empty default methods:

```
void transformRootElement(Definitions definitions, List<? extends CmmnCaseDefinition> caseDefinitions);
```

```
default void transformRootElement(Definitions definitions, List<? extends CmmnCaseDefinition> caseDefinitions) {}
```

This will make `AbstractCmmnTransformListener` obsolete. Delete the methods within, but do not delete the class. Mark it deprecated for removal (Javadoc & Annotation).

Look for subclasses of `AbstractCmmnTransformListener` and let them directly implement `CmmnTransformListener` instead of extending `AbstractCmmnTransformListener`.

Search for all implementors of `CmmnTransformListener` and remove all empty overridden methods.

PR #911, opened 2025-06-18
refactor(engine): Make CmmnTransformListener methods default.

ArneDeutsch:
Depracate AbstractCmmnTransformListener.
Remove empty methods.

relates to #901


engine/src/main/java/org/operaton/bpm/engine/impl/cmmn/transformer/AbstractCmmnTransformListener.java
```diff
@@ -16,79 +16,14 @@
...
-
 /**
  * Abstract base class for implementing a {@link CmmnTransformListener} without being forced to implement
  * all methods provided, which make the implementation more robust to future changes.
  *
  * @author Sebastian Menski
+ *
+ * @deprecated Will be removed, implement CmmnTransformListener directly in your code
  */
+@Deprecated(forRemoval = true)
```
kthoms:
```suggestion
@Deprecated(forRemoval = true, since = "1.0")
```

Issue #899, opened 2025-06-17
Make BpmnParseListener methods to empty default methods


kthoms:
**What needs to be done? Please describe.**

`BpmnParseListener` defines methods that have to be implemented by each implementation. The implementations will use just a subset to actually override the methods. By making the methods to empty default methods implementors only need to override methods that they actually fill.

Make all methods to empty default methods:

```
void parseProcess(Element processElement, ProcessDefinitionEntity processDefinition);
```

```
default void parseProcess(Element processElement, ProcessDefinitionEntity processDefinition) {}
```

This will make `AbstractBpmnParseListener` obsolete. Delete the methods within, but do not delete the class. Mark it deprecated for removal (Javadoc & Annotation).

Look for subclasses of `AbstractBpmnParseListener` and let them directly implement `BpmnParseListener` instead of extending `AbstractBpmnParseListener`.

Search for all implementors of `BpmnParseListener` and remove all empty overridden methods.

PR #912, opened 2025-06-18
refactor(engine): Default methods for BpmnParseListener.

ArneDeutsch:
relates to #899


engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/AbstractBpmnParseListener.java
```diff
@@ -27,163 +27,16 @@
...
+ * Abstract base class for implementing a {@link BpmnParseListener} without being forced
+ * to implement all methods provided, which makes the implementation more robust to
+ * future changes.
  *
  * @author ruecker
+ *
+ * @deprecated Will be removed in a future version. Implement BpmnParseListener directly.
  */
-@SuppressWarnings("java:S1186")
+@Deprecated(forRemoval = true)
```
kthoms:
```suggestion
@Deprecated(forRemoval = true, since = "1.0")
```

PR #927, opened 2025-06-21
chore(engine): bump spring dependency to version 6 (#4952)

cachescrubber:
* chore(engine): bump spring dependency to version 6

related to camunda/camunda-bpm-platform#4869

Backported commit c8374f26a2 from the camunda-bpm-platform repository.
Original author: Joaquín <165814090+joaquinfelici@users.noreply.github.com>

### Notes

I skipped most of the commit. Most notable, they decided to remove the MockInitialContextRule with the [Simple-JNDI](https://github.com/h-thurow/Simple-JNDI) dependency. Operaton already migrated that code to Junit 5, so I think we are fine without another test scoped dependency.


javahippie:
Thanks, looks good to me!

PR #932, opened 2025-06-23
fix: AsyncContinuationJobHandler#tokenizeJobConfiguration did not fail for invalid argument

kthoms:
Exception was not raised when given jobConfiguration contained more than 2 segments. The wrong array was compared.

Added unit test.

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -246,7 +246,7 @@
 
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
+      <artifactId>mockito-junit-jupiter</artifactId>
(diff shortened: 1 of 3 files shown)


javahippie (approved):
lgtm. I was a little confused about the returned array containing nulls, but it seems that throughout the engine there is a fallback in place if the atomic operaton inside Job Config is null?

kthoms:
It is the behaviour of `newConfiguration()` invoked with `null`. I doubt that this makes sense in practice and would assume that `canonicalString` must not be `null`. However, it is currently possible.

Enforcement of not nullable arguments is a larger topic.

For now, the test documents the actual behaviour.

PR #946, opened 2025-06-27
Trigger integration build on pom change

kthoms:
Extends the "build" workflow to trigger the "Integration Build" workflow when a change to a pom.xml is detected.


kthoms:
I will work on it for RC-1

kthoms:
Will be resolved via #1243

PR #947, opened 2025-06-27
Rolling database update

kthoms:
Fixes the rolling database updates build.


kthoms:
Rolling Database Update tests are executed as part of the Integration Build.

Bildschirmfoto 2025-07-03 um 07 35 50
Bildschirmfoto 2025-07-03 um 07 35 31

PR #953, opened 2025-06-30
fix: Enable DB2 Tests on Wildfly and Tomcat

### LINKED ISSUE #554: Enable Integration Tests for DB2
- [x] Able to run locally
- [x] Able to run in the pipeline

javahippie:
Resolves #554


javahippie:
Successful integration tests: https://github.com/operaton/operaton/actions/runs/15979495649

database/pom.xml
```diff
@@ -113,6 +113,10 @@
         <database.type>db2</database.type>
         <database.driver>com.ibm.db2.jcc.DB2Driver</database.driver>
         <database.datasource.class>com.ibm.db2.jcc.DB2SimpleDataSource</database.datasource.class>
+        <jboss.datasource.filename>db2-ds.xml</jboss.datasource.filename>
+        <database.tc.url>operatondb2:12.1.2.0</database.tc.url>
```
kthoms, line 117: Is it necessary to encode the version in the URL?

qa/integration-tests-engine/src/test/java/org/operaton/bpm/integrationtest/util/ArquillianEventObserver.java
```diff
@@ -44,11 +45,15 @@ public class ArquillianEventObserver {
   private static final String MARIADB = "mariadb";
   private static final String MARIADB_VERSION = "10.0";
 
+  private static final String DB2 = "db2";
+  private static final String DB2_VERSION = "12.1.2.0";
```
kthoms, line 49: Could this be replaced by a placeholder and replaced at build time by resource filtering?

qa/wildfly-runtime/src/main/common/modules/com/ibm/icu/icu4j/main/module.xml
kthoms, line 1: There is already an ICE module at `distro/wildfly/modules/src/main/modules/com/ibm/icu/icu4j/main/module.xml`
Please remove use the existing

test-utils/testcontainers/src/main/java/org/operaton/impl/test/utils/testcontainers/OperatonDb2ContainerProvider.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
```
kthoms: Operaton copyright header?

javahippie: Good catch

database/pom.xml
javahippie: Yes, this is getting passed to the ContainerProvider (for non-Arquillian tests). The database identifier and version are used by the Testcontainers JDBC driver to determine which container to use.

qa/integration-tests-engine/src/test/java/org/operaton/bpm/integrationtest/util/ArquillianEventObserver.java
javahippie: I think this area has room for improvement. Currently we are only able to run one database version per provider, which I don't think is ideal, if we want to maintain Operaton for a selection of versions.

I would propose extracting this change request into its own issue, as it would require modifying all other database versions, too, and would leave the scope of this issue / PR

database/pom.xml
kthoms: ok

kthoms (approved):
LGTM

PR #964, opened 2025-07-03
Remove trailing comma in VALID_DATABASES

javahippie:
Causes a build error, because "mariadb" is not recognized as a valid value


javahippie:
See https://github.com/operaton/operaton/actions/runs/16060801785/job/45326135778#step:5:6312 for example

Issue #564, opened 2025-03-03
Replace try/catch with expected failure in test code


kthoms:
**Is your feature request related to a problem? Please describe.**
When thrown exceptions are tested a typical pattern is to use try/catch with an empty catch block and failing the test with `fail()` when the code under test does not throw the exception.

For example:
```java
    try {
      runtimeService.getVariable(instance.getId(), "simpleBean");
      fail("Exception expected.");
    }
    catch(Exception e) {
      // happy path
    }
```

**Describe the solution you'd like**
Operaton uses AssertJ, which provides a powerful fluent assertion API.

The replacement for above mentioned code would be:
```java
    // given
    var processInstanceId = instance.getId();

    // when
    assertThatThrownBy(() -> runtimeService.getVariable(processInstanceId, "simpleBean"))
      // then
      .isInstanceOf(Exception.class);
```

Note that the lambda must should be limited to the call under test. The call's arguments must be separated to a variable that is created in the "given" block.

Search for `fail("Exception expected` in the code base for code to refactor. There are more usages of the pattern which is not directly detected because the message differs. Search for usages of the `org.assertj.core.api.Assertions#fail(java.lang.String)` method .

**Implementation instructions**

- replace usages of `testRule.assertTextPresent()` by `hasMessageContaining()`
- use method references as argument to `assertThatThrownBy()` when possible
- avoid multiple call in lambdas as argument of `assertThatThrownBy()` that could cause exceptions. Refactor by extracting all but the last statement as a variable to the "given" part. Use "var" to declare the variable. If possible, use a method reference for the remaining statement passed to `assertThatThrownBy()`
- when using `extracting` with a lambda that uses a type cast of the exception argument, call `.asInstanceOf(type(<Excpeption>.class))` before
- this is a long refactoring. do not mention in PRs that the issue is closed/fixed by a PR.
- when a refactored test does not expect an exception, make sure an assertion on the executed code is done using `expectThatCode(...).doesNotThrowAnyException()`
- messages should not be compared ignoring case. Change to match the right case.
- Do not refactor test classes
  - WebappsDatabaseNamingConsistencyTest
  - SecurityFilterRulesTest

**Describe alternatives you've considered**
Using JUnit 5's `org.junit.jupiter.api.Assertions#assertThrows(java.lang.Class<T>, org.junit.jupiter.api.function.Executable)` is a valid alternative when
a) the test class is already migrated to JUnit 5
b) it is only of interest which type of exception is thrown

[...]

Issue #915, opened 2025-06-18
Build job to clear caches


ArneDeutsch:
**What needs to be done? Please describe.**

Currently many build jobs fail because of cache misses. It seems this is related to the fact, that our build caches grow to much / are not cleaned.

![Image](https://github.com/user-attachments/assets/a95fc416-04f9-489e-80af-939fc2c27911)

As only commiters can clear the caches, and this is manual reoccuring work, it would be nice to have an automated job that clears these caches on regular bases (nightly or something).

kthoms:
In a first step we could change the existing builds to delete the "build-artifacts" caches after they used them. These caches are just used to decouple build jobs. A first job performs the build and the output is stored in the "build-artifacts" cache, the "publish" job then restores the cache to extract the build results and publish them. After the cache entries have been restores the cache could be deleted.

kthoms:
Let's observe if this is enough already. I think the main pain point is solved.

ArneDeutsch:
Just checked. Looked like the cache remains under 10GB. Let's check again after another day or two, but looks fine for me now.

kthoms:
I observed that the delete action causes trouble for builds from contributors, like [here](https://github.com/operaton/operaton/actions/runs/15994978109/job/45121164505?pr=961).

We might have to revert this change or make sure that the delete action is only performed within the "operaton/operaton" repo.

ArneDeutsch:
Yes, seems to be a general issue: https://github.com/operaton/operaton/actions/runs/16080778635/job/45448311077?pr=970

kthoms:
I did not observe any issues with the build caches in the past weeks. Thus I consider this as fixed.

PR #970, opened 2025-07-04
#564 Replace try/catch with expected failure in test code

biswa1shaw:
Replace try/catch with expected failure in test code

- this is one of the multiple future PR's coming for this feature 

related to #564

### DIFF
--- a/commons/typed-values/src/test/java/org/operaton/bpm/engine/test/variables/FileValueTypeImplTest.java
+++ b/commons/typed-values/src/test/java/org/operaton/bpm/engine/test/variables/FileValueTypeImplTest.java
@@ -150,7 +150,7 @@ class FileValueTypeImplTest {
     properties.put("mimeType", null);
     properties.put("encoding", "someEncoding");
 
-    assertThatThrownBy(() -> type.createValue(file, properties), "expected exception")
+    assertThatThrownBy(() -> type.createValue(file, properties))
         .isInstanceOf(IllegalArgumentException.class)
         .hasMessageContaining("The provided mime type is null. Set a non-null value info property with key 'filename'");
 
@@ -160,7 +160,7 @@ class FileValueTypeImplTest {
     properties.put("mimeType", "someMimetype");
     properties.put("encoding", null);
 
-    assertThatThrownBy(() -> type.createValue(file2, properties), "expected exception")
+    assertThatThrownBy(() -> type.createValue(file2, properties))
         .isInstanceOf(IllegalArgumentException.class)
         .hasMessageContaining("The provided encoding is null. Set a non-null value info property with key 'encoding'");
   }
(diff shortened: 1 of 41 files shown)


ArneDeutsch:
I think  the build issue is related to #915

commons/typed-values/src/test/java/org/operaton/bpm/engine/test/variables/FileValueTypeImplTest.java
```diff
@@ -144,35 +144,25 @@ void createValueWithProperties() {
...
 
-    // when
-    try {
-      type.createValue(file, properties);
-      fail("expected exception");
-    } catch (IllegalArgumentException e) {
-      // then
-      assertThat(e.getMessage()).contains("The provided mime type is null. Set a non-null value info property with key 'filename'");
-    }
+    assertThatThrownBy(() -> type.createValue(file, properties), "expected exception")
```
kthoms: The message "expected exception" does not add value anymore. Please remove all occurrences.
```suggestion
    assertThatThrownBy(() -> type.createValue(file, properties))
```

commons/utils/src/test/java/org/operaton/commons/utils/EnsureUtilTest.java
```diff
@@ -41,14 +41,8 @@ void ensureNotNull() {
...
     String string = null;
-
-    try {
-      EnsureUtil.ensureNotNull("string", string);
-      fail("Expected: IllegalArgumentException");
-
-    } catch(IllegalArgumentException e) {
-      // happy path
-    }
+    assertThatThrownBy(() -> EnsureUtil.ensureNotNull("string", string), "Expected: IllegalArgumentException")
```
kthoms: Also here: The description does not add value, thus remove. Therefore in general: Remove the obvious messages that have been added to `fail()`. It is one of the benefits of this refactoring that this is not needed.

engine/src/test/java/org/operaton/bpm/engine/impl/util/ImmutablePairTest.java
```diff
@@ -73,12 +74,9 @@ void shouldCompareWithRightSecond() {
   void shouldFailWithNonComparableTypes() {
     final ImmutablePair<Object, Object> pair1 = new ImmutablePair<>(new Object(), new Object());
     final ImmutablePair<Object, Object> pair2 = new ImmutablePair<>(new Object(), new Object());
-    try {
-      pair1.compareTo(pair2);
-      fail("Pairs should not be comparable");
-    } catch (IllegalArgumentException e) {
-      assertThat(e).hasMessageContaining("Please provide comparable elements");
-    }
+    assertThatThrownBy(() -> pair1.compareTo(pair2), "Pairs should not be comparable")
```
kthoms, line 76: Here is an example where the description could stay. Do you notice the difference?

kthoms (changes requested):
Mostly ok. I hope you get my feedback. It is about leaving out obvious code. Where a description adds value it could stay as you refactored.

biswa1shaw:
acknowledged. i will submit a revision

kthoms:
@biswa1shaw Please press "Re-request review" when you are ready

PR #991, opened 2025-07-16
Fix HistoricDecisionInstanceStatisticsQuery should throws NullValueException

### LINKED ISSUE #438: HistoricDecisionInstanceStatisticsQuery should throw for null decision
**Describe the bug**
The test `org.operaton.bpm.engine.test.api.history.HistoricDecisionInstanceStatisticsQueryTest#testStatisticForRootDecisionWithNullInstanceConstraintEvaluation()` is buggy. It should fail when a null value is used as `decisionInstanceId`

```
    HistoricDecisionInstanceStatisticsQuery query = historyService
        .createHistoricDecisionInstanceStatisticsQuery(
            decisionRequirementsDefinition.getId())
        .decisionInstanceId(null);

    //then
    try {
      query.count();
    } catch (NullValueException e) {
      //expected
    }
```

The test code passes when the expected `NullValueException` is thrown, but also when `query.count()` does not throw an exception. And this is actually the case.

This was uncovered by changing the code to do an assertion:
```
assertThatThrownBy(query::count)
      .isInstanceOf(NullValueException.class);
```     

**To Reproduce**
Execute the mentioned test.

**Expected behavior**
Executing the query should throw `NullValueException`.

SevDan:
Add decisionInstanceId null check

- add processInstanceId property null check
- enable related test

closes #438

### DIFF
--- a/engine/src/test/java/org/operaton/bpm/engine/test/api/history/HistoricDecisionInstanceStatisticsQueryTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/test/api/history/HistoricDecisionInstanceStatisticsQueryTest.java
@@ -146,7 +146,6 @@ class HistoricDecisionInstanceStatisticsQueryTest {
   }
 
   @Test
-  @Disabled("Should throw exception, but does not. See https://github.com/operaton/operaton/issues/438")
   void testStatisticForRootDecisionWithNullInstanceConstraintEvaluation() {
     // given
     decisionService.evaluateDecisionTableByKey(DISH_DECISION)
--- a/engine/src/test/java/org/operaton/bpm/engine/test/api/history/HistoricDecisionInstanceStatisticsQueryTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/test/api/history/HistoricDecisionInstanceStatisticsQueryTest.java
@@ -27,6 +27,7 @@ import org.operaton.bpm.engine.DecisionService;
 import org.operaton.bpm.engine.HistoryService;
 import org.operaton.bpm.engine.ProcessEngineConfiguration;
 import org.operaton.bpm.engine.RepositoryService;
+import org.operaton.bpm.engine.exception.NotValidException;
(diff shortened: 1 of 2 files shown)


SevDan:
Changed NullValueException to NotValidException for consistency with other Query implementations. Also property checking was moved to definition step (processInstanceId is nullable, but for null value it must be 'untouched' because fluid setter and pojo setter are validated now)

engine/src/test/java/org/operaton/bpm/engine/test/api/history/HistoricDecisionInstanceStatisticsQueryTest.java
```diff
@@ -156,17 +156,15 @@ void testStatisticForRootDecisionWithNullInstanceConstraintEvaluation() {
...
-            decisionRequirementsDefinition.getId())
-        .decisionInstanceId(null);
-
-    // then
-    assertThatThrownBy(query::count)
-      .isInstanceOf(NullValueException.class);
-
-    assertThatThrownBy(query::list)
-      .isInstanceOf(NullValueException.class);
+    assertThatThrownBy(() -> {HistoricDecisionInstanceStatisticsQuery query = historyService
```
kthoms: Please refactor to a lambda where the single call raises the exception. Likely here extract the query creation to the given-part. Which call leads to the exception? count() or list()?

Sonar raises warnings when multiple calls are made within the lambda.

kthoms (changes requested):
In general ok, just refactor the test

engine/src/test/java/org/operaton/bpm/engine/test/api/history/HistoricDecisionInstanceStatisticsQueryTest.java
SevDan: Fixed: removed calls because it throws exception on query builder now (like in other queries)

kthoms:
Thank you for this fix

PR #914, opened 2025-06-18
refactor(build): Harmonize all pom.xml files.

ArneDeutsch:
relates to #649

PR #1009, opened 2025-07-24
refactor(tests): Migrate integration tests to JUnit 5

ArneDeutsch:
related to #27


ArneDeutsch:
I executed the integration tests locally. Four fail, but they are the same as for this pull request https://github.com/operaton/operaton/pull/914 . Hence I would consider the run successfull.

javahippie:
Hi @ArneDeutsch 

since we added you as a project participant, you can directly branch inside of the repo instead of forking to your private account. If you do this, you can run the integration test pipeline directly on the branch instead of executing it multiple times on your local machine and save some time by running everything at once. Maybe helpful in the future

kthoms (approved):

PR #1013, opened 2025-07-25
refactor: Simplify single-element annotation

timtebeek:
Hi! Figured explore if there's any small improvements to be made against operaton using recipes. This seemed like good candidate of easy to review yet still nice to have improvement, and can then of course also be enforced going forward if desired.

Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.java.SimplifySingleElementAnnotation?organizationId=MWZkNWMwNTgtZDgyZi00YTFjLWE1OTEtMWUxMWE2MTNjMDgw

PR #1020, opened 2025-07-27
refactor: Java 17

timtebeek:
Continuing some small fixes as detailed on 
- https://github.com/operaton/operaton/pull/1013#issuecomment-3120559800

In this case I excluded the `@Serial` and `String.formatted` changes to reduce the noise, and focus on the many smaller changes. Those two could optionally be applied separately, or not at all and enforced through a customized Java 17 recipe.


timtebeek:
Apologies! We had the same idea here today @kthoms ; great to see! I'll go ahead and close this one as I only noticed after resolving conflicts that you had run these earlier.

kthoms:
@timtebeek I took your initial change to have a look at the changes the Java 17 upgrade does. It produces much changes, so I looked at the effective recipes that do changes. Now it is rather clean. As you said the Serial annotations added created noise, and I'm overall wondering what actually needs to be serializable and what not. The codebase seems not clean in regards to that.
I'm really looking forward to our interview on Wednesday :)

timtebeek:
> I'm really looking forward to our interview on Wednesday :)

Same here! Always nice to see & hear how recipes are used, and you have the perfect use case for it. In that light you might also want to briefly evaluate
- https://github.com/operaton/migrate-from-camunda-recipe/pull/124

PR #1028, opened 2025-07-30
Resolve sonar issues

kthoms:
This PR contains refactorings to address multiple Sonar findings.

See the individual commits for changes.

### DIFF
--- a/engine-spring/src/main/java/org/operaton/bpm/engine/spring/components/aop/ProcessStartingMethodInterceptor.java
+++ b/engine-spring/src/main/java/org/operaton/bpm/engine/spring/components/aop/ProcessStartingMethodInterceptor.java
@@ -164,8 +164,8 @@ public class ProcessStartingMethodInterceptor implements MethodInterceptor {
         Map<ProcessVariable, Object> vars = this.mapOfAnnotationValues(ProcessVariable.class, invocation);
 
         Map<String, Object> varNameToValueMap = new HashMap<>();
-        for (ProcessVariable processVariable : vars.keySet()) {
-            varNameToValueMap.put(processVariable.value(), vars.get(processVariable));
+        for (var entry : vars.entrySet()) {
+            varNameToValueMap.put(entry.getKey().value(), entry.getValue());
         }
         return varNameToValueMap;
 
--- a/engine/src/test/java/org/operaton/bpm/engine/test/standalone/history/FullHistoryTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/test/standalone/history/FullHistoryTest.java
@@ -1184,9 +1184,6 @@ class FullHistoryTest {
     variables.put("stringVar", "stringValue");
     variables.put("booleanVar", true);
     Date date = Calendar.getInstance().getTime();
-    Calendar otherCal = Calendar.getInstance();
-    otherCal.add(Calendar.DAY_OF_MONTH, 1);
-    Date otherDate = otherCal.getTime();
     variables.put("dateVar", date);
     variables.put("nullVar", null);
 
--- a/juel/src/main/java/org/operaton/bpm/impl/juel/AstBinary.java
+++ b/juel/src/main/java/org/operaton/bpm/impl/juel/AstBinary.java
@@ -92,7 +92,8 @@ public class AstBinary extends AstRightValue {
 	};
 
 	private final Operator operator;
-	private final AstNode left, right;
+	private final AstNode left;
+	private final AstNode right;
 
 	public AstBinary(AstNode left, AstNode right, Operator operator) {
 		this.left = left;
--- a/juel/src/main/java/org/operaton/bpm/impl/juel/AstBinary.java
+++ b/juel/src/main/java/org/operaton/bpm/impl/juel/AstBinary.java
@@ -131,6 +131,10 @@ public class AstBinary extends AstRightValue {
 
   @Override
   public AstNode getChild(int i) {
-		return i == 0 ? left : i == 1 ? right : null;
+		return switch (i) {
+			case 0 -> left;
+			case 1 -> right;
+			default -> null;
+		};
 	}
 }
--- a/juel/src/main/java/org/operaton/bpm/impl/juel/AstChoice.java
+++ b/juel/src/main/java/org/operaton/bpm/impl/juel/AstChoice.java
@@ -19,7 +19,9 @@ import jakarta.el.ELContext;
 import jakarta.el.ELException;
 
 public class AstChoice extends AstRightValue {
-	private final AstNode question, yes, no;
+	private final AstNode question;
+	private final AstNode yes;
+	private final AstNode no;
 
 	public AstChoice(AstNode question, AstNode yes, AstNode no) {
 		this.question = question;
--- a/juel/src/main/java/org/operaton/bpm/impl/juel/AstChoice.java
+++ b/juel/src/main/java/org/operaton/bpm/impl/juel/AstChoice.java
@@ -56,6 +56,11 @@ public class AstChoice extends AstRightValue {
 
   @Override
   public AstNode getChild(int i) {
-		return i == 0 ? question : i == 1 ? yes : i == 2 ? no : null;
+		return switch (i) {
+			case 0 -> question;
+			case 1 -> yes;
+			case 2 -> no;
+			default -> null;
+		};
 	}
 }
--- a/juel/src/main/java/org/operaton/bpm/impl/juel/AstMethod.java
+++ b/juel/src/main/java/org/operaton/bpm/impl/juel/AstMethod.java
@@ -114,7 +114,11 @@ public class AstMethod extends AstNode {
 
   @Override
   public Node getChild(int i) {
-		return i == 0 ? property : i == 1 ? params : null;
+		return switch (i) {
+			case 0 -> property;
+			case 1 -> params;
+			default -> null;
+		};
 	}
 
 	@Override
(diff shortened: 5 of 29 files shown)


webapps/assembly/src/main/java/org/operaton/bpm/webapp/impl/security/auth/Authentications.java
```diff
@@ -50,10 +50,7 @@ public class Authentications implements Serializable {
   /**
    * holds an entry for each processEngine->userId pair currently authenticated
    */
-  protected Map<String, UserAuthentication> authentications = new HashMap<>();
-
-  public Authentications() {
-  }
+  private static final Map<String, UserAuthentication> ENGINE_TO_USERAUTHENTICATION = new HashMap<>();
```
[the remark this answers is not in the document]
kthoms: refactored

webapps/assembly/src/main/java/org/operaton/bpm/webapp/impl/filter/AbstractTemplateFilter.java
```diff
@@ -100,13 +99,7 @@ protected boolean hasWebResource(String name) {
...
   protected String getWebResourceContents(String name) throws IOException {
 
-    InputStream is = null;
-
-    try {
-      is = filterConfig.getServletContext().getResourceAsStream(name);
-
-      BufferedReader reader = new BufferedReader(new InputStreamReader(is));
-
+    try (var reader =  new BufferedReader(new InputStreamReader(filterConfig.getServletContext().getResourceAsStream(name)))) {
```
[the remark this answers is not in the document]
kthoms: refactored

juel/src/main/java/org/operaton/bpm/impl/juel/AstBinary.java
```diff
@@ -131,6 +131,10 @@ public int getCardinality() {
 
   @Override
   public AstNode getChild(int i) {
-		return i == 0 ? left : i == 1 ? right : null;
+		return switch (i) {
+			case 0 -> left;
+			case 1 -> right;
+			default -> null;
+		};
 	}
```
ArneDeutsch, line 139: I would prefer the more simple/basic:

```
if (i == 0)
  return left;
if (i == 1)
  return right;
return null;
```

The ternary and the switch are to difficult to read for such basic code. But it is opinion, only change if you agree.

engine-spring/src/main/java/org/operaton/bpm/engine/spring/components/aop/ProcessStartingMethodInterceptor.java
```diff
@@ -164,8 +164,8 @@ protected Map<String, Object> processVariablesFromAnnotations(MethodInvocation i
         Map<ProcessVariable, Object> vars = this.mapOfAnnotationValues(ProcessVariable.class, invocation);
 
         Map<String, Object> varNameToValueMap = new HashMap<>();
-        for (ProcessVariable processVariable : vars.keySet()) {
-            varNameToValueMap.put(processVariable.value(), vars.get(processVariable));
+        for (var entry: vars.entrySet()) {
```
ArneDeutsch: Picky: add a blank right before the colon ... multiple occurrences in code, will not complain for the other instances.

juel/src/main/java/org/operaton/bpm/impl/juel/AstChoice.java
```diff
@@ -56,6 +56,11 @@ public int getCardinality() {
 
   @Override
   public AstNode getChild(int i) {
-		return i == 0 ? question : i == 1 ? yes : i == 2 ? no : null;
+		return switch (i) {
```
ArneDeutsch, line 59: Same as before ... I would prefer simple if statements a lot ...

```
if (i == 0)
  return question:
if (i == 1)
  ...
```

juel/src/main/java/org/operaton/bpm/impl/juel/AstMethod.java
```diff
@@ -114,7 +114,11 @@ public int getCardinality() {
 
   @Override
   public Node getChild(int i) {
-		return i == 0 ? property : i == 1 ? params : null;
+		return switch (i) {
```
ArneDeutsch, line 117: if-statements ....

engine/src/test/java/org/operaton/bpm/engine/test/standalone/history/FullHistoryTest.java
```diff
@@ -1186,7 +1186,6 @@ void testHistoricProcessInstanceVariableValueNotEquals() {
     Date date = Calendar.getInstance().getTime();
     Calendar otherCal = Calendar.getInstance();
```
ArneDeutsch: This looks suspicious ... why is `otherCal` there in the first place? Looks like it is not used at all. This line and the following could technically be deleted as well ... but maybe the test case is screwed!? Before deleting the unused variables double check the test case and if it is still testing what it is supposed to test.

webapps/assembly/src/main/java/org/operaton/bpm/webapp/impl/filter/AbstractTemplateFilter.java
```diff
@@ -95,20 +93,14 @@ protected boolean hasWebResource(String name) {
...
-      while ((line = reader.readLine()) != null) {
-        writer.write(line);
-        writer.append("\n");
+    try (var is = filterConfig.getServletContext().getResourceAsStream(name)) {
+      if (is == null) {
+        throw new IOException("Resource not found: " + name);
       }
-
-      return writer.toString();
+      return IoUtil.inputStreamAsString(is);
```
ArneDeutsch, line 103: This modification will change behaviour because of different newline handling. I don't know if it is critical, but in the old implementation all newlines will be replaced with '\n' while in the new version the line endings are preserved as is. This might have downstream consequences to users. Intentional? Acceptable?

webapps/assembly/src/test/java/org/operaton/bpm/webapp/impl/filter/AbstractTemplateFilterTest.java
```diff
@@ -0,0 +1,70 @@
+package org.operaton.bpm.webapp.impl.filter;
```
ArneDeutsch: Missing file header.

webapps/assembly/src/test/java/org/operaton/bpm/webapp/impl/filter/AbstractTemplateFilterTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class AbstractTemplateFilterTest {
```
ArneDeutsch: Starting the test class name with `Abstract` is confusing because we usually use this to indicate test classes that do not execute directly but are extended by other, non-abstract, test classes instead. Propose to rename it wo `TemplateFilterTest` and add a short JavaDoc comment with is tested.

juel/src/main/java/org/operaton/bpm/impl/juel/AstBinary.java
kthoms: I would leave it as is. When using if, I would try to avoid to use it without braces. 
```
if (i == 0) {
  return left;
}
if (i == 1) {
  return right;
}
return null;
```
But this would be less readable.

Further, I consider switch-pattern-matching as "modern" style.

In the end a matter of taste. Of course your proposal is also simple and valid.

webapps/assembly/src/test/java/org/operaton/bpm/webapp/impl/filter/AbstractTemplateFilterTest.java
kthoms: renamed

webapps/assembly/src/test/java/org/operaton/bpm/webapp/impl/filter/AbstractTemplateFilterTest.java
kthoms: thx, changed

engine-spring/src/main/java/org/operaton/bpm/engine/spring/components/aop/ProcessStartingMethodInterceptor.java
kthoms: changed, seems to be the more common style

webapps/assembly/src/main/java/org/operaton/bpm/webapp/impl/filter/AbstractTemplateFilter.java
kthoms: It is acceptable. The only known usage is `org.operaton.bpm.webapp.impl.engine.ProcessEnginesFilter#serveIndexPage(java.lang.String, java.lang.String, java.lang.String, java.lang.String, jakarta.servlet.http.HttpServletResponse, jakarta.servlet.ServletContext)`

It serves the index.html page to the browser. The browser does not care.

Anyway, good catch and was worth to rethink.

engine/src/test/java/org/operaton/bpm/engine/test/standalone/history/FullHistoryTest.java
kthoms: deleted the other 2 lines, they did not make sense

ArneDeutsch (approved):

PR #1058, opened 2025-08-14
fix(tomcat): Remove WebListener annotation

kthoms:
This reverts c8e1735edbf475bf21b2c41366e100b914292735

The annotation lead to duplicate registration of the InvoiceProcessApplication as listener.


kthoms:
Integration build succeeded

https://github.com/operaton/operaton/actions/runs/16962831799

PR #1070, opened 2025-08-19
feat: Database migration tests with PostgreSQL, MySQL, MariaDB, Oracl…

jmaster1985:
…e and MS SQL Server

to run database migrations tests, first navigate to the `qa` directory, from there execute `./db-migration-test.sh [DB_NAME]`, where `[DB_NAME]` can have one of the following values:

- postgres
- mysql
- mariadb
- oracle
- sqlserver

please be aware, that the tests with `sqlserver` would fail - because of the migrations themselves, not the test script that executes the migrations.

expected is, that the database containers are removed after the execution of the `db-migration-test.sh` , regardless of a successful or unsuccessful test.

for awareness: @kthoms


kthoms:
Thank you @jmaster1985 for picking this up. I will soon have a look!

.github/workflows/migration-test.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Docker
+        uses: docker/setup-buildx-action@v3
+
+      - name: Run QA script
+        run: |
+          ./mvnw clean install -DskipTests -PskipFrontendBuild
```
jmaster1985, line 32: @kthoms : I don't exactly know when this workflow is going to be triggered, therefore I just built the software before actually testing the migrations. Maybe we can use a pre-built artifact to speed up the migration-tests

.github/workflows/migration-test.yml
```diff
@@ -0,0 +1,30 @@
...
+      - main
+
+jobs:
+  qa:
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        db: [mysql, mariadb, postgres, oracle, sqlserver]
```
jmaster1985, line 21: @kthoms : currently the tests for Ms SQL Server are failing (please see the logs in this [job](https://github.com/operaton/operaton/actions/runs/17070678899/job/48398492941)) - maybe one can remove `sqlserver` from this matrix to keep the tests green, and create an issue when this PR is merged, so the developer knows how to test the migrations for Ms SQL Server

.github/workflows/migration-test.yml
kthoms: It is totally fine to perform a fast rebuild for now.

kthoms: But a separate step would be fine for it

.github/workflows/migration-test.yml
```diff
@@ -0,0 +1,30 @@
...
+        uses: actions/checkout@v4
+
+      - name: Set up Docker
+        uses: docker/setup-buildx-action@v3
+
+      - name: Run QA script
+        run: |
+          ./mvnw clean install -DskipTests -PskipFrontendBuild
+          cd qa
+          chmod +x ./db-migration-test.sh
```
kthoms: The script is already executable
```-rwxrwxrwx   1 codespace root  663 Aug 19 16:05 db-migration-test.sh````
remove the line

File: qa/test-database-container/mariadb/docker-compose.yaml
Comment (kthoms) [inline qa/test-database-container/mariadb/docker-compose.yaml:1]: I think the `docker-compose.yaml` and `qa-xxx` files can be consolidated.

Use a single compose file with all services defined, and just start the desired service.

The qa script can be generified.

In the end I would expect to have just a 3 files.

File: .github/workflows/migration-test.yml (hunk 2 above)
Comment (kthoms) [reply .github/workflows/migration-test.yml:21]: It is fine that this fails. I expect a fix for the SQL Server scripts.

File: qa/test-database-container/mariadb/docker-compose.yaml
Comment (kthoms) [reply qa/test-database-container/mariadb/docker-compose.yaml:1]: Looks better now!

File: database/pom.xml [hunk 4]
```diff
@@ -310,6 +310,212 @@
...
+      <properties>
+        <database.type>mariadb</database.type>
+        <database.driver>org.mariadb.jdbc.Driver</database.driver>
+        <hibernate.dialect>org.hibernate.dialect.MariaDBDialect</hibernate.dialect>
+        <database.datasource.class>org.mariadb.jdbc.MariaDbDataSource</database.datasource.class>
+        <database.port>${env.MT_MARIADB_HOST_PORT}</database.port>
+        <database.name>${env.MT_MARIADB_DATABASE}</database.name>
+        <database.username>${env.MT_MARIADB_USERNAME}</database.username>
+        <database.password>${env.MT_MARIADB_PASSWORD}</database.password>
+        <database.params>${database.name}?user=${database.username}&amp;amp;password=${database.password}</database.params>
```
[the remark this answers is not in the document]
jmaster1985: actually maven can't handle a single `&amp;` and complains about a missing `;` after the entity. It parses the XML and then substitutes properties, so the extra escaping ensures the final string is correctly interpreted as `&` in the generated URL.

database/pom.xml
```diff
@@ -310,6 +310,212 @@
...
+        <database.type>postgres</database.type>
+        <database.driver>org.postgresql.Driver</database.driver>
+        <hibernate.dialect>org.hibernate.dialect.PostgreSQLDialect</hibernate.dialect>
+        <database.datasource.class>org.postgresql.ds.PGSimpleDataSource</database.datasource.class>
+        <jboss.datasource.filename>postgresql-ds.xml</jboss.datasource.filename>
+        <database.port>${env.MT_POSTGRES_HOST_PORT}</database.port>
+        <database.name>${env.MT_POSTGRES_DATABASE}</database.name>
+        <database.username>${env.MT_POSTGRES_USERNAME}</database.username>
+        <database.password>${env.MT_POSTGRES_PASSWORD}</database.password>
+        <database.params>${database.name}?user=${database.username}&amp;amp;password=${database.password}</database.params>
```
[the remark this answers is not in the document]
jmaster1985: see the comment above

PR #1085, opened 2025-08-20
chore(deps): Upgrade Liquibase to 4.33.0

kthoms:
Streamline version between liquibase-maven-plugin and liquibase dependency by moving version.liquibase to Root POM


kthoms:
[Integration build succeeded](https://github.com/operaton/operaton/actions/runs/17111151526)

Issue #1055, opened 2025-08-12
Refactor Property*Key classes to records


kthoms:
**What needs to be done?**

The classes

- org.operaton.bpm.engine.impl.core.model.PropertyKey
- org.operaton.bpm.engine.impl.core.model.PropertyListKey
- org.operaton.bpm.engine.impl.core.model.PropertyMapKey

seem to be convertible to records without loose of functionality or breaking any contract.

Discuss if the assumption is true and if so, refactor. Keep the license statement.

The type parameter seem obsolete in these classes, as detected by[ Sonar](https://sonarcloud.io/project/issues?impactSoftwareQualities=MAINTAINABILITY&rules=java%3AS2326&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton&open=AZLaQn0v8ZwKcH34-VI4).

Bonus: Identify other classes that have record-like semantics.

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it. 

Are you a first-time contributor? We are happy to see you here! 
Just state in a comment of your PR "I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md)"

Contributors will be added to the release notes of the next release.

herminagarg:
@kthoms Hi, I'm a first-time contributor. Can I take this issue?

khawaja-abdullah:
Hey @kthoms I have a couple of questions and would really appreciate you addressing them;
- Since you mentioned that the generic type parameter is obsolete in these classes, would it make sense to remove it entirely?
- If that’s the case, then once converted to records, all three classes would essentially look the same — should they be consolidated into a single record type and reused everywhere instead?

kthoms:
@khawaja-abdullah Sorry for coming back delayed. I was not available on the weekend.

1) Yes, if the type key seems obsolete. Double check and remove it when you come to the same conclusion.
2) For now I would keep the 3 separate. They have the same content, but different semantic. I see potential consolidating them, but this would be a separate story. But thank you for pointing this out, I will think about that.

khawaja-abdullah:
@kthoms no worries at all, thanks for your response.

[...]

PR #1097, opened 2025-08-23
refactor(class-to-record): Refactor Property*Key classes to records

khawaja-abdullah:
Closes https://github.com/operaton/operaton/issues/1055


kthoms (approved):
I was not expecting the refactoring from `getName` to `name`. An additional getter would have made this obsolete. But actually this is now better. Since the classes are inside `impl` packages, this change is valid.

Issue #1127, opened 2025-08-26
Create ADR to decide on null safety solution


ungerts:
**What needs to be done?**

- Create an ADR (Architecture Decision Record) to decide on a null safety solution for Operaton’s Java APIs.
- Include decision, rationale, trade-offs, and rollout plan for public APIs first.

**Additional context**

Operaton currently relies on runtime checks (`EnsureUtil`, `Objects.requireNonNull`) without a standardized, tool-friendly way to declare nullability intent.

javahippie:
I guess this can be closed because #1131 was already closed, right?

kthoms:
Yes, can be closed now.

PR #1131, opened 2025-08-26
ADR-0001 Null Safety Strategy for Operaton

ungerts:
ADR: Null Safety Strategy for Operaton — first proposal for discussion

- Adds docs/decisions/0001-null-safety-strategy-for-operaton.md.
- Status: Proposed; first discussion draft.
- Scope: Docs only; no production code changes.
- Branch: adr/0001-nullable.
- Reference: Refs: #1127.

Please review with focus on:

- Goals, scope, and consequences.
- Coverage of options and rationale.
- Consistency with the existing ADR format.

### DIFF
--- a/docs/decisions/0001-null-safety-strategy-for-operaton.md
+++ b/docs/decisions/0001-null-safety-strategy-for-operaton.md
@@ -63,110 +63,54 @@ Compliance will be confirmed via code review and CI checks:
 
 ### General JSpecify considerations (applies to options 2–4)
 
-#### Pros
-
-- Public API contracts become explicit; IDEs/analyzers can detect violations earlier.
-- Alignment with tooling and documentation; annotations surface in Javadoc and IDE tooltips.
-- Can reduce redundant runtime null checks on hot paths once confidence is established.
-
-#### Cons
-
-- Requires migration effort and contributor familiarity with annotations and tool configuration.
-- Potential annotation noise without sensible defaults (e.g., package-level `@NullMarked`).
-
-#### Neutral
-
+- Good, because public API contracts become explicit; IDEs/analyzers can detect violations earlier.
+- Good, because annotations surface in Javadoc and IDE tooltips, aligning with tooling and documentation.
+- Good, because redundant runtime null checks can be reduced on hot paths once confidence is established.
+- Bad, because adoption requires migration effort and contributor familiarity with annotations and tool configuration.
+- Bad, because annotation noise can occur without sensible defaults (e.g., package-level `@NullMarked`).
 - Neutral, because a compile-only dependency means no runtime/transitive impact.
 - Neutral, because adding annotations is binary-safe; changing contracts (nullable ↔ non-null) is a behavioral change.
 - Neutral, because conventions for collections/Optional element nullability can be established.
 
 ### Status Quo — Runtime checks only, no nullability annotations
 
-#### Pros
-
-- No immediate migration cost; preserves current behavior and CI setup.
-
-#### Cons
-
-- Ambiguity about nullability persists; limited design-time guidance and tool-assisted safety.
-- Runtime overhead from defensive checks remains.
-
-#### Neutral
-
+- Good, because it has no immediate migration cost and preserves current behavior and CI setup.
 - Neutral, because runtime checks can still catch nulls at module boundaries.
+- Bad, because ambiguity about nullability persists; limited design-time guidance and tool-assisted safety.
+- Bad, because runtime overhead from defensive checks remains.
 
 ### JSpecify for New Public APIs Only
 
-#### Pros
-
-- Minimizes churn while adding clarity for newly added surfaces.
-- Allows gradual reduction of runtime checks on hot paths for new code.
-
-#### Cons
-
-- Existing APIs remain ambiguous for a long time; inconsistent experience for users.
-
-#### Neutral
-
-- Neutral, because mixed coverage can be acceptable temporarily during transition.
+- Good, because it minimizes churn while adding clarity for newly added surfaces.
+- Good, because runtime null checks can be reduced on hot paths for new code.
+- Neutral, because mixed annotation coverage can be acceptable temporarily during transition.
+- Bad, because existing APIs remain ambiguous for a long time; inconsistent experience for users.
 
 ### JSpecify for All Public APIs
 
-#### Pros
-
-- Clear, consistent contracts at API boundaries where they matter most.
-- Consistent experience for users across public APIs.
-
-#### Cons
-
-- Requires a focused sweep of public APIs and careful handling where exception semantics must be preserved.
-
-#### Neutral
-
+- Good, because it sets clear, consistent contracts at API boundaries where they matter most.
+- Good, because it provides a consistent experience for users across public APIs.
 - Neutral, because internals can adopt annotations gradually.
+- Bad, because it requires a focused sweep of public APIs and careful handling where exception semantics must be preserved.
 
 ### JSpecify Project-Wide
 
-#### Pros
-
-- Maximizes clarity and static analysis coverage across the codebase.
-
-#### Cons
-
-- Highest migration cost and risk of initial noise/false positives.
-
-#### Neutral
-
+- Good, because it maximizes clarity and static analysis coverage across the codebase.
 - Neutral, because it may be a long-term aspiration after initial adoption on public APIs.
+- Bad, because it has the highest migration cost and risk of initial noise/false positives.
 
 ### Defer Adoption — Revisit after a defined period/criteria
 
-#### Pros
-
-- Avoids immediate migration cost while gathering more ecosystem/tooling data.
-
-#### Cons
-
-- Delays benefits to users; increases future migration burden.
-
-#### Neutral
-
+- Good, because it avoids immediate migration cost while gathering more ecosystem/tooling data.
 - Neutral, because this is distinct from Status Quo: an explicit, time-bound decision to postpone, with revisit triggers (date, tooling maturity, or dependency adoption milestones).
+- Bad, because it delays benefits to users and increases future migration burden.
 
 ### Use Another Annotation Framework — JetBrains, Checker Framework, SpotBugs/FindBugs, JSR 305, Eclipse JDT
 
-#### Pros
-
-- Familiar options with existing tooling in some ecosystems; Checker Framework offers powerful type-checking.
-
-#### Cons
-
-- Potential divergence from emerging ecosystem consensus; mixed semantics across frameworks.
-- JSR 305 is legacy/unmaintained and may cause ambiguity in tooling behavior.
-
-#### Neutral
-
+- Good, because these are familiar options with existing tooling in some ecosystems; Checker Framework offers powerful type-checking.
 - Neutral, because per-module interop may require a particular framework.
+- Bad, because they may diverge from emerging ecosystem consensus and introduce mixed semantics across frameworks.
+- Bad, because JSR 305 is legacy/unmaintained and may cause ambiguity in tooling behavior.
 
 ## More Information
 


docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+
+Null safety has long been a challenge in Java. Operaton currently enforces non-null assumptions primarily at runtime using `EnsureUtil` and `Objects.requireNonNull`, but lacks a standardized, tool-friendly way to declare nullability intent across APIs, which causes ambiguity for contributors and users of the public API. JSpecify is emerging as the ecosystem's preferred approach to express nullability contracts (standardized `@Nullable`/non-null annotations, good IDE/static analysis support, ecosystem alignment such as Spring Framework 7). JSpecify is developed by consensus of major stakeholders in the Java ecosystem (e.g., Google, JetBrains, Microsoft, Oracle, Meta, Uber, Broadcom/Spring), indicating strong industry backing. Project Valhalla’s future null-related features are out of scope for the current baseline and do not remove the need for annotations in the near term. These capabilities are expected to arrive incrementally across multiple JDK releases. Operaton currently targets Java 17, so it will likely take several years until Operaton runs on a JDK that fully supports Valhalla, keeping annotation-based nullability necessary for the foreseeable future.
```
kthoms: The text in this section frames already towards JSpecify and against Valhalla instead of focusing on the problem.

IMO the text here should be more open, and discuss more the problem itself.

Discuss JSR 305

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+- Do not rely on preview/incubating language features.
+- Align with major ecosystem dependencies (e.g., Spring 7) to reduce friction for users.
+
+## Considered Options
+
+1. Status Quo — Keep `EnsureUtil` and `Objects.requireNonNull`, no nullability annotations.
+2. JSpecify for New Public APIs Only — Annotate only newly introduced public APIs going forward.
+3. JSpecify for All Public APIs (Selected) — Annotate all public-facing APIs; internals optional/gradual.
+4. JSpecify Project-Wide — Annotate public and internal codebase; highest migration cost.
+5. Defer Adoption — Revisit when JSpecify/Valhalla matures further.
```
kthoms: This is actually the same as "Status Quo", isn't it?

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+
+- Improve clarity and safety of Operaton’s public APIs regarding nullability.
+- Leverage static tooling to catch issues earlier.
+- Keep migration cost reasonable and avoid accidental external behavior changes during refactors.
+- Do not rely on preview/incubating language features.
+- Align with major ecosystem dependencies (e.g., Spring 7) to reduce friction for users.
+
+## Considered Options
+
+1. Status Quo — Keep `EnsureUtil` and `Objects.requireNonNull`, no nullability annotations.
```
kthoms: also: Javadocs could be added for more clarity on the behavior

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+2. JSpecify for New Public APIs Only — Annotate only newly introduced public APIs going forward.
```
kthoms: The main decision option is: Use JSpecify yes/no
Then the scope. Maybe 2 a,b,c?

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+3. JSpecify for All Public APIs (Selected) — Annotate all public-facing APIs; internals optional/gradual.
+4. JSpecify Project-Wide — Annotate public and internal codebase; highest migration cost.
+5. Defer Adoption — Revisit when JSpecify/Valhalla matures further.
+6. Use Another Annotation Framework (e.g., JetBrains) — Diverges from ecosystem direction; similar migration cost.
```
kthoms: Framing. This should be discussed in the Pros/Cons of the option

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+
+- 100% of public APIs annotated with JSpecify.
+- CI flags nullability violations on public APIs.
+- No externally visible behavior changes (e.g., exception types) in the engine module due to refactors.
+
+## Pros and Cons of the Options
+
+### Status Quo — Keep EnsureUtil/Objects.requireNonNull, no annotations
+
+- Good, because it has no immediate migration cost and preserves current behavior and CI setup.
```
kthoms, line 77: Don't like "Good","Bad". Can this be rephrased to "Pro","Contra" or "Positive","Negative"?

Could the options formatted differently, e.g. having subheaders or in tables (might be hard to read in textual display)?

docs/decisions/0001-null-safety-strategy-for-operaton.md
kthoms: Considering certain frameworks with reasonable adoption as extra options:
https://checkerframework.org/
com.google.code.findbugs:jsr305 
Spotbugs annotations (https://www.javadoc.io/doc/com.github.spotbugs/spotbugs-annotations/3.1.0/edu/umd/cs/findbugs/annotations/Nullable.html)
Eclipse JDT

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+- Good, because it minimizes churn and adds clarity for newly added surfaces.
+- Good, because runtime null checks can be reduced on hot paths to cut overhead.
+- Good, because null safety is part of the contract/API.
+- Neutral, because mixed annotation coverage can be acceptable temporarily.
+- Bad, because it leaves existing APIs ambiguous for a long time; inconsistent experience for users.
+- Bad, because of inconsistency in null safety across public APIs.
+
+### JSpecify for All Public APIs (Selected)
+
+- Good, because it sets clear, consistent contracts at API boundaries where they matter most; strong tool support; ecosystem alignment.
```
kthoms: Strong tool support is not a pro of this option, but for JSpecify itself. Therefore I would consider general JSpecify Pro/Cons and then discuss the scope.

Maybe the scope is even a separate ADR?

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+- Bad, because it has the highest migration cost and risk of noise/false positives initially.
+
+### Defer Adoption
+
+- Good, because it avoids immediate cost and waits for further maturation.
+- Neutral, because current runtime checks remain in place.
+- Bad, because it delays benefits to users; misses ecosystem alignment; increases future migration burden.
+- Bad, because imitating null safety using `Objects.requireNonNull` creates runtime overhead.
+
+### Use Another Annotation Framework (e.g., JetBrains)
```
kthoms: I would distinguish between IDE centric alternatives (Jetbrains, Eclipse) and independent ones (e.g. Spotbugs)
The support of IDEs for other frameworks than their own is possible, but requires additional configuration.

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,105 @@
...
+
+- Good, because it provides similar expressiveness with existing tooling support.
+- Neutral, because many teams are familiar with these annotations.
+- Bad, because it diverges from emerging ecosystem consensus and Spring 7 alignment; offers no clear advantage over JSpecify.
```
kthoms: IMO it is still too early to judge the "ecosystem consensus". Spring's adoption is very important, but JSpecify is still away from being a de-facto standard. This is a bet into the future.

shubhamos-ai (approved):
🎯 Perfect implementation! The separation of concerns is clear, the code is DRY, and the error messages are helpful. This follows industry best practices beautifully.

ungerts:
Thanks for the feedback. I’ll work on an updated version. Plan is to have something ready by Wednesday.

kthoms:
The update reads already quite good. We need more input now. @javahippie what is your opinion? Others: Feel free to join the discussion.

docs/decisions/0001-null-safety-strategy-for-operaton.md
```diff
@@ -0,0 +1,125 @@
...
+## Pros and Cons of the Options
+
+### General JSpecify considerations (applies to options 2–4)
+
+- Good, because public API contracts become explicit; IDEs/analyzers can detect violations earlier.
+- Good, because annotations surface in Javadoc and IDE tooltips, aligning with tooling and documentation.
+- Good, because redundant runtime null checks can be reduced on hot paths once confidence is established.
+- Bad, because adoption requires migration effort and contributor familiarity with annotations and tool configuration.
+- Bad, because annotation noise can occur without sensible defaults (e.g., package-level `@NullMarked`).
+- Neutral, because a compile-only dependency means no runtime/transitive impact.
```
javahippie, line 71: This does not seem to be true, if we include JSpeficy in our library, we also ship it as a transient dependency with all downsides: 

https://mastodon.social/@rotnroll666/115180147816231735

michael-simons:
Because @javahippie summoned me: It does not end with adding JSpecify. You will also want to have a build that uses https://github.com/uber/NullAway or similar, otherwise you have the nice annotations and it gives you: Nothing.

Here's the full change in SDN https://github.com/spring-projects/spring-data-neo4j/commit/7fcacb79911630e07f43963e644e1b1aec815907

For reference, here's a similar discussion for Jackson:

https://github.com/FasterXML/jackson-future-ideas/discussions/87#discussioncomment-13383776

docs/decisions/0001-null-safety-strategy-for-operaton.md
michael-simons: If the guy who wrote the toot could think, he would have written runtime. It's `@Retention(RUNTIME)` ;) I fixed that in the toot, otherwise @kthoms would be correct.

kthoms:
Thank you @michael-simons for bringing your valuable experience into the discussion! It is great to have these insights to this decision.

You write: 
> when done right JSpecify and NullAway are great changes, but they are IMHO pricy

Our aim is to make the code base future proof and best-of-breed. This comes to a price, and we are willing to pay it.

An important question is: Could we incrementally change the code base? I assume so. We did (and are still doing) such a big move with the JUnit 4->5 migration. This is spanning already several (beta) releases. At the end it does not matter how long it takes, as clients do not recognize which tests are JUnit 4 based and which JUnit 5. They just work. If we can do this also with JSpecify/NullAway then the size of the task does not matter much. When we have a (huge?) quality gain at the end, then it is worth.

michael-simons:
Sure, if you accept it right from the start todo in two steps, this is a valid approach. For us in SDN this was a no go: adding the annotations itself is just a marker, no verification. To benefit from it, you want to have proper analysis on top of it, so that you introduce proper null checks. This is btw mentioned in one of the annotations: there will be a lot of places in which an analysis can’t always determine if a value is null or not, so objects requirenonnull and friends are still due. 

Having said that: I am a fan of the combined outcome but now that I have the checks in place thanks to the second step I can’t wait for Java to have null ability in the type system proper to kick out step 1 again. 

https://openjdk.org/jeps/8303099

*Ofc this would be breaking API, which in turn reminds me of:*
Jspecify annotations do have an effect on the kotlin compiler. Nonnull types will become nonnull kotlin types, hence breaking the api for kotlin users. You should be aware of that.

ungerts:
I'm a little bit afraid that we lose focus here. Please find below my comments:

- Tool support: For me this is an important criterion. It’s good that tools like NullAway and IntelliJ inspections exist. However, integrating NullAway should be a follow-up once we have decided on a null-safety solution. The decision itself must stand on its own.

- Future type system: This is currently too far out to play a role in this ADR. Operaton has a minimum requirement of Java 17, and it’s not even foreseeable when we’ll be on Java 25 (and that will also depend on Spring, Quarkus, … as discussed in [operaton 2.0 scope](https://forum.operaton.org/t/operaton-2-0-scope/295/18)).We should also avoid using preview functionality, since it creates significant maintenance effort.

- Kotlin support and new type system: Both may introduce breaking changes. For Operaton 2.0 I think we should allow breaking changes if needed. Otherwise, it will be hard to adopt changes in dependent libraries (like Spring) and to introduce new features. This was also discussed in the [2.0 scope thread](https://forum.operaton.org/t/operaton-2-0-scope/295/18).

- Template / Good vs. Bad: We already agreed on MADR, and that gives us a clear structure for pros/cons. Deviating from this would partially revise the decision for MADR.

Thanks a lot for the great discussion so far!

kthoms:
My feeling: 3 core contributors except the author. The actual process is not defined yet. I would like to move forward. The ADR is in proposed state anyway. I'll ask on Slack.

Issue #962, opened 2025-07-02
Resolve com.sun.jersey.contribs:jersey-apache-client4 dependency


kthoms:
**What needs to be done? Please describe.**

Library `com.sun.jersey.contribs:jersey-apache-client4` is outdated (1.9.17 released in 2017).

Analyze the usages and replace by a recent library.

**Additional context**

The following dependencies exist:

```
org.operaton.bpm.qa:operaton-qa-integration-tests-webapps:jar:1.0.0-beta-5-SNAPSHOT
+- com.sun.jersey.contribs:jersey-apache-client4:jar:1.19.4:provided
|  \- com.sun.jersey:jersey-client:jar:1.19.4:provided
|     \- com.sun.jersey:jersey-core:jar:1.19.4:provided
|        \- javax.ws.rs:jsr311-api:jar:1.1.1:provided

org.operaton.bpm.run:operaton-bpm-run-qa-runtime:jar:1.0.0-beta-5-SNAPSHOT
+- com.sun.jersey.contribs:jersey-apache-client4:jar:1.19.4:test
|  \- com.sun.jersey:jersey-client:jar:1.19.4:test
|     \- com.sun.jersey:jersey-core:jar:1.19.4:test
|        \- javax.ws.rs:jsr311-api:jar:1.1.1:test
```

tlb-lemrabott:
Hey @kthoms, 

I would like to work on this. Sending PR soon...

kthoms:
This is now resolved via #1175.

The dependencies do not contain any dependency for group id `com.sun.jersey.contribs` anymore.

PR #1134, opened 2025-08-27
chore(deps): Remove dependency on org.apache.httpcomponents:httpclient/httpcore

tlb-lemrabott:
- Replaced outdated jersey-apache-client4 with jersey-apache-connector 3.1.10
- Updated dependency management in BOM and 4 test modules
- Eliminated javax namespace usage for Jakarta EE migration

Issue #962

### DIFF
--- a/bom/internal-dependencies/pom.xml
+++ b/bom/internal-dependencies/pom.xml
@@ -444,12 +444,6 @@
       </dependency>
 
       <dependency>
-        <groupId>org.glassfish.jersey.connectors</groupId>
-        <artifactId>jersey-apache-connector</artifactId>
-        <version>${version.jersey3}</version>
-      </dependency>
-
-      <dependency>
         <groupId>org.jboss.resteasy</groupId>
         <artifactId>resteasy-jaxrs</artifactId>
         <version>${version.resteasy}</version>
--- a/distro/run/qa/runtime/pom.xml
+++ b/distro/run/qa/runtime/pom.xml
@@ -98,11 +98,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpclient</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.springframework.boot</groupId>
       <artifactId>spring-boot-starter-test</artifactId>
       <scope>test</scope>
--- a/engine-rest/engine-rest/pom.xml
+++ b/engine-rest/engine-rest/pom.xml
@@ -155,11 +155,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.httpcomponents</groupId>
-      <artifactId>httpcore</artifactId>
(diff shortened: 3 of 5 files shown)


bom/internal-dependencies/pom.xml
```diff
@@ -423,7 +423,15 @@
       <dependency>
         <groupId>org.glassfish.jersey.core</groupId>
         <artifactId>jersey-client</artifactId>
-        <version>${version.jersey3}</version>
+        <version>3.1.10</version>
```
kthoms: Use the version property`version.jersey3`here
Why did the scope changed to `provided`? I would not declare this in the dependency management already. Leave this to the usages.
An exception to declare default scopes are:
- `test` for dependencies that should usually only be used in tests
- `provided` for dependencies that are usually provided by containers, e.g. Jakarta APIs

qa/integration-tests-webapps/pom.xml
```diff
@@ -51,9 +51,9 @@
         <version>4.5.14</version>
       </dependency>
       <dependency>
-        <groupId>com.sun.jersey.contribs</groupId>
-        <artifactId>jersey-apache-client4</artifactId>
-        <version>1.19.4</version>
+        <groupId>org.glassfish.jersey.connectors</groupId>
+        <artifactId>jersey-apache-connector</artifactId>
+        <version>3.1.10</version>
```
kthoms: The parent POM already imports the BOM operaton-core-internal-dependencies
This should make the declaration of a specific version obsolete
Rule of thumb: We try to avoid declaring versions in specific POMs as far as possible and have a central dependency management by the operaton-core-internal-dependencies BOM for that. This way we can assure that we use the same version throughout the code base.

kthoms:
@tlb-lemrabott I appreciate your work on updating dependencies ❤️ This is one of our core asserts.

If you are looking for other upgrade possibilities, you don't have to wait for raised issues. Create an overview of upgrade candidates with
```
mvn versions:dependency-updates-aggregate-report
```
(results in /target/reports) 

This is basically what I'm looking at.

bom/internal-dependencies/pom.xml
tlb-lemrabott: Thanks for the information! 

I’ve made the updates accordingly.

kthoms (changes requested):
Integration tests are not successful
https://github.com/operaton/operaton/actions/runs/17287768638

The failing build executes
```
./mvnw -DskipTests -Pdistro,distro-webjar,h2-in-memory,distro-run,integration-test-operaton-run clean install
./mvnw -Pengine-integration,integration-test-operaton-run,h2 clean verify -f distro/run/qa
```

kthoms:
After rebasing the original PR has become obsolete. By backport #1175 the HTTP client has changed to unirest/HTTPComponents5.

The dependency jersey-apache-client4 has been removed with that.

The changes here are now about removal of HTTPComponents 4.
```
<groupId>org.apache.httpcomponents</groupId>
```

I'll rename this PR and plan to merge it. You remain in the commit history as co-author.

Issue #1253, opened 2025-09-08
Refactor tests that call `managementService.executeJob(jobId)` and ignore the exception


kthoms:
**What needs to be done?**

Throughout the test code there are several occurrences of a pattern like this:

```
    try {
      // when
      managementService.executeJob(jobId);
    } catch (Exception ignored) { }
```
(Exception variable name or formatting may differ)
Example: `org.operaton.bpm.engine.test.api.history.removaltime.batch.BatchSetRemovalTimeInChunksTest#shouldSetRemovalTime_ByteArray_JobLog_Hierarchical`

This repetitive code should be refactored for all occurrences. To do so:

1. Introduce a method `org.operaton.bpm.engine.impl.test.TestHelper#executeJobIgnoringException(ManagementService managementService, String jobId)`.
2. Cover the method by a test simulating that the execution of a job would throw an exception. Add the test to `org.operaton.bpm.engine.impl.test.TestHelperTest`
3. Refactor tests with the given pattern to use the new utility method.
4. Use static import of the utility method

The refactored code would look like:
```
executeJobIgnoringException(managementService, jobId);
```

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it.
Please start soon with the task when you signaled that you want to work on it and be responsive to questions or feedback.
We also do our best to review and merge your PR in a timely manner.

Are you a first-time contributor? We are happy to see you here! 
Just state in a comment of your PR "I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md).
Consider adding a star ⭐️ to the [repository](https://github.com/operaton/operaton) if you like it, we appreciate that.

Contributors will be added to the release notes of the next release.

khawaja-abdullah:
Hello @kthoms, may I take this up please?

kthoms:
Sure! Great that you pick it up. [This Sonar query](https://sonarcloud.io/project/issues?rules=java%3AS108&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton&open=AZLaQrRE8ZwKcH34-Wzy) might be a guide to find candidates.

kthoms:
Thank you. The first change was already good. I found more candidates:

[...]

PR #1256, opened 2025-09-08
refactor(tests): standardize executeJob calls with ignore-exception helper

khawaja-abdullah:
Addresses #1253


kthoms:
The failing "Database Migration Test" builds can be ignored for now. The change looks good.

Issue #1265, opened 2025-09-09
Refactor tests that call managementService.executeJob(jobId) and expect an exception


kthoms:
**What needs to be done?**

Throughout the test code there are several occurrences of a pattern like this:

```
try {
      managementService.executeJob(jobId);
      fail("exception expected");
    } catch (Exception e) {
      // expected
    }
```

Here an exception is expected and the test fails if it does not. The message "exception expected" does not add value here.

This could result into another helper method `executeJobExpectingException` to `rg.operaton.bpm.engine.impl.test.TestHelper`. The method could look like this:

```
assertThatCode(() -> managementService.executeJob(jobId))
      .isInstanceOf(ProcessEngineException.class);
```

In some cases the fail message adds value, e.g. in `JobQueryTest`:

```
try {
      managementService.executeJob(timerJobId);
      fail("RuntimeException from within the script task expected");
    } catch(RuntimeException re) {
      assertThat(re.getMessage()).contains(EXCEPTION_MESSAGE);
    }
```

An overloaded method could be provided that takes the expected message to check

```
    assertThatCode(() -> managementService.executeJob(jobId))
      .withFailMessage(EXCEPTION_MESSAGE)
      .isInstanceOf(ProcessEngineException.class);
```

**Refactoring candidates**

Use this regexp to find candidates:
```
try .*\n.*managementService\.executeJob
```

**Tests**

Add tests for the new helper methods.

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it.
Please start soon with the task when you signaled that you want to work on it and be responsive to questions or feedback.
We also do our best to review and merge your PR in a timely manner.

Are you a first-time contributor? We are happy to see you here! 
Just state in a comment of your PR "I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md).
Consider adding a star ⭐️ to the [repository](https://github.com/operaton/operaton) if you like it, we appreciate that.

Contributors will be added to the release notes of the next release.

khawaja-abdullah:
Hey @kthoms, may I take this up?

khawaja-abdullah:
Hey @kthoms, I don't see a need for the second suggested overload, here's why;

```
1. try {
[...]

PR #1271, opened 2025-09-10
refactor(tests): standardize executeJob calls with expect-exception helper

khawaja-abdullah:
Addresses #1265

### DIFF
--- a/engine/src/main/junit-shared/org/operaton/bpm/engine/impl/test/TestHelper.java
+++ b/engine/src/main/junit-shared/org/operaton/bpm/engine/impl/test/TestHelper.java
@@ -567,12 +567,8 @@ public abstract class TestHelper {
   }
 
   public static void executeJobExpectingException(ManagementService managementService, String jobId) {
-    executeJobExpectingException(managementService, jobId, Exception.class);
-  }
-
-  public static void executeJobExpectingException(ManagementService managementService, String jobId, Class<? extends Exception> clazz) {
     assertThatCode(() -> managementService.executeJob(jobId))
-        .isInstanceOf(clazz);
+        .isInstanceOf(ProcessEngineException.class);
   }
 
 }
--- a/engine/src/main/junit-shared/org/operaton/bpm/engine/impl/test/TestHelper.java
+++ b/engine/src/main/junit-shared/org/operaton/bpm/engine/impl/test/TestHelper.java
@@ -55,6 +55,8 @@ import org.operaton.bpm.engine.repository.DeploymentBuilder;
 import org.operaton.bpm.engine.test.Deployment;
 import org.operaton.bpm.engine.test.RequiredHistoryLevel;
 
+import static org.assertj.core.api.Assertions.assertThatCode;
+
 
 /**
  * @author Tom Baeyens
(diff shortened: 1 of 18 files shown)


khawaja-abdullah:
Hey @kthoms, please check my [comment](https://github.com/operaton/operaton/issues/1265#issuecomment-3275194399) on the issue as well when you take a look at this PR, thanks!

engine/src/main/junit-shared/org/operaton/bpm/engine/impl/test/TestHelper.java
```diff
@@ -564,4 +565,14 @@ public static void executeJobIgnoringException(ManagementService managementServi
       // do nothing
     }
   }
+
+  public static void executeJobExpectingException(ManagementService managementService, String jobId) {
+    executeJobExpectingException(managementService, jobId, Exception.class);
```
kthoms: Can we assume that the exception is `ProcessEngineException`. `AuthorizationException` is also a subclass. I think any other RuntimeException should be unexpected and make the caller fail.

engine/src/main/junit-shared/org/operaton/bpm/engine/impl/test/TestHelper.java
```diff
@@ -564,4 +565,14 @@ public static void executeJobIgnoringException(ManagementService managementServi
...
+  }
+
+  public static void executeJobExpectingException(ManagementService managementService, String jobId, Class<? extends Exception> clazz) {
```
kthoms: Same here, `ProcessEngineException`?

engine/src/main/junit-shared/org/operaton/bpm/engine/impl/test/TestHelper.java
khawaja-abdullah: Good observation @kthoms, the reason why I created two overloads is because from caller side I see two distinct patterns; at some places the expected exception was explicitly `ProcessEngineException` while in most of the others generic `Exception` was expected (hence why keeping any hidden intent intact made sense) - but let me update this and see how tests behave. Thank you for your input.

khawaja-abdullah: @kthoms one such scenario is depicted by this test failure https://github.com/operaton/operaton/actions/runs/17620862034/job/50066006694#:~:text=Error%3A%202%2C546%20%5BERROR%5D%20%20%20HistoricJobLogTest,java.lang.RuntimeException%3A

What do you suggest, should I revert the last 2 commits that were added assuming that only `ProcessEngineException` can be thrown from job exec? Or is there a better alternative here?

kthoms:
There are actually 2 cases that do throw a RuntimeException instead of the ProcessEngineException.

I'll check them now and make my mind about them. Thank you for the change anyway.

kthoms:
I decided that these 2 cases just use the assertion
```
assertThatCode(() -> managementService.executeJob(jobId)).isInstanceOf(RuntimeException.class);
```
directly.

I was thinking of making the new method expecting just Exception, or make the expected Exception class parametrizable, but I think it is better to just assume the more stricter type.

Furthermore, this makes me think if a `RuntimeException` should be just sent through, or signaled by a `ProcessEngineException`. This would change the behavior and this we likely should not do ATM, as it could break client expectations.

I'll do the changes now and merge when the build is successful.

kthoms (approved):

Issue #930, opened 2025-06-23
Webapp Tests not runnable for Tomcat


javahippie:
**Describe the bug**

When running the test suite in the combinatein `webapps-integration` and `tomcat`, no tests are found.

**To Reproduce**

1. `.devenv/scripts/build/build-and-run-integration-tests.sh --testsuite=webapps --distro=tomcat --db=h2`

**Expected behavior**

The tests schould be executed, they should be green-.

**Additional context**
This seems to have happened after introducing the `operaton-test-utils-testcontainers` into the dependencies of the tomcat runtime, which add Junit 5 to the classpath as a transitive dependency. 

When excluding JUnit as a transitive dependency, tests are found and executed, but they all fail with several issues (some javax.* related)

javahippie:
I managed to clean some things up already and got Selenium to work again, but now the tests are breaking because Selenium expects the default user to be present in the qa distro of Tomcat and tries to log in on the "user creation page". Continuing

![Image](https://github.com/user-attachments/assets/03acc1fc-4eaa-4704-a6fe-038e24bbe2c1)

PR #1280, opened 2025-09-11
Make Integration Tests for Web Apps runnable

javahippie:
Verfies that #930 was fixed

Summary:
* Changed default context path for Web Apps to `/operaton` from `/camunda`
* Added newest version of Maven Cargo plugin which is needed for the latest Wildfly container
* Added `operaton-commons-utils` as dependency to engine-rest, because it depends on the Collection Utils which were moved there
* Removed handling of Chromedriver versions from the build process 
* Increased logging details for Maven Cargo Plugin
* Added bash script to install Chrome and Chromedriver in the latest version in GitHub Actions

### DIFF
--- a/.github/workflows/integration-build.yml
+++ b/.github/workflows/integration-build.yml
@@ -30,7 +30,7 @@ on:
         description: 'Integration Tests: Test suite to run'
         type: choice
         required: true
-        default: '["engine"]'
+        default: '["engine", "webapps"]'
         options:
           - '["engine"]'
           - '["webapps"]'
--- a/.github/workflows/integration-build.yml
+++ b/.github/workflows/integration-build.yml
@@ -326,6 +326,12 @@ jobs:
         run: |
           .devenv/scripts/build/build-and-run-integration-tests.sh --testsuite=${{ matrix.testsuite }} --distro=${{ matrix.distro }} --db=${{ matrix.database }} --no-test
           .devenv/scripts/build/build-and-run-database-update-tests.sh --db=${{ matrix.database }} --no-test
+      - name: Install Chromedriver
+        id: install-chromedriver
+        if: ${{ matrix.testsuite == 'webapps' }}
+        shell: bash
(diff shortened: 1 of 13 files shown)


.github/workflows/integration-build.yml
```diff
@@ -30,7 +30,7 @@ on:
         description: 'Integration Tests: Test suite to run'
         type: choice
         required: true
-        default: '["engine"]'
+        default: '["engine", "webapps"]'
```
kthoms, line 33: [Not yet working](https://github.com/operaton/operaton/actions/runs/17645033633/job/50140822414)
Maybe leave out for now? The `cargo.wildfly.container.id` change is undoubted.

javahippie: The tests as a whole are not working, i'd leave it in for now and only merge if the tests are successful on GitHub Actions

javahippie: Has been resolved, the tests are now working and capable of running in GitHub Actions

qa/integration-tests-webapps/shared-engine/pom.xml
```diff
@@ -89,6 +89,8 @@
                 <home>${jboss.runtime}</home>
                 <timeout>${cargo.timeout}</timeout>
                 <output>${project.build.directory}/cargo.log</output>
+                <log>${project.build.directory}/server.log</log>
+                <logLevel>DEBUG</logLevel>
```
kthoms, line 93: still necessary or wanted?

javahippie: The logs which are produced here are rather sparse, without these settings debugging was not really possible for me. I'd like to keep these settings for that reason

kthoms: Fine for me

kthoms:
Let's go!

PR #1290, opened 2025-09-12
Enable users to start integration build scripts with mvnd

javahippie:
I'm using Maven Daemon locally on my machine to speed up builds for integration testing, often cutting the test times in half.

I added a parameter to the script, so developers can choose to use the existing scripts either with the maven wrapper (default) or with Maven daemon:

`.devenv/scripts/build/build-and-run-integration-tests.sh --testsuite=engine --distro=tomcat --db=h2 --daemon`

https://github.com/user-attachments/assets/a604bdfb-8e4e-422c-af6a-e364444d96a9

### DIFF
--- a/.devenv/scripts/build/build-and-run-integration-tests.sh
+++ b/.devenv/scripts/build/build-and-run-integration-tests.sh
@@ -5,6 +5,8 @@ EXECUTE_TEST=true
 TEST_SUITE="engine"
 DATABASE="h2"
 DISTRO="tomcat"
+RUNNER="./mvnw"
+DAEMON=false
 VALID_TEST_SUITES=("engine" "webapps" "db-rolling-update")
 VALID_DISTROS=("operaton" "tomcat" "wildfly")
 VALID_DATABASES=("h2" "postgresql" "postgresql-xa" "mysql" "mariadb" "oracle" "db2" "sqlserver")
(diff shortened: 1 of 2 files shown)


javahippie:
@kthoms Will add it to build.sh, too, good catch!

I had runner as the first option, but wondered if it included too many possibilities. Is there a motivation to use a locally installed Maven over the wrapper? Maybe for testing version compatibility? More flexibility is always nice to have, will take a shot at this.

kthoms:
```
.devenv/scripts/build/build.sh --runner mvnd
ℹ️ ./mvnw -Pdistro,distro-webjar,distro-run,distro-tomcat,tomcat,h2-in-memory --runner mvnd clean install
Unable to parse command line options: Unrecognized option: --runner
```
(side note: mvnd not installed)

javahippie:
@kthoms interesting, `.devenv/scripts/build/build.sh --runner=mvnd` (with =) works for me. I think it's handled the same as `--profile`, should we adapt both?

kthoms:
I will check later

kthoms:
My mistake. I wrote `--runner mvnd` instead of `--runner=mvnd`

Small improvement: Add
```
# Check if RUNNER exists, fallback if not
if ! command -v ${RUNNER} &> /dev/null; then
  echo "⚠️  Warning: Runner '${RUNNER}' not found. Falling back to './mvnw'."
  RUNNER="./mvnw"
fi
```

after `parse_args "$@"`

javahippie:
Good idea, just added the check.

kthoms (approved):
LGTM

PR #1321, opened 2025-09-18
Update pom.xml

YUVI43:
This update fixes Maven warnings caused by using a property reference in the <version> tag of the project POM.

🔴 Problem

Previously, the project version was defined as:

<version>${operaton.dbscheme.current.version}</version>

While this worked, Maven reported the following warning during builds:

[WARNING] 'version' contains an expression but should be a constant

Maven expects the <version> element to be a fixed, constant value. Using a property there causes warnings and may break plugins such as maven-release-plugin or CI/CD pipelines that depend on a stable project version.

🟢 Solution

Set <project><version> to a constant (1.0.0-SNAPSHOT) to comply with Maven requirements.

Moved the dynamic version into <properties> as:

<operaton.dbscheme.current.version>1.0.0-SNAPSHOT</operaton.dbscheme.current.version>

Added a missing property version.liquibase to centralize version management for Liquibase dependencies.

Left existing <dependencies>, <build>, and <profiles> untouched, ensuring no change in functionality.

⚡ Benefits

✅ Maven warnings removed ('version' contains an expression but should be a constant)

✅ Project version is now stable and compatible with release plugins

✅ Still supports dynamic version usage via the new property (operaton.dbscheme.current.version)

✅ No change in existing build outputs (SQL scripts, resources, dependencies, etc.)

✅ Codebase is cleaner and future-proof for automation pipelines

📖 Example

Before (warning triggered):

<version>${operaton.dbscheme.current.version}</version>

After (no warning, dynamic version still available as a property):

<version>1.0.0-SNAPSHOT</version>

<properties>
  <operaton.dbscheme.current.version>1.0.0-SNAPSHOT</operaton.dbscheme.current.version>
</properties>

This change is purely a build configuration improvement — it does not affect runtime behavior, dependencies, or functionality.

### DIFF
--- a/distro/sql-script/pom.xml
+++ b/distro/sql-script/pom.xml
@@ -1,21 +1,37 @@
-<?xml version='1.0' encoding='UTF-8'?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


YUVI43:
if you are satisfied with my work then add me to the conributor list thats only i want

YUVI43:
Add me to the contributor list

kthoms:
Closing.

Issue #1320, opened 2025-09-18
Upgrade Jackson to 2.20.0


kthoms:
**What needs to be done?**

The attached `dependencies.txt` is the result of 
```
./mvnw dependency:tree > dependencies.txt
```
[dependencies.txt](https://github.com/user-attachments/files/22399879/dependencies.txt)

In there you can see that the Jackson library is used in version `2.19.2`:
```
[INFO] +- org.operaton.commons:operaton-commons-utils:jar:1.0.0-rc-1-SNAPSHOT:compile
[INFO] |  \- org.junit.jupiter:junit-jupiter:jar:5.13.4:compile
[INFO] |     \- org.junit.jupiter:junit-jupiter-engine:jar:5.13.4:runtime
[INFO] |        \- org.junit.platform:junit-platform-engine:jar:1.13.4:runtime
[INFO] +- com.fasterxml.jackson.core:jackson-core:jar:2.19.2:compile
[INFO] +- com.fasterxml.jackson.core:jackson-databind:jar:2.19.2:compile
```

The version should actually be managed by the BOM `bom/internal-dependencies/pom.xml`, but this is somehow not picked up.
```
      <dependency>
        <groupId>com.fasterxml.jackson</groupId>
        <artifactId>jackson-bom</artifactId>
        <version>2.20.0</version>
        <scope>import</scope>
        <type>pom</type>
      </dependency>
```

Also `parent/pom.xml` declares this property:
```
<version.jackson>2.20.0</version.jackson>
```

Analyze why this happens and how to make the dependent modules use the right version. Why is 2.19.2 used?

In general, we try to use central dependency management via `bom/internal-dependencies/pom.xml` and avoid with this the need to use `<version>` tags on the dependencies. But since `version.jackson` is a property in the parent pom.xml, it is an option to add version tags with use of this property. Just saying, try to avoid it if possible, otherwise you are good to go.

**Acceptance Criteria**

- The dependency tree does not list any jackson dependency in version 2.19.2
- The build succeeds

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it.
Please start soon with the task when you signaled that you want to work on it and be responsive to questions or feedback.
We also do our best to review and merge your PR in a timely manner.

Are you a first-time contributor? We are happy to see you here! 
[...]

PR #1327, opened 2025-09-18
fix(deps): upgrade jackson to 2.20.0

khawaja-abdullah:
**What was the issue?**
Despite having the Jackson BOM properly configured with version 2.20.0 in `bom/internal-dependencies/pom.xml` and the corresponding property set in `parent/pom.xml`, the dependency tree still showed Jackson artifacts resolving to version 2.19.2 instead of the intended 2.20.0.
> _Issue Link_ https://github.com/operaton/operaton/issues/1320

**Solution Implemented**
The root cause was Maven's BOM import ordering mechanism and its "first wins" rule for dependency management. When multiple BOMs manage the same artifacts, Maven processes them sequentially and keeps the version declarations from the first BOM that declares each artifact. In this case, other BOMs (such as Spring Boot's BOM) were being imported before the Jackson BOM in the <dependencyManagement> section, so their older Jackson version declarations (2.19.2) took precedence over the Jackson BOM's 2.20.0 declarations. The solution was simply reordering the BOM imports so that the Jackson BOM was imported first, allowing it to establish version 2.20.0 for all Jackson artifacts before any other BOM could override them.

**How was the fix tested?**
1. `./mvnw dependency:tree | grep "jackson"` outputs intended version 2.20.0 of all jackson related dependencies [dependencies_jackson-fix.txt](https://github.com/user-attachments/files/22413944/dependencies_jackson-fix.txt)
2. `./mvnw clean install` confirmed that the build is passing on local with no test failures.
3. CI workflows - Look good to me, a couple of integration test workflows failing on exec but it seems to be unrelated as I can see the same failures on the last merged PR in main.

### DIFF
--- a/bom/internal-dependencies/pom.xml
+++ b/bom/internal-dependencies/pom.xml
@@ -49,6 +49,13 @@
         <scope>import</scope>
       </dependency>
       <dependency>
+        <groupId>com.fasterxml.jackson</groupId>
+        <artifactId>jackson-bom</artifactId>
+        <version>${version.jackson}</version>
+        <scope>import</scope>
+        <type>pom</type>
+      </dependency>
+      <dependency>
         <groupId>org.springframework.boot</groupId>
         <artifactId>spring-boot-dependencies</artifactId>
         <version>${version.spring-boot}</version>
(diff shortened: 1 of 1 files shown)


kthoms:
Don't care about the build failures. They are unrelated and I fix that now.

In general, your explanation looks valid and I have checked the dependency change. 

I plan to rebase you branch once the integration tests are working again. When the build succeeds, I plan to merge this.

Thank you!

kthoms (approved):

Issue #1324, opened 2025-09-18
Improve logging performance by using parameterized messages


kthoms:
**What needs to be done?**

This issue is part of our effort to modernize the codebase and improve readability.

SonarCloud identified a **Code Smell** where string concatenation is used in logging arguments. This can lead to a performance penalty, as the string is concatenated every time the logging method is called, regardless of whether the message is actually logged.

A simple fix is to use parameterized messages offered by logging frameworks like SLF4J, which only evaluate the arguments if the logging level is enabled.

You can find an example of this finding at the following URL:
[SonarQube](https://sonarcloud.io/project/issues?rules=java%3AS2629&issueStatuses=OPEN%2CCONFIRMED&id=operaton_operaton&open=AZbn-BJdOSOYNS1JdHNK)

---

### Steps to Resolve

1.  Navigate to the file indicated by the Sonar finding.
2.  Locate the logging statement that uses string concatenation (e.g., `logger.debug("message: " + variable);`).
3.  Refactor the code to use a parameterized message (e.g., `logger.debug("message: {}", variable);`).

**Example of the change:**

From this:
```java
LOG.debug("Processing file: " + fileName);
```
 To this:
```java
LOG.debug("Processing file: " + fileName);
```

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it.
Please start soon with the task when you signaled that you want to work on it and be responsive to questions or feedback.
We also do our best to review and merge your PR in a timely manner.

Are you a first-time contributor? We are happy to see you here! 
Just state in a comment of your PR "I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md).
Consider adding a star ⭐️ to the [repository](https://github.com/operaton/operaton) if you like it, we appreciate that.

Contributors will be added to the release notes of the next release.

joonseolee:
I will do it right away

kthoms:
Done

PR #1328, opened 2025-09-18
refactor(logging): Migrate to SLF4J with placeholder logging

joonseolee:
## Description

This PR migrates the logging system from `java.util.logging.Logger` to SLF4J and replaces string concatenation with placeholder-based logging for better performance and consistency.

## Example Changes

**Before:**
```java
LOGGER.info("Start 3 instances of " + processDefinition.getName() + ", version " + processDefinition.getVersion());
```

**After:**
```java
LOGGER.info("Start 3 instances of {}, version {}", processDefinition.getName(), processDefinition.getVersion());
```

## Related Issue

#1324

### DIFF
--- a/examples/invoice/src/main/java/org/operaton/bpm/example/invoice/InvoiceApplicationHelper.java
+++ b/examples/invoice/src/main/java/org/operaton/bpm/example/invoice/InvoiceApplicationHelper.java
@@ -195,7 +195,7 @@ public final class InvoiceApplicationHelper {
         processEngine.getIdentityService().clearAuthentication();
       }
     } else {
-      LOGGER.info(() -> "No new instances of {} version {} started, there are {} instances running",
+      LOGGER.info("No new instances of {} version {} started, there are {} instances running",
           processDefinition.getName(), processDefinition.getVersion(), numberOfRunningProcessInstances);
     }
   }


distro/wildfly/subsystem/src/main/java/org/operaton/bpm/container/impl/jboss/service/ProcessApplicationDeploymentService.java
```diff
@@ -262,7 +262,7 @@ protected void logDeploymentSummary(Collection<String> resourceNames, String dep
               .append(resourceName)
               .append("\n");
     }
-    LOGGER.log(Level.INFO, builder.toString());
+    LOGGER.info(builder.toString());
```
ungerts: Instead of manually building a multi-line StringBuilder, maybe we can use SLF4J's fluent API. May also lead to cleaner logs addKeyValue() enriches the log event with structured fields, making them machine-readable for JSON appenders and log aggregators, instead of burying the data in free-form text.. Example:

```java
private static String sanitize(String s) {
  if (s == null) return "(null)";
  return s.replace("\r", "\\r").replace("\n", "\\n");
}

void logDeploymentSummary(Collection<String> resourceNames,
                          String deploymentName,
                          String processApplicationName) {

  var names = (resourceNames == null) ? List.<String>of() : resourceNames;

  LOGGER.atInfo()
      .setMessage("Deployment '{}' of app '{}' with {} resources")
      .addArgument(sanitize(deploymentName))
      .addArgument(sanitize(processApplicationName))
      .addArgument(names.size())
      .addKeyValue("resourceCount", names.size())
      .addKeyValue("resources", names.stream().sorted().map(this::sanitize).toList())
      .log();
}
```

ungerts (changes requested):
@joonseolee Thanks a lot for your contribution! 🙏 I noticed one point around the use of StringBuilder in the logger. Could you take a look at that when you get a chance?

distro/wildfly/subsystem/src/main/java/org/operaton/bpm/container/impl/jboss/service/ProcessApplicationDeploymentService.java
joonseolee: @ungerts 

Actually, I wasn’t aware of the atInfo style API before, so thank you very much for the suggestion.
By the way, I moved the sanitize method to org.operaton.commons.utils.StringUtil. Would that be okay?
Also, I noticed the following two lines seem redundant, so I removed one and recommitted.
Could you please review when you have a chance?

```java
      .addArgument(names.size())
      .addKeyValue("resourceCount", names.size())
```

ungerts: @joonseolee I'll have a look

ungerts: Good call on moving to `StringUtil`. Regarding these two lines:

```java
.addArgument(names.size())
.addKeyValue("resourceCount", names.size())
```

They are not redundant. The first is for the message and the second is for structured metadata. I would keep both.

Regarding the placement of sanitize, maybe @kthoms can comment. I think he knows the right home for it.

kthoms: The method can be kept in StringUtil. However, I will add a test.

kthoms:
Sorry for reacting that lately. I just saw that there are discussions ongoing, but was not aware that my feedback was waited for. @ungerts please assign me as reviewer next time, I might not be aware enough when just tagged in a comment.

The change looks fine for me now. I have added a test for `sanitize`and merged `main`. If the build succeeds I will merge it.

kthoms:
The Wildfly integration tests are failing. I need to investigate the problem later.

kthoms:
Finally got a chance to look locally:
`qa/wildfly-runtime/target/server/wildfly-37.0.1.Final/standalone/log/server.log`

```
java.lang.ClassNotFoundException: org.slf4j.LoggerFactory from [Module "org.operaton.bpm.wildfly.operaton-wildfly-subsystem" 
```

kthoms:
Module dependencies `org.slf4, org.operaton.commons.operaton-commons-utils` are required by module `org.operaton.bpm.wildfly.operaton-wildfly-subsystem`

PR #1339, opened 2025-09-19
Update pom.xml

YUVI43:
I fixed the Maven warning caused by using a property in the <version> tag (${operaton.dbscheme.current.version}) inside distro/sql-script/pom.xml. It has been replaced with the explicit version 7.24.0.

Additionally, it is recommended to extend .devenv/scripts/maintenance/init-database-version.py so that future DB schema updates also update this POM automatically. This ensures Maven compliance, removes build warnings, and improves project stability.

Request
Kindly add me to the contributors list for this task.


kthoms:
Closing. This change shows that the requirements are misunderstood.

PR #1347, opened 2025-09-21
chore(engine): rename OperatonIntegration constants and references

hauptmedia:
This PR renames two constants to align them with the Operaton renaming and to remove the remaining CAMUNDA_* references:

•	CAMUNDA_EJB_SERVICE → OPERATON_EJB_SERVICE
•	CAMUNDA_BPM_RUN → OPERATON_BPM_RUN

These constants were previously only used in the telemetry code, so they are unlikely to be consumed externally. Renaming them ensures consistency with other changes where CAMUNDA_* identifiers were replaced by their OPERATON_* equivalents.


hauptmedia:
@cachescrubber do we need to add these to the migration script? There are also a few other constants left i would like to rectify before 1.0 release

cachescrubber:
No need for a new recipe. There is already the `org.operaton.rewrite.ChangeConstant` Recipe which covers basically all constants starting with CAMUNDA. Just amend the list in change-constant.yml. Same for methods and types.

hauptmedia:
I'll keep track of what i renamed in this working branch here and raise a PR when I'm finished => https://github.com/operaton/migrate-from-camunda-recipe/tree/chore/add-renamed-constants

Issue #1319, opened 2025-09-18
Resolve build warning: `'version' contains an expression but should be a constant`


kthoms:
**What needs to be done?**

A build warning is caused by using a property for the `<version>` tag in `distro/sql-script/pom.xml`:

```
./mvnw install -DskipTests -Dskip.frontend.build=true
```

At the beginning:
```
Warning: 281 [WARNING] 
Warning: 281 [WARNING] Some problems were encountered while building the effective model for org.operaton.bpm.distro:operaton-sql-scripts:jar:7.24.0
Warning: 281 [WARNING] 'version' contains an expression but should be a constant. @ org.operaton.bpm.distro:operaton-sql-scripts:${operaton.dbscheme.current.version}, /home/runner/work/operaton/operaton/distro/sql-script/pom.xml, line 12, column 12
Warning: 281 [WARNING] 
Warning: 281 [WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
Warning: 281 [WARNING] 
Warning: 281 [WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
Warning: 281 [WARNING] 
```

The property is defined in `database/pom.xml` with current value `7.24.0`. Replace the property with this version in `distro/sql-script/pom.xml`.

The property has to be changed when a new DB schema version is initiated. This happens usually with a release. The script `.devenv/scripts/maintenance/init-database-version.py` does this.

Extend the script that it updates `distro/sql-script/pom.xml` with the new specified DB schema version.

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it.
Please start soon with the task when you signaled that you want to work on it and be responsive to questions or feedback.
We also do our best to review and merge your PR in a timely manner.

Are you a first-time contributor? We are happy to see you here! 
Just state in a comment of your PR "I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md).
Consider adding a star ⭐️ to the [repository](https://github.com/operaton/operaton) if you like it, we appreciate that.

Contributors will be added to the release notes of the next release.

YUVI43:
Hello, I would like to contribute by working on this. Please assign it to me.

kthoms:
[...]

PR #1366, opened 2025-09-23
fix(build): Resolve build warning and update maintenance script

sogladev:
For this, I first tried with a regex, but there are other `<version>` elements, so I went with an approach similar to `get_property_from_pom()` to parse the XML instead.

- update pom.xml to current db version
- update maintenance script to set version

- Closes: #1319

### Additional notes
observed SyntaxWarning with `Python 3.13.7`

```sh
.../operaton/.devenv/scripts/maintenance/init-database-version.py:108: SyntaxWarning: invalid escape sequence '\)'
  m = re.search(r"values \('(\d+)', CURRENT_TIMESTAMP, '"+re.escape(current_version)+"'\);", line)
```

### DIFF
--- a/.devenv/scripts/maintenance/init-database-version.py
+++ b/.devenv/scripts/maintenance/init-database-version.py
@@ -212,17 +212,15 @@ def update_old_engine_pom(current_version, previous_version):
     with open(pom, "w", encoding="utf-8") as f:
         f.write(content)
 
-def update_sql_script_pom(current_version):
+def update_sql_script_pom(new_version):
     pom_path = "distro/sql-script/pom.xml"
     tree = ET.parse(pom_path)
     root = tree.getroot()
     ns = {'m': 'http://maven.apache.org/POM/4.0.0'}
-    parent = root.find('m:parent', ns)
-    if parent is not None:
-        version = parent.find('m:version', ns)
-        if version is not None:
-            version.text = current_version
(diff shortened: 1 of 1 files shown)


sogladev:

:+1:

.devenv/scripts/maintenance/init-database-version.py
```diff
@@ -272,6 +284,7 @@ def main():
     update_liquibase_changelog(cur_vwp, new_vwp, new_version)
     create_test_fixture(cur_vwpwd, new_vwpwd, current_version, new_version)
     update_old_engine_pom(current_version, previous_version)
+    update_sql_script_pom(current_version)
```
kthoms:
```suggestion
    update_sql_script_pom(new_version)
```

.devenv/scripts/maintenance/init-database-version.py
```diff
@@ -212,6 +212,18 @@ def update_old_engine_pom(current_version, previous_version):
     with open(pom, "w", encoding="utf-8") as f:
         f.write(content)
 
+def update_sql_script_pom(current_version):
```
kthoms: This changes the wrong version property. It must not change `/parent/version`, but `version`, which is the 2nd occurrence of the `version` tag

Bildschirmfoto 2025-09-23 um 23 19 15

PR #1367, opened 2025-09-23
refactor(tests): Replace try/catch with expected failure in test code

sogladev:
This is the only remaining pattern I could find that uses AssertJ

- Progresses #564


kthoms (approved):

kthoms:
There are MANY more usages of this pattern using `org.assertj.core.api.Assertions#fail(java.lang.String)` in the code.

Anyway, this change can be already accepted.

Issue #1353, opened 2025-09-22
Streamline usage of jetty-maven-plugin


kthoms:
**What is the issue?**

The root `pom.xml` declares 2 different `jetty-maven-plugin` plugins:

```
        <plugin>
          <groupId>org.mortbay.jetty</groupId>
          <artifactId>jetty-maven-plugin</artifactId>
          <version>8.1.16.v20140903</version>
        </plugin>
        <plugin>
          <groupId>org.eclipse.jetty</groupId>
          <artifactId>jetty-maven-plugin</artifactId>
          <version>${version.jetty}</version>
        </plugin>
```

The plugin with group `org.mortbay.jetty` is outdated. Instead, only `org.eclipse.jetty:jetty-maven-plugin` should be used. The property `${version.jetty}` is unbound in the root `pom.xml`. The current version of the plugin is [`11.0.26`](https://central.sonatype.com/artifact/org.eclipse.jetty/jetty-maven-plugin).

**What needs to be done?**

- `/pom.xml`
  - Remove the plugin `org.mortbay.jetty:jetty-maven-plugin`
  - Replace `${version.jetty}` with `11.0.26`
- Search for `<groupId>org.mortbay.jetty</groupId>`, replace by `<groupId>org.eclipse.jetty</groupId>`
- Check if the plugin configuration if it needs adjustment

**Verify**
```
.devenv/scripts/build/build-and-run-integration-tests.sh --testsuite=webapps
```

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it.
Please start soon with the task when you signaled that you want to work on it and be responsive to questions or feedback.
We also do our best to review and merge your PR in a timely manner.

Are you a first-time contributor? We are happy to see you here! 
Just state in a comment of your PR "I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md).
Consider adding a star ⭐️ to the [repository](https://github.com/operaton/operaton) if you like it, we appreciate that.

Contributors will be added to the release notes of the next release.

sgorski00:
I will send PR in a while.

kthoms:
Thank you @sgorski00 for resolving the task

PR #1368, opened 2025-09-23
fix: remove unnecessary jetty plugin

sgorski00:
fix: remove unnecessary jetty plugin

- remove unnecessary jetty plugin
- replaced use of mortbay maven-jetty-plugin to eclipse maven-jetty-plugin
- fixed version of eclipse maven-jetty-plugin

related to #1353


kthoms:
closes #1353

PR #1382, opened 2025-09-25
fix(queryAPI): correct TenantQueryProperty class visibility. Replaced `GROUP_ID` with `TENANT_ID` for consistency.

hauptmedia:
The keycloak plugin needs to be able to access to the `TenantQueryProperty` class, like it does with the `GroupQueryProperty` class. I also noticed that the TENANT_ID was erroneously named GROUP_ID in `TenantQueryProperty` which seems to be a copy & paste fail when it was initially created.


kthoms (approved):
LGTM

PR #1383, opened 2025-09-25
chore: replace "Camunda" prefixes with "Operaton" across codebase

hauptmedia:
There were a few places left where we missed some renames. Refactored constant names and properties for consistency.

### DIFF
--- a/distro/wildfly/subsystem/src/main/java/org/operaton/bpm/container/impl/jboss/extension/BpmPlatformExtension.java
+++ b/distro/wildfly/subsystem/src/main/java/org/operaton/bpm/container/impl/jboss/extension/BpmPlatformExtension.java
@@ -61,7 +61,7 @@ public class BpmPlatformExtension implements Extension {
 
   @Override
   public void initializeParsers(ExtensionParsingContext context) {
-    context.setSubsystemXmlMapping(ModelConstants.SUBSYSTEM_NAME, Namespace.CAMUNDA_BPM_PLATFORM_1_1.getUriString(), parser);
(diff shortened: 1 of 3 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/cmd/AbstractGetDeployedFormCmd.java
```diff
@@ -35,9 +35,9 @@
 public abstract class AbstractGetDeployedFormCmd implements Command<InputStream> {
 
   protected static final String EMBEDDED_KEY = "embedded:";
-  protected static final String CAMUNDA_FORMS_KEY = "operaton-forms:";
+  protected static final String OPERATON_FORMS_KEY = "operaton-forms:";
```
kthoms, line 38: Isn't this how the form models we get are used? We must not change it then. Maybe support both?

hauptmedia: Good catch here, the problem is the renaming of the content, not the variable name. We have to investigate here.

kthoms: @hauptmedia Could you have a look?

hauptmedia: Turns out this a real issue. I fixed it in https://github.com/operaton/operaton/pull/1435

This PR is on hold until the other PR is merged.

Issue #378, opened 2024-12-25
Enable SQL script and backward compatibility checks


kthoms:
**Is your feature request related to a problem? Please describe.**
The test `SqlScriptTest` checks that Liquibase updates are successfully applied. The test is working against the "old" scripts version. These scripts are copied from the dependency `org.operaton.bpm.distro:operaton-sql-scripts:${operaton.version.old}`.

The property is defined by
```
 <operaton.version.old>7.22.0</operaton.version.old>
```
Since this refers to the Camunda version Operaton is based on, the respective Operaton artifact in that version does not exist. Actually the test should work against the Camunda scripts.

However, the scripts artifact is not published to Maven Central.

**Describe the solution you'd like**
The tests should be working again, checking against the latest Camunda scripts in version 7.22.0/7.23.0/7.24.0.
Since these artifacts are not directly available, my proposal is to build and publish the scripts in the Operaton namespace.

It has not finally been decided how the versioning of Operaton database is, see [forum discussion](https://forum.operaton.org/t/deciding-on-the-database-versioning-for-operaton/105/9).
For the case that the versioning is continuing >= 7.23.0 we could simply deploy the 7.22.0 version of the scripts as `org.operaton.bpm.distro:operaton-sql-scripts:7.22.0`.
For the case that SQL scripts versioning aligns to Operaton, e.g. starting with 1.0.0, the 7.22 version should be somehow reflected in a different artifact id, e.g. `org.operaton.bpm.distro:camunda-sql-scripts:7.22.0`.

The resulting artifact has to be manually published to Maven Central.

In order to deploy Operaton SQL scripts in the future, the module has to be added to the `sonatype-oss-release` profile.

The execution of these checks has to be added to the Nightly Build, likely by adding the `check-sql` profile.

kthoms:
The Camunda version would contain a changelog `camunda-changelog.xml`, but Operaton expects that this file is named `operaton-changelog.xml`. This has to be patched in the built jar.

Another problem is the versioning: The sql-scripts build expects to get the current database version from the project version. But this does not match anymore: The DB version would be `7.23`, while Operaton is at `1.0`.

This will require that a property for the current database version is introduced.

kthoms:
The version 7.22 of the artifact has to be deployed to Maven Central to make the checks work against the "old" version.

javahippie:
@kthoms could you give a short update what is missing to close this issue?

PR #1424, opened 2025-10-06
feature(engine) Migration of serialized data with org.camunda namespace

### LINKED ISSUE #941: Migration of serialized data with org.camunda namespace
**What needs to be done? Please describe.**

Process definitions and instances (what else?) might contain references to Camunda classes, which are not valid anymore. In order to execute such processes on Operaton these entries have to be migrated.

**Additional Context**

Known usages: Several users have Javascript code that access `org.camunda.*` classes.

SevDan:
- add migration mechanism
- describe solution design in adr
- add configuration variant to run migration

close #941

### DIFF
--- a/engine/pom.xml
+++ b/engine/pom.xml
@@ -359,11 +359,13 @@
     <dependency>
       <groupId>org.graalvm.polyglot</groupId>
       <artifactId>polyglot</artifactId>
+      <scope>provided</scope>
     </dependency>
 
     <dependency>
       <groupId>org.graalvm.js</groupId>
       <artifactId>js-scriptengine</artifactId>
+      <scope>provided</scope>
     </dependency>
 
     <dependency>


docs/decisions/0001-add-runtime-resources-migration-configuration.md
```diff
@@ -0,0 +1,89 @@
...
+---
+
+# Camunda 7 to Operaton Runtime Resources Migration
+
+## Context and Problem Statement
+
+There is a problem with code migration for production runtimes if any platform code with FQN was used in resources like scripts or configurations. It is necessary to migrate it in a database
+
+<!-- This is an optional element. Feel free to remove. -->
+## Decision Drivers
```
ungerts: Please format decision drivers as a bullet list for clarity. Example:

- Must enable smooth migration from Camunda to Operaton without data loss  
- Must remain DB-agnostic  
- Must allow parallel cluster operation (Camunda + Operaton)  
- Should avoid manual SQL scripts  
- Should not break existing deployments

docs/decisions/0001-add-runtime-resources-migration-configuration.md
```diff
@@ -0,0 +1,89 @@
+---
+# These are optional metadata elements. Feel free to remove any of them.
+status: "proposed"
+date: 2025-10-02
+---
+
+# Camunda 7 to Operaton Runtime Resources Migration
+
+## Context and Problem Statement
```
ungerts: Please make the scope explicit — specify what artifacts are affected:
- In scope: BPMN, DMN resources (currently rewritten)
- Out of scope: scripts, forms, or serialized Java classes inside history tables

ungerts:
Hi @SevDan ,  

thanks for taking care of the issue. Just a few of my thoughts:  

I'm not sure this approach will work reliably in a clustered environment. There’s no global lock, guard, or persistent marker to ensure that the migration runs only once. If multiple nodes start up in this mode, they could all execute the migration in parallel, leading to duplicate deployments or inconsistent data.  
See `CamundaToOperatonLiquibaseMigration.java` lines 164–208 — migration runs directly without any synchronization or `ACT_GE_PROPERTY` marker.  

Even on a single node, the process isn’t idempotent — without a durable “already migrated” flag, re-running it would create duplicate entries.  
See the same file lines 245–282, where new deployments are inserted unconditionally without checking for existing Operaton versions.  

The code also swallows some exceptions and doesn’t define clear transactional boundaries or rollback behavior. If a failure occurs mid-way, the database could end up in a partially migrated state. Wrapping the logic in explicit transactions, handling errors properly, and recording a completion marker would make the process much safer and easier to operate.  
See `CamundaToOperatonLiquibaseMigration.java` line 180 — comment `// swallow the exception !`; no transaction demarcation around inserts.  

The current logic scans the entire `ACT_GE_BYTEARRAY` table and processes everything in memory. For large installations, that could cause high memory or transaction load. Pagination or batching would make it safer for production-scale migrations.  
See `CamundaToOperatonLiquibaseMigration.java` lines 120–150, where all entries are selected and iterated in one loop without paging or streaming.  

Schema identifiers like `PUBLIC` look H2-specific and may not be portable across different databases. It seems the existing configuration options (for example `camunda.bpm.database.schema-update`, `.type`, `.table-prefix`, `.schema-name`) are not being considered here. These parameters already define how schema and table naming should behave depending on the environment and database type, and the migration logic should respect those instead of hardcoding schema names or prefixes.  
See SQL references to `PUBLIC.ACT_RE_DEPLOYMENT` and `PUBLIC.ACT_GE_BYTEARRAY` in `CamundaToOperatonLiquibaseMigration.java` lines 101–110 — hardcoded schema name.  

Finally, I’m not sure this should be implemented as a permanent engine mode. Unlike schema updates, which are designed to run repeatedly on startup, a migration should only ever be executed once under controlled conditions. It might make more sense as a standalone migration tool or explicit command, rather than something that can run automatically on every engine startup.  
See `SchemaOperationsProcessEngineBuild.java` lines 95–115, where `"migrate-from-camunda"` is added as another `databaseSchemaUpdate` mode and treated similarly to schema updates.  

Thanks again for your work on this.

engine/src/main/java/org/operaton/camunda/migration/CamundaToOperatonLiquibaseMigration.java
```diff
@@ -0,0 +1,280 @@
...
+  private byte[] transformBytesForDmn(byte[] bytes) {
+    final InputStream inputStream = new ByteArrayInputStream(bytes);
+
+    final DmnModelInstance dmnModelInstance = Dmn.readModelFromStream(inputStream);
+    final DmnModelInstance instanceClone = dmnModelInstance.clone();
+
+    final ModelElementInstance documentElement = instanceClone.getDocumentElement();
+
+    final String replacedText = documentElement.getTextContent()
+        .replaceAll("org\\.camunda\\.bpm", "org.operaton.bpm")
```
kthoms: Just use replace().

engine/src/main/java/org/operaton/camunda/migration/CamundaToOperatonLiquibaseMigration.java
```diff
@@ -0,0 +1,280 @@
...
+import org.operaton.bpm.model.dmn.Dmn;
+import org.operaton.bpm.model.dmn.DmnModelInstance;
+import org.operaton.bpm.model.xml.instance.ModelElementInstance;
+
+/**
+ * <p>Task for resources in database runtime migration.
+ * Do nothing if current database is not on latest version of Camunda 7.
+ * Creates new deployments and saves artifacts there.</p>
+ */
+public class CamundaToOperatonLiquibaseMigration {
```
kthoms: In the end, there should be extensive tests for this

kthoms (commented):
Great to see this feature evolving. I'm not sure if the chosen approach is the best one, but I'm lacking a better idea. It is pragmatic. It is _not_ a _schema_ update operation, this is a bit misleading.
I cannot judge really that it is performing all necessary updates. This I have to leave to others to review.

SevDan:
Hello, @ungerts @kthoms  thanks for a great review, I will fix problems, refactor the code and maybe will suggest some other variants of configuration-based solution (i.e just add one more configuration parameter)

I think also user experience thoughts are highly required for good solution design in this case

ungerts:
Hi @SevDan, @kthoms,

I just wanted to raise a broader question about the overall approach.  
There are essentially two ways to solve this problem. One is migration, where we physically rewrite the data. The other is compatibility, where the engine keeps the data as it is but applies conventions during class loading and scripting so that Camunda namespaces are interpreted correctly at runtime.

As seen above, migration has its trade-offs — especially with very large tables. I’m mostly concerned about how long a full migration would take and how much memory it would consume if all entries are processed in one go.

That’s why I’d like to suggest evaluating a runtime compatibility layer as an alternative. The idea is that the engine transparently resolves `org.camunda.*` → `org.operaton.*` references during class loading and scripting. This would allow existing data to run unchanged, without any database writes or duplication. It could complement the current migration option or, for large environments, even replace it as a safer, zero-downtime path.

To clarify what I mean by compatibility: the engine’s reflection and scripting utilities could be extended so that, whenever a class name starting with `org.camunda.*` is encountered, it automatically retries loading the corresponding `org.operaton.*` class. Script engines could apply the same rule for import statements. This would make legacy process data and serialized variables executable as-is, controlled by a configuration flag, and could coexist with the migration mode for users who prefer a non-intrusive upgrade.

SevDan:
@ungerts I think it's good idea and may be another option of compatibility mode, available by parameter value like 'compatibility' (in contrast of 'migrate-deployments' for artifacts duplication). I don't see any significant problems with that approach now (also it can be combined with data migration, executed by background batch processing task)

kthoms:
The "compatibility" is the approach AFAIK cibseven is taking. While this sounds like a good option on first sight, I have serious doubts about it. Right now chances are really high that this magic works, since it is mostly just package renamings.

An obvious positive aspect is that the data can stay untouched for now.

However, I am pro data migration. The data is not valid from Operaton's perspective, and we would just work around data garbage. IMO it is better to have clean data. Data migrations should of course be avoided whenever possible, but this scenario is one I would consider a migration to be valid.

We are speaking only for installations that use serialized data, which is an anti-pattern anyway. Most Camunda usages do not make use of serialized Java objects. If they do and consider Operaton and we offer a migration tool, I think this is good enough. There might be a downtime to perform the migration, this would have to be communicated. Still, I think this valid.

The tool could also do it stepwise. Most critical are active process instances. Maybe active processes can be migrated in a first step with a mandatory downtime. How many _active_ processes exist for one installation with using serialized Java objects? I would bet this is overseeable.

But the most mass will be in historic data. This could be done asnych without downtime, maybe in batches.

Just thoughts.

ungerts:
Hi @kthoms,

I agree that having clean data is the right long-term goal, and a proper migration is a valid approach in principle. However, I think the operational aspect needs to be clarified before we commit to that path.

If downtime is acceptable, then a one-time, controlled migration during startup could work. If we only allow a short restart window, just enough for the engine to stop and start again, that time might not be sufficient to migrate all data, especially in large installations. The `ACT_GE_BYTEARRAY` table can become very large in production; there are reports of it reaching around 200 GB ([Camunda Forum](https://forum.camunda.io/t/delete-old-records-from-act-ge-bytearray/43137)). Running a migration against such a table would put significant load on the database in terms of I/O, CPU, and transactions, and could temporarily double its size because of record duplication during deployment recreation.

From that perspective, a runtime compatibility option could serve as a practical complement rather than a replacement. It would allow systems to start and handle existing data immediately without downtime, while a proper migration could still be performed later as a cleanup step to ensure data consistency and correctness.

We should probably also clarify the requirements on how fast the migration is expected to complete. If it needs to finish within a short maintenance window, that would strongly influence the technical approach we choose.

SevDan:
Probably the data management here is a very complicated question that depends on much specific requirements such as history store strategy (in-db or out), archiving policies, time-traveling capability (for audit reports maybe) etc. Probably, it may be a good stategy to provide two different ways for native runtime migration that will cover most of the cases but not all of them. So, the first way will be suitable for small to medium databases, migrates data in duplicated deployments by built-in task and the second one will be suitable for large ones and provide compatible mode to use legacy data

I have assumption that installations with small to medium sized databases will tolerate maintanance downtime (that also will be not so long) and large ones will not, but large installations probably could be possible to be migrated some custom way that fulfill complicated requirements. Some recommendations may be enought for that type of users

upd: I will also add suggested compatible mode as an option to adr and write down pros and cons from current discussion there

hauptmedia:
Just a quick thought from me: If we decide to add the runtime rewriting, let’s please backport the already tested solution from CibSeven instead of creating yet another implementation.

We might need to extend it slightly, since we not only renamed namespaces but also class names.
The good news is that we already maintain a list of these mappings in our openrewrite recipe, which could potentially be reused to dynamically rewrite class references at runtime.

SevDan:
Quick status update: got some problems with data migration tests (tried default arquillian way but seems a bit tricky to have database filled with fixtures data). I will share some options and details later

Also the locks probably should be implemented as simple property check instead of real locking. It will be consistent with current schema upgrade 'locking' (also just single queries with simple check without real locking). We can describe migration strategy with dedicated single migration instance for users with clustered deployments (seen same strategy already used by camunda users for schema upgrade in clusters)

My plan is to describe more solution details with all tradeoffs in next commit

SevDan:
Update: Got a problem when tried to merge actual main branch and new tests failed with exception. Will investigate a little and describe problem in corresponding issue if the scripting is a problem

SevDan:
Backported cibseven solution as code saves code authors. Changed test script to be compatible with graal polyglot 25.0.0

SevDan:
Seems data migration testing approach should be discussed or moved into dedicated issue because chosen one (data migration tests as part of instance migration suite) depends on instance migration suite, which is disabled yet (https://github.com/operaton/operaton/issues/378).

SevDan:
Decided to observe some points:
* Transactions and locks
Didn't used any of it due to consistency with camunda 7 migration style. It is necessary to run migration self-isolated way (i.e run single migration instance and after shutdown it and run actual replicated cluster)
* cibseven backport
Commit picking failed, so backported just as source code copy, modified to be compatible with current graal polyglot
* instance migration still unavailable
Instance migration tests still does not works, but instance data migration test depends on it

PR #1483, opened 2025-10-17
chore(test): Update db-rolling-tests to JUnit5.

ArneDeutsch:
relates to #27


kthoms:
@ArneDeutsch This does not seem to work

PR #1501, opened 2025-10-27
New License book

kthoms:
Adds a Python script that generates the Operaton license book as markdown file. The license book uses the SBOMs produced by `.devenv/scripts/build/build-sbom.sh`.

The all used 3rd party licenses are stored as text files.

Extended the integration build & release workflow to include the license book generation and attach it as release artifact in `jreleaser.yml`.

Improved the "Dry Run" mode of the release workflow by this chance. The workflow does not do ay commits or uploads when Dry-Run is set- This made also the input field "release_branch" obsolete.

Removed the old `license-book.txt` and replaced references by `license-book.md`.

The result can be viewed at [license-book.md](https://github.com/operaton/operaton/blob/licensebook/distro/license-book/src/main/resources/license-book.md)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -305,7 +305,6 @@
             </additionalJOptions>
             <doctitle>Operaton Javadocs ${operaton.version.majorVersion}.${operaton.version.minorVersion} API</doctitle>
             <windowtitle>Operaton Javadocs ${operaton.version.majorVersion}.${operaton.version.minorVersion} API</windowtitle>
-            <excludePackageNames>*.impl:*.impl.*</excludePackageNames>
           </configuration>
         </plugin>
         <plugin>


distro/license-book/src/main/mustache/license-book.tpl
```diff
@@ -0,0 +1,14 @@
+# License Book for Operaton {{version}}
+
+{{>toc}}
+
+{{>preamble}}
+
+{{>license-summary}}
+
+{{> license}}
```
cachescrubber: Do we really need to include each license text here? if yes, could we place it at the very end of this file?

kthoms: I've done it the same way as the current license book. I'm not sure about this actually. This is a discussion point.

The original file is a large text file and not really readable, also because of this. I thought having it as markdown would at least make it a bit better consumable.

Actually, I would expect that each library's license file is included, but this is not usable at all. In the end, most are just containing the license again and again, and having them once reduces the friction. But it is not completely accurate.

I understood the license book to be a single file containing all dependencies with their licenses. If it is usable could be doubtful. But having this for legal reasons can be useful, I think.

But thanks for bringing up the discussion! This is what I am asking for. Moving the licenses to the end is of course an option.

cachescrubber: thanks for the explanation. I think we should move the license text it to the end - this makes it even more easier / comfortable to consume this file (even when links are not working).

kthoms: Moved the license text to the end

kthoms:
@javahippie I need your review here. The script will be cleaned up later. Please focus on the produced license book. What must be done to ship this new version with the release instead of the old one?

cachescrubber:
LGTM - Even the anchors are working now.

javahippie:
I have compared the output to a local trivy run (`trivy fs --scanners=license --severity UNKNOWN,HIGH,CRITICAL .`) on my machine with the following deviations:

`mousetrap` has the license "Apache 2.0 WITH LLVM-exception", the MySQL dependencies are "The GNU General Public License, v2 with Universal FOSS Exception, v1.0". I think both are missing from the packaged license texts

Are these so deeply transitive, that we don't need to list them?

I attached trivys output here:

[licenses.log](https://github.com/user-attachments/files/23183850/licenses.log)

kthoms:
@javahippie Thank you for the pointers. I'll check them later and come back here.

kthoms:
@javahippie `moustrap` should be listed, it is used also in the source. Means to me that `license-checker` is not accurate enough. I saw that `moustrap` is listed in the newly produced SBOMs, and there is license information in them. I'll use that as a source now.

kthoms:
Regarding MySQL and other drivers: They are used only on test scope. Since they are not shipped, I think it is correct that they are not listed.

kthoms:
@javahippie I'm using now the SBOMs that CycloneDX is producing. This makes the implementation easier as the sources for Maven & NPM dependencies have the same format.

However, I am still missing `mousetrap` and other NPM dependencies. The CycloneDX generator does not list them in the [SBOM](https://github.com/operaton/operaton/releases/download/early-access/operaton-webapps.cyclonedx-json.sbom).

I'm not sure how much time I should invest for 1.0 to fix these gaps. Maybe we can just document this as known issue: Open an issue and make a note with link to it in the license book?

javahippie:
I'm not sure how much effort it would be to test this: trivy can generate reports in CycloneDX format: `trivy fs --format cyclonedx --output result.json .` This contains all npm dependencies and their licenses - but it also contains libraries which are only listed in test scope

I attached the output from my local run here: 
[result.json](https://github.com/user-attachments/files/23212380/result.json)

kthoms:
@javahippie Thanks for the pointer with trivy. I have integrated it into the SBOM generation script. But I have used the Docker image one to avoid the need for an installation.

Have a look again, `mousetrap` and bpmn.io licensed libraries are now listed.

IMO this should be good enough for version 1.0.

kthoms:
@javahippie @cachescrubber From my view this should be good now. I have included the license book generation into the release process and the file is attached as release artifact. The old file is removed, references are updated to the new file.

The book contains now references for each library to Maven Central and npmjs.com.

javahippie (approved):
I agree, this looks good to me!

Issue #1526, opened 2025-10-31
[ci] Add a workflow to add the "Contribution" section when label "good first issue" is assigned


kthoms:
**What needs to be done?**

The issue template `.github/ISSUE_TEMPLATE/good_first_issue.md` contains the "Contribution" section that is seen below. This template should be removed. Instead, a workflow action should check if the label "good first issue" has been added. On creation of the label, the workflow should add the contribution template. Create a template `.github/templates/comment_good_first_issue.md`. When the label is created, read the template and add it as content.

Note that new workflows can only be tested when added to `main`. Try to keep your work clean for testing. Once tested, produce a branch and open a PR. Attach a link to an issue that has been commented by the workflow.

**Contribution**

If you want to work on this task, just state it in a comment. You don't have to wait until the task is assigned to you.
First come, first serve! Avoid starting work on this task when someone else has already claimed it.
Please start soon with the task when you signaled that you want to work on it and be responsive to questions or feedback.
We also do our best to review and merge your PR in a timely manner.

Are you a first-time contributor? We are happy to see you here! 
Just state in a comment of your PR "I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) 
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md).
Consider adding a star ⭐️ to the [repository](https://github.com/operaton/operaton) if you like it, we appreciate that.

Contributors will be added to the release notes of the next release.

Jaehyeon-Han:
Hi @kthoms, I'm new here and I'd like to contribute to the issue.

kthoms:
Hey @Jaehyeon-Han ! Thank you for providing the PR. Could you take care to address the feedback?

Jaehyeon-Han:
@kthoms I left a question regarding your review for clarification. Could you answer the comment in the review? I will work on it right away as soon as what I assumed is clarified.

kthoms:
Thank you @Jaehyeon-Han for resolving it.

kthoms:
Works like a charm
https://github.com/operaton/operaton/issues/1554

Jaehyeon-Han:
Good to know! One thing to note is that if the label is removed and then added again, there will be two identical comments, which is fortunately quite unlikely.

PR #1538, opened 2025-11-04
[#1526] Add a workflow to add the "Contribution" section when label "good first issue" is assigned

Jaehyeon-Han:
This is the PR for [issue #1526](https://github.com/operaton/operaton/issues/1526).

I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0)
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md).

## Changes
- Removed `.github/ISSUE_TEMPLATE/good_first_issue.md`.
- Created `.github/templates/comment_good_first_issue.md` with only "Contribution" section of the removed template.
- Created `good-first-issue.yml` to add contribution template comment to the issue when labeled "good first issue".

## Links
- [Issue that has been commented by the workflow](https://github.com/Jaehyeon-Han/operaton/issues/7)
- [The workflow result that added the comment section](https://github.com/Jaehyeon-Han/operaton/actions/runs/19192151221)

---

Resolves: #1526

### DIFF
--- a/.github/workflows/good-first-issue.yml
+++ b/.github/workflows/good-first-issue.yml
@@ -12,7 +12,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v5
         
-      - name: Add template to the issue
+      - name: Add template comment to the issue
         uses: actions/github-script@v8
         with:
           script: |
@@ -30,11 +30,8 @@ jobs:
               issue_number: context.issue.number
             });
             
-            const issueBody = issue.body ?? '';
-            
-            await github.rest.issues.update({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              issue_number: context.issue.number,
-              body: issueBody + '\n\n' + template
+            await github.rest.issues.createComment({
+              ...context.repo,
+              issue_number: issue.number,
+              body: template
             });


.github/workflows/good-first-issue.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  issues:
+    types: [labeled]
+  
+jobs:
+  add-good-first-issue-comment:
+    runs-on: ubuntu-latest
+    if: github.event.label.name == 'good first issue'
+    steps:
+      - name: Checkout repository
```
kthoms, line 12: why is the repository checked out? seems not necessary here.

.github/workflows/good-first-issue.yml
```diff
@@ -0,0 +1,40 @@
...
+              issue_number: context.issue.number
+            });
+            
+            const issueBody = issue.body ?? '';
+            
+            await github.rest.issues.update({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
+              body: issueBody + '\n\n' + template
```
kthoms: This here _modifies_ the issue body. The workflow should _add_ a comment to the workflow. My bad, this was not clear from the description. I would like that the workflow adds a comment to the issue.

kthoms (changes requested):
In general this change is going the right direction. Please update as requested.

.github/workflows/good-first-issue.yml
Jaehyeon-Han: I used it to access the template file on the runner. Maybe checking out only that file could be better?

.github/workflows/good-first-issue.yml
Jaehyeon-Han: I see. I will change the workflow so that it leaves a comment instead of adding it to the issue content, test them, and modify the links in the PR accordingly.

.github/workflows/good-first-issue.yml
kthoms: Does not matter too much. I was not aware that a file from the repo is needed. I have overseen it. Leave it as is.

.github/workflows/good-first-issue.yml
kthoms: Great! Looking forward to the change.

Jaehyeon-Han: Changed the workflow, and updated the links in the PR as well. Let me know if you need other changes!

kthoms (approved):

Issue #1528, opened 2025-11-01
Produce distribution specific SBOMs


kthoms:
**Context**

The script `.devenv/scripts/build/build-sbom.sh` produces 2 SBOMs in CycloneDX format to the directory `target/sbom`:
- `operaton-webapps.cyclonedx-json.sbom` for the frontend apps
- `operaton-modules.cyclonedx-json.sbom` for the backend

The backend SBOM contains a listing for _all_ dependencies that the backend has. In order to do so, it activates several Maven profiles, since some dependencies are dependent on the activated profiles.

The SBOMs should be more targeted to what users actually download or integrate. This depends on the distribution.

**What needs to be done?**

- Create distribution specific SBOMs
   - Profile `distro-run` => `target/sbom/operaton-modules-run.cyclonedx-json.sbom`
   - Profile `distro-tomcat` => `target/sbom/operaton-modules-tomcat.cyclonedx-json.sbom`
   - Profile `distro-wildfly` => `target/sbom/operaton-modules-wildfly.cyclonedx-json.sbom`
   - Profile `distro-starter` => `target/sbom/operaton-modules-starter.cyclonedx-json.sbom`
   - Profile `distro-webjar` => `target/sbom/operaton-modules-webjar.cyclonedx-json.sbom`
   - Profile `distro-serverless` => `target/sbom/operaton-modules-serverless.cyclonedx-json.sbom`
- Create a script that merges the frontend with the backend SBOM
  - For example, merge `target/sbom/operaton-modules-run.cyclonedx-json.sbom` and `target/sbom/operaton-webapps.cyclonedx-json.sbom` => `target/sbom/operaton-run-sbom.cdx`
- Copy the final SBOMs to the target directories where the packaging of distros is happening, e.g. `distro/run/assembly/resources`
  Question: Where for Spring Boot Starter, Serverless?
- Add a workflow: When a PR gets merged that has a "dependencies" label, regenerate the SBOMs and check them in

**Additional context**

The license book shall be also produced distribution specific. This task is a preparation for this.

kthoms:
@javahippie reported that he observed issues executing the build and proposed to activate the license book generator by a profile. The license book is not mandatory to build the distribution archives, but the assembly descriptor reference them. The final distributions should include them.

Therefore I will modify the assembly.xml files that a build does not fail when the license book is missing and make sure the license book is only generated on github ci by default.

PR #1561, opened 2025-11-12
feat(build): Define profile 'github-ci' and produce license book when it is activated

kthoms:
The build should not assume that Python is installed to generate the license book. In regular builds the license book is just a placeholder file. When the profile 'github-ci' is activated, the license book is generated.

### DIFF
--- a/distro/run/assembly/pom.xml
+++ b/distro/run/assembly/pom.xml
@@ -111,42 +111,26 @@
       <artifactId>groovy-datetime</artifactId>
     </dependency>
   </dependencies>
+
   <build>
     <plugins>
       <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>exec-maven-plugin</artifactId>
+        <artifactId>maven-antrun-plugin</artifactId>
         <executions>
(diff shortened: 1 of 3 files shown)


distro/run/assembly/pom.xml
```diff
@@ -171,11 +149,64 @@
...
         </executions>
       </plugin>
     </plugins>
   </build>
-</project>
\ No newline at end of file
+
+  <profiles>
+    <profile>
+      <id>github-ci</id>
```
javahippie, line 167: Maybe a name like `licensebook-release` or similar would show the intent of this profile a little better?

kthoms: I chose this profile name with the intention that it could be used for other steps that should only run on CI by default and with the aim to not have profiles for each corner case.
Anyway, I'm open here.

javahippie: Makes sense to me!

javahippie (approved):
lgtm, thanks!

kthoms:
closes #1528

Issue #360, opened 2024-12-20
Publish distribution through Chocolatey


kthoms:
**Is your feature request related to a problem? Please describe.**
As a **Windows** user I would like to be able to install Operaton via Chocolatey.

**Describe the solution you'd like**
As part of the release, JReleaser should publish to Chocolatey. At least for "Operaton Run", but if possible for the other distributions as well.

This will require the following
- Distributions are (additionally) released as zip. Currently they are .tar.gz, but this is [not supported by JReleaser](https://github.com/jreleaser/jreleaser/issues/1594) (yet)
- A new repository chocolatey-bucket
- adding distributions to `jreleaser.yml`

**Describe alternatives you've considered**
This could also be manually added and added to the release workflow.

**Additional context**

- [JReleaser Install Doc](https://jreleaser.org/guide/latest/install.html)
- [JReleaser homebrew-tap repository](https://github.com/jreleaser/chocolatey-bucket)

hauptmedia:
Isn't winget the new de-facto standard package manager for Windows?

kthoms:
I'm not a Windows user myself, but heard of Chocolatey at some customers. 

winget is also supported by JReleaser. In the end I would like to offer different distributions, one after the other and by separate feature requests.

First, I would like to have _any_ distribution mechanism for each OS, then support some more. The process is then automated with JReleaser.

kthoms:
I withdraw this feature request until there are concrete user needs for distribution(s) installable via Chocolatey.

PR #1566, opened 2025-11-14
fix(stale): relax stale rules for milestone and deferred issues (#1510)

### LINKED ISSUE #1510: [ci] Improve Stale Check: Ignore issues assigned to milestone
**What needs to be done?**

The workflow `stale.yml` is too aggressive. Issues that have been assigned to a milestone should be ignored for now, or if possible, the period for those made longer.

**Additional context**

For example issue https://github.com/operaton/operaton/issues/360 is known that it is to be done later, there is no need to mark those stale.

aashvgit:
- Added `exempt-issue-milestones: true` so milestone issues (e.g., #360) are not marked stale
- Increased inactivity threshold to 90 days
- Added label exemptions: backlog, later, do not stale

Fixes #1510

### DIFF
--- a/database/pom.xml
+++ b/database/pom.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="https://maven.apache.org/POM/4.0.0" xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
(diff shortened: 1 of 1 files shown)


database/pom.xml
kthoms, line 1: revert this file change

.github/workflows/stale.yml
```diff
@@ -4,22 +4,37 @@ name: Mark stale issues & PRs
...
+          # Labels
+          stale-issue-label: 'status:stale'
+          stale-pr-label: 'status:stale'
+
+          # Timing (extend if your project has slower cycles)
+          days-before-stale: 90
+          days-before-close: 14
+
+          # Exemptions
+          exempt-issue-milestones: true
```
kthoms: The exemptions seem invalid, looks AI halluszinated. exempt-issue-milestones should name milestones, it is not a boolean value
exempt-issue-labels: these labels do not exist

It seems that 
[exempt-all-milestones](https://github.com/actions/stale?tab=readme-ov-file#exempt-all-milestones) should be the right flag to set

kthoms: changed

kthoms:
Thank you,  changed the code according to my findings

PR #1571, opened 2025-11-14
chore: Removed deprecated SecurityManager API

### LINKED ISSUE #1570: Remove usage of the deprecated SecurityManager classes
**What needs to be done?**

We are currently seeing build warnings because we still have references to the deprecated Security Manager in some places. There is no replacement for this and the recommendation is to remove the calls to the Security Manager. In recent JVM versions these calls were replaced with noop-methods anyway.

javahippie:
Fixes #1570


kthoms:
@javahippie Something missing?

javahippie:
I did not see that you reviewed it, my bad.


I decided on my own to remove this API, but as commented in the issue related to this PR, I don't really see any other way forward. There will be no replacement, so it's just removed now. If that's fine with you, I will merge this now

PR #1582, opened 2025-11-16
feat(health): add runtime‑agnostic Health SPI and Run /health endpoint

### LINKED ISSUE #1245: Add /health endpoint for monitoring and load balancer integration
### Problem
Currently, Operaton returns a 404 error when accessing `/health`, which is commonly expected for:
- Load balancer health checks
- Monitoring systems
- Container orchestration platforms
- Uptime monitoring

### Current Behavior
- GET `/health` returns 404 with "Whitelabel Error Page"
- Error message: "This application has no explicit mapping for /error"

### Requested Feature
Add a simple health check endpoint that returns:
- HTTP 200 when application is healthy
- Basic JSON response like `{"status": "UP", "timestamp": "..."}`
- Optional: Include basic system info (database connectivity, version, etc.)

### Use Case
We're running Operaton behind a reverse proxy and need a reliable endpoint for health monitoring in our infrastructure.

### Proposed Implementation
Standard Spring Boot Actuator health endpoint or custom controller at `/health`

Tomnm1:
Introduce a lightweight, runtime-agnostic health surface and integrate it across modules so health checks work consistently in different deployment models.

- add HealthService (SPI) and HealthResult (value object)
- expose GET /health returning JSON
- HealthController delegates to HealthService and is outside REST auth
- new OperatonBpmHealthServiceConfiguration registers DefaultHealthService using optional DataSource, JobExecutor, FrontendHealthContributor
- implement SpringWebappFrontendHealthContributor to report webapp presence and configured application path
- no hard dependency on Actuator in Run; Actuator remains optional

related to #1245, closes #1245

### DIFF
--- a/distro/run/core/src/main/java/org/operaton/bpm/run/health/HealthController.java
+++ b/distro/run/core/src/main/java/org/operaton/bpm/run/health/HealthController.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2025 the Operaton contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.operaton.bpm.run.health;
+
+import org.operaton.bpm.engine.health.HealthResult;
+import org.operaton.bpm.engine.health.HealthService;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.MediaType;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+/**
+ * Health endpoint for Operaton Run.
+ * Provides a lightweight /health JSON suitable for load balancers and uptime checks.
+ * Note: This endpoint is specific to the Operaton Run distribution. When deploying
+ * Operaton on application servers like Tomcat or WildFly, those runtimes typically
+ * provide their own health endpoints (e.g., via vendor-specific tooling or
+ * Spring Boot actuator in custom apps), so this endpoint is not exposed there.
+ *
+ * @author <a href="mailto:tomnm77@gmail.com">Tomasz Korcz</a>
+ */
+@RestController
+@ConditionalOnProperty(prefix = "management.health.operaton", name = "enabled", matchIfMissing = true)
+public class HealthController {
+
+  private final HealthService healthService;
+
+  public HealthController(HealthService healthService) {
+    this.healthService = healthService;
+  }
+
+  @GetMapping(path = "/health", produces = MediaType.APPLICATION_JSON_VALUE)
+  public ResponseEntity<Map<String, Object>> health() {
+    Map<String, Object> body = new LinkedHashMap<>();
+
+    HealthResult result = healthService.check();
+    body.put("status", result.status());
+    body.put("timestamp", result.timestamp());
+    if (result.version() != null) {
+      body.put("version", result.version());
+    }
+    body.put("details", result.details());
+    return ResponseEntity.ok()
+      .header(HttpHeaders.CACHE_CONTROL, "no-store, no-cache, must-revalidate, max-age=0")
+      .body(body);
+  }
+}
--- a/distro/run/core/src/test/java/org/operaton/bpm/run/test/health/HealthEndpointTest.java
+++ b/distro/run/core/src/test/java/org/operaton/bpm/run/test/health/HealthEndpointTest.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2025 the Operaton contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.operaton.bpm.run.test.health;
+
+import org.junit.jupiter.api.Test;
+import org.operaton.bpm.run.test.AbstractRestTest;
+import org.springframework.http.ResponseEntity;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class HealthEndpointTest extends AbstractRestTest {
+
+  @Test
+  void shouldReturnHealthStatus() {
+    ResponseEntity<JsonNode> entity = testRestTemplate.getForEntity("/health", JsonNode.class);
+    assertThat(entity.getStatusCode().value()).isEqualTo(200);
+    JsonNode body = entity.getBody();
+    assertThat(body).isNotNull();
+    assertThat(body.has("status")).isTrue();
+    assertThat(body.get("status").isTextual()).isTrue();
+    assertThat(body.has("timestamp")).isTrue();
+    assertThat(body.get("timestamp").isTextual()).isTrue();
+    assertThat(body.has("details")).isTrue();
+  }
+}
--- a/engine/src/main/java/org/operaton/bpm/engine/health/HealthService.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/health/HealthService.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2025 the Operaton contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.operaton.bpm.engine.health;
+
+/**
+ * SPI for obtaining Operaton health information independent of the runtime.
+ *
+ * @author <a href="mailto:tomnm77@gmail.com">Tomasz Korcz</a>
+ * @since 1.1
+ */
+public interface HealthService {
+
+  /**
+   * Perform a health check and return a {@link HealthResult}.
+   */
+  HealthResult check();
+}
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/health/DefaultHealthService.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/health/DefaultHealthService.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2025 the Operaton contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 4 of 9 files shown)


distro/run/core/src/test/java/org/operaton/bpm/run/test/health/HealthEndpointTest.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
```
kthoms: Operaton header

distro/run/core/src/test/java/org/operaton/bpm/run/test/health/HealthEndpointTest.java
```diff
@@ -0,0 +1,41 @@
...
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
```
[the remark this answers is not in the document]
kthoms: Legit. Please change.

engine/src/main/java/org/operaton/bpm/engine/impl/health/DefaultHealthService.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  @Override
+  public HealthResult check() {
+    String timestamp = Instant.now().toString();
+
+    Map<String, Object> details = new LinkedHashMap<>();
+
+    boolean jobExecutorActive = jobExecutor != null && jobExecutor.isActive();
+    boolean engineRegistered = jobExecutor != null && jobExecutor.engineIterator().hasNext();
```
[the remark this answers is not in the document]
kthoms: can't be null

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
```diff
@@ -0,0 +1,102 @@
...
+      frontend = new LinkedHashMap<>(frontendHealthContributor.frontendDetails());
+    } else {
+      frontend = new LinkedHashMap<>();
+      frontend.put("operational", "unknown");
+    }
+    details.put("frontend", frontend);
+
+    boolean dbOk = (dataSource == null) || dbConnected;
+    boolean jobOk = (jobExecutor == null) || jobExecutorActive;
+    String status = (dbOk && jobOk) ? "UP" : "DOWN";
```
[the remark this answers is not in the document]
kthoms: not necessary

spring-boot-starter/starter-webapp-core/src/main/java/org/operaton/bpm/spring/boot/starter/webapp/SpringWebappFrontendHealthContributor.java
```diff
@@ -0,0 +1,55 @@
...
+
+  SpringWebappFrontendHealthContributor(ResourceLoader resourceLoader, WebappProperty webappProperty) {
+    this.resourceLoader = resourceLoader;
+    this.webappProperty = webappProperty;
+  }
+
+  @Override
+  public Map<String, Object> frontendDetails() {
+    String webjarClasspath = webappProperty != null ? webappProperty.getWebjarClasspath() : "/META-INF/resources/webjars/operaton";
+    String applicationPath = webappProperty != null ? webappProperty.getApplicationPath() : "/operaton";
```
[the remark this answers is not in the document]
kthoms: not necessary

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
```diff
@@ -0,0 +1,102 @@
...
+    Map<String, Object> queue = new LinkedHashMap<>();
+    queue.put("available", engineRegistered);
+    details.put("queue", queue);
+
+    Map<String, Object> frontend;
+    if (frontendHealthContributor != null) {
+      frontend = new LinkedHashMap<>(frontendHealthContributor.frontendDetails());
+    } else {
+      frontend = new LinkedHashMap<>();
+      frontend.put("operational", "unknown");
```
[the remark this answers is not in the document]
kthoms: seems legit. please rethink how this case should be handled. Either do not set the value, set it to a boolean, or introduce a tri-state.

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
```diff
@@ -0,0 +1,102 @@
...
+    Map<String, Object> jobExec = new LinkedHashMap<>();
+    jobExec.put("active", jobExecutorActive);
+    jobExec.put("engineRegistered", engineRegistered);
+    details.put("jobExecutor", jobExec);
+
+    boolean dbConnected = false;
+    String dbError = null;
+    if (dataSource != null) {
+      try (Connection c = dataSource.getConnection()) {
+        dbConnected = c != null && c.isValid(2);
```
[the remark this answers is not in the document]
kthoms: A constant makes sense here, e.g. `DATASOURCE_CONNECTION_TIMEOUT`

distro/run/core/src/main/java/org/operaton/bpm/run/health/HealthController.java
kthoms, line 1: The REST API is usually defined in the `engine-rest` API. This would make the health endpoint just available for the `Operaton` distribution.
I think that is fine, as installations on Tomcat or Wildfly would have their own endpoints. Am I correct?

spring-boot-starter/starter/src/main/java/org/operaton/bpm/spring/boot/starter/actuator/OperatonHealthIndicator.java
```diff
@@ -0,0 +1,61 @@
...
+
+  public OperatonHealthIndicator(HealthService healthService) {
+    Assert.notNull(healthService, "healthService must not be null");
+    this.healthService = healthService;
+  }
+
+  @Override
+  protected void doHealthCheck(Health.Builder builder) {
+    HealthResult result = healthService.check();
+    boolean up = result != null && "UP".equalsIgnoreCase(result.status());
```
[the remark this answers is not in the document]
kthoms: not needed

spring-boot-starter/starter/src/main/java/org/operaton/bpm/spring/boot/starter/actuator/OperatonHealthIndicator.java
kthoms, line 1: IMO this should be merged into `ProcessEngineHealthIndicator`.
The `JobExecutorHealthIndicator` could be made obsolete, but its check goes beyond what the HealthService is adding. Maybe the HealthService should be extended?

engine/src/main/java/org/operaton/bpm/engine/health/HealthService.java
kthoms, line 1: The package structure should be changed. The current package is too high-level and details are not hidden.

I think the both interfaces and the result record should go to `org.operaton.bpm.engine.health` and the `DefaultHealthService` to `org.operaton.bpm.engine.impl.health`.

Also mark the public interfaces/record with "since 1.1" in Javadoc

kthoms (changes requested):
In general, a very valuable addition. Let's make this now ready. There are some nitpicks from Copilot and a general discussion related the package structure.

In general for such an addition extensive tests are required. There should be at least one integration test in `spring-boot-starter/starter/src/test/java`. And unit tests for each added component.

Please press "Re-request Review" when you are ready for the next review.

distro/run/core/src/main/java/org/operaton/bpm/run/health/HealthController.java
javahippie, line 1: The run distribution is based on Spring Boot, I'm wondering what the benefit is of providing a custom health endpoint over the Spring Boot actuator? The OperatonHealthIndicator was added in the code below, this would be a good fit in my opinion

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
```diff
@@ -0,0 +1,102 @@
...
+    this.jobExecutor = jobExecutor;
+    this.frontendHealthContributor = frontendHealthContributor;
+    this.version = DefaultHealthService.class.getPackage() != null
+      ? DefaultHealthService.class.getPackage().getImplementationVersion()
+      : null;
+  }
+
+  @Override
+  public HealthResult check() {
+    String timestamp = Instant.now().toString();
```
javahippie: This will always generate a formatted String in UTC, which might be a deviation to other Health Checks potential users have in place

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
```diff
@@ -0,0 +1,102 @@
...
+    if (frontendHealthContributor != null) {
+      frontend = new LinkedHashMap<>(frontendHealthContributor.frontendDetails());
+    } else {
+      frontend = new LinkedHashMap<>();
+      frontend.put("operational", "unknown");
+    }
+    details.put("frontend", frontend);
+
+    boolean dbOk = (dataSource == null) || dbConnected;
+    boolean jobOk = (jobExecutor == null) || jobExecutorActive;
```
javahippie: Having an inactive Job Executor might be a valid choice, I would not expect the health check to fail in this case

javahippie (commented):
Hi, thanks for your contribution!

I'm wondering if the way this health check is structured is grouping too many things together. A database health check and the engine health check might be better treated as two different things. The engine can only be healthy if the database is healthy.

As an example: If I were to write my own Spring Boot application using the Operaton Starter, I already have an actuator in place which does readiness and liveness checks on the database. Duplicating this database health check into the `DefaultHealthService` would then not be necessary, because another subsystem is already taking care of this. I could reuse this in my health check, stating "Operaton can only be healthy if the database is healthy".

For the service checks themselves, I'm worndering how the check for the JobExecutor and the check for the Webapps could ever be negative. Have you checked if it is even possible to start an Operaton based application if the JobExecutor cannot be created? This is a central part of the system and my guess would be that the whole engine initialization failed in this case – including the HealthService. 

For the webapps, the test checks if the WebJars are on the classpath. From my understanding this would only be possible if a user failed to include the dependencies during build time, so this would not really be a check for correct configuration and runtime behavior.

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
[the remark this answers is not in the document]
Tomnm1: After reconsideration, I’d say keeping it as a boolean is a reasonable choice. Since we first check whether we have a `frontendHealthContributor`—meaning, colloquially, “do I even have a frontend?”—we can assume that if there is no frontend, then it is not operational. On the other hand, using a tri-state value would provide clearer information about whether the frontend actually exists

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
Tomnm1: Thanks! `OffsetDateTime.now().toString()` should be suitable

engine/src/main/java/org/operaton/bpm/health/DefaultHealthService.java
Tomnm1: I was considering the use of Operation in a HA environment and assuming that the JobExecutor would be turned off to avoid split-brain scenarios, but that wouldn’t necessarily mean the application is unhealthy. Under that assumption, it could give a completely incorrect result, so right that shall be removed.

distro/run/core/src/main/java/org/operaton/bpm/run/health/HealthController.java
Tomnm1: That's correct — only the Operation Run distro keeps the HTTP health. The only thing I'm considering now is whether it should be made ConditionalOnProperty, since that would give users the option to disable it if they want. On the other hand, I think we should assume that the health check should be enabled by default.

Tomnm1: Prefer Spring Boot Actuator for Spring Boot apps. The custom endpoint is only useful as a lightweight /health for Run (e.g., load balancer pings) or in environments where Actuator isn’t enabled and could be a baseline for user-specific implementation of health. The OperatonHealthIndicator integrates Operaton health into Actuator and is the recommended approach in Spring Boot apps.

Tomnm1:
I am still not sure about the details being hidden enough and if it's still not trying to group too many things together, but would like to check out with you and exchange thoughts

distro/run/core/src/main/java/org/operaton/bpm/run/health/HealthController.java
kthoms: Sounds good. Having it ConditionalOnProperty where the property is assumed to be true by default seems good.

kthoms:
The build error will disappear after rebase. There was a problem in restructuring the integration build.

Tomnm1:
> The build error will disappear after rebase. There was a problem in restructuring the integration build.

@kthoms Sorted out

Tomnm1:
I've fix the mess I've caused with the rebase, PR looks clear, howver there are some issues with tests - also on when building from clear main.

Tomnm1:
@kthoms if you could please take a look at this one :)

kthoms:
This PR has become obsolete by https://github.com/operaton/operaton/pull/2730

PR #1624, opened 2025-11-22
Remove Spring Boot Starter Dependencies from engine-spring

cachescrubber:
### Remove Spring Boot Starter Dependencies from engine-spring

The **engine-spring**  module should only depend on **spring-framework**, not **spring-boot**. Therefore I removed **spring-boot-starter-validation** and **spring-boot-starter-data-jpa**. Hibernate is needed in test scope only to support `SpringJobExecutorTest`. For reference, compare the camunda 7 original.

### DIFF
--- a/engine-spring/pom.xml
+++ b/engine-spring/pom.xml
@@ -34,14 +34,6 @@
       <version>${project.version}</version>
     </dependency>
     <dependency>
-      <groupId>org.springframework.boot</groupId>
-      <artifactId>spring-boot-starter-data-jpa</artifactId>
-    </dependency>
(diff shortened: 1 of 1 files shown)


kthoms:
Nitpick: I have renamed the new label. Having a concrete minor version in the label would be inflational. I would like to use labels more generic. But we have to rethink if certain fixes should go to multiple versions.

kthoms:
Commit https://github.com/operaton/operaton/commit/7e1a9e7021b4670826f70732e285a171647baf47 cherry-picked to branch `release/1.0.x`

Issue #1618, opened 2025-11-20
Cannot inject qualified `operatonBpmDataSource`


javahippie:
**Describe the bug**
This change leads now to a bug, where you cannot define the Operaton Datasource somewhere in your spring-context.
We provide a FactoryBean which fills the `operatonBpmDataSource`-Qualifier.

Since the Qualifier is missing in the constructor, Spring just pick a random Datasource and puts it into the `operatonDataSource`-Parameter.

I guess this issue needs to be reopened and the Qualifier should also placed on the Constructor.

A fix for this is the following snipped, where one can overwrite the `OperatonDatasourceConfiguration`

```
@Bean public OperatonDatasourceConfiguration operatonDatasourceConfiguration(OperatonBpmProperties operatonBpmProperties, PlatformTransactionManager operatonTransactionManager, @Qualifier(YOUR_CUSTOM_BEAN) DataSource operatonWorkDataSource) { return new DefaultDatasourceConfiguration(operatonBpmProperties, operatonTransactionManager, Optional.of(operatonTransactionManager), operatonWorkDataSource, Optional.of(operatonWorkDataSource)); }
```

(reported by @Waize here: https://github.com/operaton/operaton/issues/1053#issuecomment-3556804898)

javahippie:
This was resolved by #1686 and (afaik) shipped with 1.0.1 today, is that correct, @kthoms ?

kthoms:
Yes, that is correct. I'll close this now as completed.

PR #1686, opened 2025-11-26
#1618 moves declaration of bean qualifiers from field to factory method parameter

shiftyseaotter:
fix(spring-boot-starter): 

- moves declaration of bean qualifiers from field to factory method parameter

relates to #1618


kthoms:
cherry-picked to branch `release/1.0.x`

Issue #1609, opened 2025-11-19
Cleanup on ProcessEngineExtension breaks following testcases if configured in abstract class


Waize:
**Describe the bug**

In our testsuite we have a few abstract classes which provides an processEngineConfiguration for lots of tests: 
```java
public abstract class AbstractTest {

    public static ProcessEngine myProcessEngine = ProcessEngineConfiguration
            .createStandaloneInMemProcessEngineConfiguration()
            .setHistory("full")
            .setJdbcUrl("jdbc:h2:mem:operaton;DB_CLOSE_DELAY=1000")
// more advanced configuration, i.e. own plugins
            .buildProcessEngine();

    @RegisterExtension
    static final ProcessEngineExtension processEngine = ProcessEngineExtension.builder().useProcessEngine(myProcessEngine).build();

}
```

After migrating to Operaton we faced some  issues with running the whole testsuite. 
Given are two testclass-implementations: 
```java
@Deployment(resources = {"testresourceA.bpmn"})
class TestA extends AbstractTest{


    @Test
    void testDeployment() {
        // do sth
    }
}
```

```java
@Deployment(resources = {"testresourceB.bpmn"})
class TestB extends AbstractTest{


    @Test
    void testDeployment() {
        // do otherstuff 
    }
}
```

If we ran now our full suite, the TestA is initializing the processEngineExtension with our configuration and is executing fine, because all of our Plugins are available. 
The TestB will fail because we now have a fresh processEngine, but our plugins are mussing.

What happened?
In `afterAll`-Method from the [ProcessExtension](https://github.com/operaton/operaton/blob/main/engine/src/main/junit5/org/operaton/bpm/engine/test/junit5/ProcessEngineExtension.java#L311) the processEngine is set to `null` after the first testclass. 
Now comes the second iteration, where the extension faces a `null`-processEngine and creates a default new one without the given configuration. This needs the user to provide a operaton.cfg.xml.

We could navigate around this problem with using the `configurator()`-Method from the `ProcessEngineExtension`. 

From our point of view, the cleanup in the `afterAll`-Method is a little bit to harsh. 
Either the flag "closeProcessEngine` should be respected in the `finally`-block or the extension should safe the initial given processEngine/configuration and reinitialize the extension for the next run.

kthoms:
I will take care of this. Likely next week.

kthoms:
Fixed via https://github.com/operaton/operaton/pull/1719
Backported for 1.0.2

PR #1719, opened 2025-12-01
fix(engine): ProcessEngineExtension closes engine and removes references in afterAll only if a closeEngine flag is set

shiftyseaotter:
fix(engine): 

- ProcessEngineExtension closes engine and removes references in afterAll only if a closeEngine flag is set

relates to #1609


shiftyseaotter:
> Makes sense.
> 
> Could you add a test case to ProcessEngineExtensionJunit5Test? You might to work with `@Nested` to handle test cases for a default configuration and one for closeEngine.

I made a separate test that should cover both configuration possibilities

kthoms:
The build failure is unrelated. Looks like an instability due to introduction of `mvnd`.

kthoms:
@shiftyseaotter I had a deeper look now at your test. It was failing. Then I refactored it to make the assertions a bit clearer, i.e. testing on expected instances instead of just the engine names.

Since this was still failing I revised the bug description. From that I have now created `ProcessEngineExtensionEngineRetentionTest2`. This resembles the reported class structure. This fails with the original `ProcessEngineExtension` and succeeds with your change.

Please have a look again. If you agree I would delete the `ProcessEngineExtensionEngineRetentionTest` and rename `ProcessEngineExtensionEngineRetentionTest2` to `ProcessEngineExtensionEngineRetentionTest`.

shiftyseaotter:
@kthoms can you point me to the build where build fails with intial test configuration (process engine name comparison)?
Refactoring with ProcessEngine variables won't work due to two ProcessEngineExtensions simultaneously updating test class instance in ProcessEngineExtensions.postProcessTestInstance method

kthoms:
@shiftyseaotter Sorry my fault. The test was failing on my machine only and this because the extension class was not up-to-date. It seems a Maven install is needed to use the changed extension class.

I have reverted the test changes. When the build is green, it is good to merge.

kthoms:
cherry-picked to `release/1.0.x`

PR #1736, opened 2025-12-03
Add workflow to automatically check Slack URL and create an issue

aounhaider1:
This pull request adds a GitHub Action that checks whether the Slack invite link in the README is still valid.
Slack redirects differently for valid, expired, or incorrect invites, but this logic is executed via JavaScript in the browser. Because curl cannot evaluate these redirects, the workflow uses Playwright with system Chromium to determine the final URL.

The workflow will:
  - Extract the Slack invite link from the README
  - Open it in a headless browser
  - Detect whether it resolves to a valid /join/... page
  - Identify expired or invalid links (redirecting to /signup or /join/*/error)
  - Automatically create an issue if the link is no longer valid
  - Avoid duplicates by checking for an existing issue first
  - Assign the issue directly to @kthoms
  
  Here is the [link](https://github.com/aounhaider1/operaton/issues/7) to created issue.
  And here [a link](https://github.com/aounhaider1/operaton/actions/runs/19891756987/job/57012103640) to workflow run.
  Closes https://github.com/operaton/operaton.org/issues/39

### DIFF
--- a/.github/workflows/slack-link-check.yml
+++ b/.github/workflows/slack-link-check.yml
@@ -48,11 +48,14 @@ jobs:
 
       - name: Check Slack invite status
         id: check
+        env:
+          SLACK_URL: ${{ steps.extract.outputs.url }}
+          CHROMIUM_PATH: ${{ steps.find_chromium.outputs.path }}
         run: |
           node - << 'EOF'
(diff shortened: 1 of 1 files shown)


aounhaider1:

PR #1745, opened 2025-12-05
chore(build): Execute npm build in development mode on local machine

kthoms:
This change configures the npm builds for
- engine-rest/docs
- webapps to use "npm install" instead of "npm ci" on local machines, while using "npm ci" on GitHub actions. Further, configure webpack to use the "development" mode & configuration.

This speeds up local builds.

### DIFF
--- a/engine-rest/docs/pom.xml
+++ b/engine-rest/docs/pom.xml
@@ -12,6 +12,7 @@
   <description>${project.name}</description>
   <properties>
     <skip-third-party-bom>true</skip-third-party-bom>
+    <npm.build.args>install</npm.build.args>
   </properties>
   <dependencies>
     <dependency>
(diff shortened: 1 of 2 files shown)


alfh:
@kthoms This change causes the "distro/run/distro/target/operaton-bpm-1.1.0-SNAPSHOT.tar.gz" that I build on my Ubuntu 25.10 machine to not work properly.
I extract the tar file, and start it using "start.sh". It starts fine, but none of the webapps work, the browser just hangs forever, not loading the cockpit, for example.

I've used "git bisect", and "660b909001e31c0797d9cb5f4c2e518590e829ff" is the last good commit, while "c43ab3d26ec9844d1f9b0a1e1358c396399e82df" is the first bad one.

Am I supposed to build in a different way than "mvn clean install" now, when I want to test things locally ?

PR #1804, opened 2025-12-09
feat(engine,engine-dmn): Introduce SPI configuration factories

kthoms:
This change adds 2 factory interfaces
- ProcessEngineConfigurationFactory
- DmnEngineConfigurationFactory

The factories are used to decouple the configuration classes
- ProcessEngineConfiguration
- DmnEngineConfiguration 
from their implementation classes.

**Motivation**
This is a non-breaking preparational step to decouple the API from its implementation. The abstract classes ProcessEngineConfiguration and DmnEngineConfiguration are direct dependent on their implementation.

### DIFF
--- a/engine-dmn/engine/src/main/java/org/operaton/bpm/dmn/engine/spi/DmnEngineConfigurationFactory.java
+++ b/engine-dmn/engine/src/main/java/org/operaton/bpm/dmn/engine/spi/DmnEngineConfigurationFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 the Operaton contributors.
+ * Copyright 2026 the Operaton contributors.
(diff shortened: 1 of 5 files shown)


engine-dmn/engine/src/main/java/org/operaton/bpm/dmn/engine/impl/DmnEngineConfigurationFactoryImpl.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2025 the Operaton contributors.
```
kthoms:
```suggestion
 * Copyright 2026 the Operaton contributors.
```

PR #1831, opened 2025-12-11
chore(engine): Fix Sonar Findings (#511)

LimpBizgit:
Please merge as one squashed commit.

### DIFF
--- a/engine/src/test/java/org/operaton/bpm/engine/impl/util/ExceptionUtilTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/impl/util/ExceptionUtilTest.java
@@ -31,60 +31,103 @@ class ExceptionUtilTest {
 
   @Test
   void getExceptionStackTraceReturnsStackTrace() {
+    // given
     Throwable throwable = new Throwable();
     StackTraceElement stackTraceElement = new StackTraceElement("ExceptionUtil", "getExceptionStackTrace",
         "ExceptionUtil.java", 55);
     throwable.setStackTrace(new StackTraceElement[]{stackTraceElement});
 
-    String stackTrace = ExceptionUtil.getExceptionStacktrace(throwable).replaceAll("\\s", "");
+    // when
+    String stackTrace = ExceptionUtil.getExceptionStacktrace(throwable);
 
-    assertThat(stackTrace).isEqualTo("java.lang.ThrowableatExceptionUtil.getExceptionStackTrace(ExceptionUtil.java:55)");
+    // then
+    assertThat(stackTrace).isEqualTo("java.lang.Throwable\n\tat ExceptionUtil.getExceptionStackTrace(ExceptionUtil.java:55)\n");
   }
 
   @Test
   void checkValueTooLongException() {
-    assertThat(ExceptionUtil.checkValueTooLongException(mock(SQLException.class))).isFalse();
+    // when
+    boolean resultForUnknown = ExceptionUtil.checkValueTooLongException(mock(SQLException.class));
+    // then
+    assertThat(resultForUnknown).isFalse();
 
+    // given
     SQLException tooLong = mock(SQLException.class);
     doReturn("too long").when(tooLong).getMessage();
-    assertThat(ExceptionUtil.checkValueTooLongException(tooLong)).isTrue();
+
+    // when
+    boolean resultForTooLong = ExceptionUtil.checkValueTooLongException(tooLong);
+    // then
+    assertThat(resultForTooLong).isTrue();
   }
 
   @Test
   void checkConstraintViolationException() {
-    assertThat(ExceptionUtil.checkConstraintViolationException(new ProcessEngineException(new PersistenceException(mock(SQLException.class))))).isFalse();
+    // given
+    PersistenceException pe = new PersistenceException(mock(SQLException.class));
+    // when / then
+    assertThat(ExceptionUtil.checkConstraintViolationException(new ProcessEngineException(pe))).isFalse();
 
+    // given
     SQLException constraintViolation = mock(SQLException.class);
     doReturn("ora-00001").when(constraintViolation).getMessage();
-    assertThat(ExceptionUtil.checkConstraintViolationException(new ProcessEngineException(new PersistenceException(constraintViolation)))).isTrue();
+
+    // when
+    boolean result = ExceptionUtil.checkConstraintViolationException(new ProcessEngineException(new PersistenceException(constraintViolation)));
+    // then
+    assertThat(result).isTrue();
   }
 
   @Test
   void checkForeignKeyConstraintViolation() {
-    assertThat(ExceptionUtil.checkForeignKeyConstraintViolation(mock(SQLException.class))).isFalse();
+    // when
+    boolean unknown = ExceptionUtil.checkForeignKeyConstraintViolation(mock(SQLException.class));
+    // then
+    assertThat(unknown).isFalse();
 
+    // given
     SQLException constraintViolation = mock(SQLException.class);
     doReturn("integrity constraint").when(constraintViolation).getMessage();
-    assertThat(ExceptionUtil.checkForeignKeyConstraintViolation(constraintViolation)).isTrue();
+
+    // when
+    boolean result = ExceptionUtil.checkForeignKeyConstraintViolation(constraintViolation);
+    // then
+    assertThat(result).isTrue();
   }
 
   @Test
   void checkVariableIntegrityViolation() {
-    assertThat(ExceptionUtil.checkVariableIntegrityViolation(new PersistenceException(mock(SQLException.class)))).isFalse();
+    // when
+    boolean none = ExceptionUtil.checkVariableIntegrityViolation(new PersistenceException(mock(SQLException.class)));
+    // then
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/util/JsonUtil.java
```diff
@@ -684,32 +684,32 @@ public static Gson getGsonMapper() {
...
-      })
-      .create();
+        .serializeNulls()
+        .registerTypeAdapter(Map.class, (JsonDeserializer<Map<String, Object>>) (json, typeOfT, context) -> {
+          Map<String, Object> map = new HashMap<>();
+
+          getObject(json)
+              .entrySet()
+              .stream()
+              .filter(entry -> entry != null && entry.getValue() != null)
```
kthoms: entry can't be null, right? Did you want to check the key for not null?
effective, value can't be null, too, but this requires knowledge about the JSON parsing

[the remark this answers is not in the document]
LimpBizgit: In the original implementation there were to if statements, both checking that entry.getValue() is not null. Only in those cases an entry was added to the map. Otherwise not. So I'd argue, that the new implementation is correct.

engine/src/test/java/org/operaton/bpm/engine/impl/util/ExceptionUtilTest.java
```diff
@@ -29,6 +29,18 @@
...
 class ExceptionUtilTest {
 
+  @Test
+  void getExceptionStackTraceReturnsStackTrace() {
+    Throwable throwable = new Throwable();
+    StackTraceElement stackTraceElement = new StackTraceElement("ExceptionUtil", "getExceptionStackTrace",
+        "ExceptionUtil.java", 55);
+    throwable.setStackTrace(new StackTraceElement[]{stackTraceElement});
+
+    String stackTrace = ExceptionUtil.getExceptionStacktrace(throwable).replaceAll("\\s", "");
```
kthoms: nit (optional): I would not remove all whitespace, maybe replace by single space or hyphen?

engine/src/test/java/org/operaton/bpm/engine/impl/util/JsonUtilTest.java
```diff
@@ -0,0 +1,101 @@
...
+    assertThat(result).containsKey("nullKey");
+    assertThat(result).doesNotContainKey("objectKey");
+    assertThat(result).doesNotContainKey("arrayKey");
+  }
+
+  @Test
+  void createGsonMapper_emptyObject() {
+    String json = "{}";
+
+    Map<String, Object> result = JsonUtil.createGsonMapper().<Map<String, Object>>fromJson(json, Map.class);
```
kthoms: nit: I prefer `var` in tests when the type is clear. Which should, because it is explictly dereferenced for `fromJson`

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
```diff
@@ -0,0 +1,281 @@
...
+  void shouldConvertEmptyPermissionArray() {
+    Permission[] permissions = PermissionConverter.getPermissionsForNames(new String[] {},
+        Resources.PROCESS_DEFINITION.resourceType(), processEngineConfiguration);
+
+    assertThat(permissions).isEmpty();
+  }
+
+  @Test
+  void shouldConvertSinglePermissionName() {
+    DefaultPermissionProvider permissionProvider = new DefaultPermissionProvider();
```
kthoms: please add comments
```
// given
...
// when
...
// then
...
```

Also for other tests

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  @Test
+  void shouldThrowExceptionForUnknownPermissionName() {
+    DefaultPermissionProvider permissionProvider = new DefaultPermissionProvider();
+    when(processEngineConfiguration.getPermissionProvider()).thenReturn(permissionProvider);
+    String[] names = { "read" };
+    int resourceType = Resources.PROCESS_DEFINITION.resourceType();
+
+    Exception exception = assertThrows(Exception.class,
```
kthoms: I'd prefer AssertJ's `assertThatThrownBy`. This makes the assertions more fluent and compact.

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
```diff
@@ -0,0 +1,281 @@
...
+        () -> PermissionConverter.getPermissionsForNames(names, resourceType, processEngineConfiguration));
+
+    String expectedMessage = String.format("The permission '%s' is not valid for '%s' resource type.", names[0],
+        Resources.PROCESS_DEFINITION.name());
+    assertThat(exception).hasMessageContaining(expectedMessage);
+  }
+
+  @Test
+  void shouldConvertDuplicatePermissionNames() {
```
kthoms: Is it really desired to have duplicate names returned? Surely it is the current behavior. But is it correct?

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
```diff
@@ -0,0 +1,281 @@
...
+        processEngineConfiguration);
+
+    Stream<String> permissionNames = Arrays.stream(permissions).map(Permission::getName);
+    assertThat(permissionNames).containsExactly(names);
+  }
+
+  @Test
+  void shouldHandlePermissionsForResourceTypeNotInPermissionEnums() {
+    DefaultPermissionProvider permissionProvider = new DefaultPermissionProvider();
+    when(processEngineConfiguration.getPermissionProvider()).thenReturn(permissionProvider);
```
kthoms: These both setup lines are duplicated many times. Move to field/setup method?

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
```diff
@@ -0,0 +1,281 @@
...
+    when(authorization.isEveryPermissionGranted()).thenReturn(true);
+    Permission[] permissions = { Permissions.READ, Permissions.UPDATE };
+
+    String[] names = PermissionConverter.getNamesForPermissions(authorization, permissions);
+
+    assertThat(names).containsExactly("ALL");
+  }
+
+  @Test
+  void shouldReturnAllForRevokeAuthorizationWithAllPermissionsRevoked() {
```
kthoms, line 167: Combine all "ReturnAll" tests to parametrized test and pass in the Authorization type

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
```diff
@@ -0,0 +1,281 @@
...
+
+  @Test
+  void shouldConvertDuplicatePermissionsToNames() {
+    when(authorization.getAuthorizationType()).thenReturn(Authorization.AUTH_TYPE_GRANT);
+    when(authorization.isEveryPermissionGranted()).thenReturn(false);
+    Permission[] permissions = { Permissions.READ, Permissions.UPDATE, Permissions.READ };
+
+    String[] names = PermissionConverter.getNamesForPermissions(authorization, permissions);
+
+    assertThat(names).containsExactly("READ", "UPDATE", "READ");
```
kthoms: see above: Duplicates desired?

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
kthoms: I think the permissions should be deduplicated

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
kthoms: done

engine/src/test/java/org/operaton/bpm/engine/impl/util/JsonUtilTest.java
kthoms: Removed the type info before `fromJson`as it is redundant

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
kthoms: I have overseen that the revoke case is special, but the other 2 can be combined

engine/src/test/java/org/operaton/bpm/engine/impl/util/PermissionConverterTest.java
kthoms: changed the implementation to deduplcate

kthoms:
I have updated the code and plan to merge when ready

Issue #1898, opened 2025-12-17
Create auto-merge workflow


kthoms:
**What needs to be done?**

We are receiving frequently updates from dependabot. Most of them can be merged when all checks have been passed. But not all of them. For example, changes to DB drivers require integration tests against the respective database, but this is not automated yet. Also 3rd party GitHub actions should not be automatically merged.

Since it is more selective we cannot make use of GitHub's auto-merge feature so far.

A new workflow "auto-merge" shall be created. It should determine if the PR is eligible for auto-merging.

It shall read a configuration file with whitelisted patterns of dependencies. The configuration file (yaml) shall list patterns dependent on labels on the PR.

Example (`.github/workflows/auto-merge.config`):
```
- author: 'dependabot'
  labels: [ 'dependencies:maven']
  includes:
  - 'org.openrewrite*' # dependencies starting with 'org.openrewrite'
- author: 'dependabot'
  labels: [ 'dependencies:npm']
  includes:
  - '*' # all npm dependencies can be merged after successful build
```

The workflow should run scheduled at 4 AM or on manual dispatch. It should list all open PRs that have been passed all checks and match the configuration. The eligible PRs should be collected in an array. This array is then taken by another step to approve the PR and then merge it using the "squash" option.

kthoms:
Done

PR #1901, opened 2025-12-17
Fixes to auto-merge workflow

kthoms:
This change fixes problems with the initial workflow.

Related to #1898


kthoms:
@copilot 
- The PR should only be approved when it does not have conflicts
- When conflicts exist add a comment `@dependabot rebase`

PR #1979, opened 2026-01-06
chore(engine): Fix Sonar issues

LimpBizgit:
related to #511

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
@@ -98,7 +98,7 @@ public final class ProcessEngines {
 
     ClassLoader classLoader = ReflectUtil.getClassLoader();
     Set<URL> configResources = getResources(classLoader, "operaton.cfg.xml", "activiti.cfg.xml", forceCreate);
-    if (configResources.isEmpty()) {
+    if (configResources.isEmpty() && !forceCreate) {
       return;
     }
 
@@ -107,7 +107,7 @@ public final class ProcessEngines {
     }
 
     Set<URL> springResources = getResources(classLoader, "activiti-context.xml", null, forceCreate);
-    if (springResources.isEmpty()) {
+    if (springResources.isEmpty() && !forceCreate) {
       return;
     }
 


kthoms:
The build failures seem unrelated. I have seen such failures before. I'll check them. Anyway, already thank you for the PR!

engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
```diff
@@ -95,53 +95,63 @@ public static synchronized void init(boolean forceCreate) {
...
-        if (forceCreate) {
-          throw new ProcessEngineException(
-              "problem retrieving operaton.cfg.xml and activiti.cfg.xml resources on the classpath: "
-                  + System.getProperty("java.class.path"),
-              ex);
-        } else {
-          return;
-        }
-      }
+    Enumeration<URL> configResources = getResources(classLoader, "operaton.cfg.xml", "activiti.cfg.xml", forceCreate);
```
kthoms: getResources() should not return null, but empty

engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
```diff
@@ -95,53 +95,63 @@ public static synchronized void init(boolean forceCreate) {
...
+      String message = "problem retrieving " + resourceName
+          + (fallbackResourceName != null ? " and " + fallbackResourceName : "") + " resources on the classpath: "
+          + System.getProperty("java.class.path");
+      throw new ProcessEngineException(message, e);
     }
+    return null;
+  }
 
-    isInitialized = true;
+  private static Set<URL> collectResourcesAsSet(Enumeration<URL> resources) {
```
kthoms: This could be a utility method in CollectionUtil

engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
```diff
@@ -95,53 +95,63 @@ public static synchronized void init(boolean forceCreate) {
...
+
+    while (springResources.hasMoreElements()) {
+      URL resource = springResources.nextElement();
+      initProcessEngineFromSpringResource(resource);
+    }
+
+    isInitialized = true;
+  }
+
+  private static Enumeration<URL> getResources(ClassLoader classLoader, String resourceName,
```
kthoms: return non-null Set

engine/src/main/java/org/operaton/bpm/engine/impl/HistoricCaseInstanceQueryImpl.java
```diff
@@ -157,26 +157,26 @@ public HistoricCaseInstanceQuery createdBefore(Date date) {
...
     return this;
   }
 
   @Override
   public HistoricCaseInstanceQuery closedBefore(Date date) {
-    if (state!= null && (!state.equals(CaseExecutionState.CLOSED.getStateCode()))) {
-      throw new NotValidException("Already querying for case instance state '%s'".formatted(state));
-    }
-
+    throwExceptionIfStateNotClosed();
```
kthoms: I would rename this. `throwException` is unspecific, it is a NotValidException that is thrown.
I would rather align it to the naming of EnsureUtil's methods, i.e. `ensureStateClosedWhenSet`.

engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
kthoms: done

engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
kthoms: done

engine/src/main/java/org/operaton/bpm/engine/ProcessEngines.java
kthoms: done

engine/src/main/java/org/operaton/bpm/engine/impl/HistoricCaseInstanceQueryImpl.java
kthoms: done

PR #1988, opened 2026-01-07
Resolve Serialization issues

kthoms:


javahippie:
@kthoms what was the origin of this issue, a build warning? I'd like to know where and how these classes are serialized and why the non-serializable fields are safe to annotate as transient?

PR #2020, opened 2026-01-12
refactor(engine): Removed 'volatile' modifier from ProcessEngineConfgurationImpl fields

kthoms:
The fields have been made volatile due to flaky tests during the JUnit 5 migration. This PR removes the `volatile` modifiers from fields of `ProcessEngineConfgurationImpl`.


kthoms:
Multiple builds without observing flakiness

PR #2028, opened 2026-01-12
Refactor: Resolve setting static fields from instance methods

kthoms:
related to #511


kthoms:
@copilot fix the findings

PR #2048, opened 2026-01-14
Resolve various Sonar issues

kthoms:
related to #511


kthoms:
Test failure is unrelated, happens also on `main` currently

PR #2201, opened 2026-01-26
chore(ci): Make the Publish job more resilient

kthoms:
When multiple builds are running in parallel, the build artifact cache might be gone. It is not a problem to not have build artifacts published then. The job must not fail.

Made the remaining steps dependend on the cache-hit output


kthoms:
Checked per manual workflow execution

PR #2235, opened 2026-01-30
refactor(tests): run email tasks tests in parallel

fabcipriano:
I confirm that my contribution and following ones comply with the [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0)
and the [Code of Conduct](https://github.com/operaton/operaton/blob/main/CODE_OF_CONDUCT.md)

This is a refactor to use Testcontainers and Mailpit: https://testcontainers.com/modules/mailpit/ for https://github.com/operaton/operaton/issues/2021

PR #2239, opened 2026-01-31
Fix port collision in email tests when running in parallel

rushikeshfulmali27:
### What this PR does
Fixes intermittent test failures caused by port collisions when email-related tests are executed in parallel.

### Root cause
Email tests were binding to fixed ports, which caused conflicts under parallel execution.

### Solution
Introduced dynamic port allocation to ensure isolation between parallel test runs.

### Verification
Tests were executed locally using:
./mvnw -pl engine -am test -Dsurefire.includes="**/Email*.java"

All Email-related tests pass consistently.

### Related tests
- EmailSendTaskTest
- EmailServiceTaskTest


kthoms:
Thank you @rushikeshfulmali27 for your contribution.

Unfortunately I have to decline your PR, because:
1) Someone else signaled that he is working on the task and he wrote this in the issue
2) The PR from this contributor uses a solution based on Testcontainers as proposed in the issue
3) You did not accept the contribution agreement so far as requested in the comment above

This PR is superseeded by https://github.com/operaton/operaton/pull/2235 . That change will remove the dependency on Wiser, which makes this change conflicting.

Thank you anyway for your contribution. I hope to see another next time. Please do not forget to comment on the task before  starting work.

PR #2254, opened 2026-02-02
chore(various): resolve sonar findings

LimpBizgit:
related to #511

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/cmmn/handler/SentryHandler.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/cmmn/handler/SentryHandler.java
@@ -54,54 +54,26 @@ public class SentryHandler extends CmmnElementHandler<Sentry, CmmnSentryDeclarat
 
   @Override
   public CmmnSentryDeclaration handleElement(Sentry element, CmmnHandlerContext context) {
+
     String id = element.getId();
     Collection<OnPart> onParts = element.getOnParts();
     IfPart ifPart = element.getIfPart();
     List<OperatonVariableOnPart> variableOnParts = queryExtensionElementsByClass(element, OperatonVariableOnPart.class);
 
-    if (!isSentryValid(id, ifPart, onParts, variableOnParts)) {
-      return null;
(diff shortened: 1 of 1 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/cmmn/model/CmmnActivity.java
```diff
@@ -186,39 +186,46 @@ public void addExitCriteria(CmmnSentryDeclaration exitCriteria) {
    * the listener is defined on.
    */
   public Map<String, List<VariableListener<?>>> getVariableListeners(String eventName, boolean includeCustomListeners) {
-    Map<String, Map<String, List<VariableListener<?>>>> listenerCache;
+    Map<String, Map<String, List<VariableListener<?>>>> listenerCache = getVariableListenerCache(includeCustomListeners);
+
+    return listenerCache.computeIfAbsent(eventName, k -> resolveVariableListeners(eventName, includeCustomListeners));
+  }
+
+  protected Map<String, Map<String, List<VariableListener<?>>>> getVariableListenerCache(boolean includeCustomListeners) {
```
kthoms: General note: Please avoid using the `protected` scope for newly introduced helper methods. This implies extensibility which may not be needed and makes these methods potential public API. Yes, it public API, because you could potentially derive a class that could override. Especially for classes that do not reside in an "impl" subpackage.

Use
- `private` as far as possible
- package protected (i.e. no modifier) when the method is tested explicitly by a unit test

engine/src/main/java/org/operaton/bpm/engine/impl/cmmn/handler/SentryHandler.java
```diff
@@ -90,18 +109,6 @@ public CmmnSentryDeclaration handleElement(Sentry element, CmmnHandlerContext co
     // the variableOnParts will be initialized immediately as it does not have any dependency
     initializeVariableOnParts(element, sentryDeclaration, context, variableOnParts);
 
-    // ...whereas the onParts will be initialized later because the
```
kthoms: this removal is a behavior change, restore

kthoms:
Fixed my findings. Currently there are problems with the GitHub runners. Let's wait until they can be used again.

PR #2339, opened 2026-02-10
chore(various): fix Sonar issues

LimpBizgit:
related to #511

### DIFF
--- a/engine/src/test/java/org/operaton/bpm/engine/impl/json/JsonTaskQueryConverterTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/impl/json/JsonTaskQueryConverterTest.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2026 the Operaton contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


commons/utils/src/main/java/org/operaton/commons/utils/IoUtil.java
```diff
@@ -64,13 +64,16 @@ public static byte[] inputStreamAsByteArray(InputStream inputStream) {
    * @deprecated Do not use. This method will be removed.
    */
   @Deprecated(since = "1.1", forRemoval = true)
+  @SuppressWarnings("java:S2093") // exception on closing reader is ignored on purpose
```
kthoms: This has been refactored to try-with-resources recently. Please rebase and revert this change.

engine/src/test/java/org/operaton/bpm/engine/impl/json/JsonTaskQueryConverterTest.java
```diff
@@ -0,0 +1,74 @@
+package org.operaton.bpm.engine.impl.json;
```
kthoms, line 17: add Operaton license header

kthoms (approved):

PR #2399, opened 2026-02-16
chore(various): remove usage of Thread.sleep()

LimpBizgit:
related to #511


kthoms (approved):
Looks good to me. Many thanks!

PR #2404, opened 2026-02-17
WIP: Add ADR 0002 - Preview Features

ungerts:
## What
This PR adds a new Architecture Decision Record (ADR): `docs/decisions/0002-preview-features.md`.

## Why
Operaton ships frequently and benefits from early feedback on new functionality. At the same time, committing too early to long-term compatibility guarantees (especially around BPMN/DMN semantics and persistence behavior) increases risk and maintenance cost.

This ADR proposes a formal concept of **Preview Features**, aligned with the principles of **JEP 12** (explicit opt-in, time-boxed, and no long-term compatibility guarantees).

### DIFF
--- a/docs/decisions/0002-preview-features.md
+++ b/docs/decisions/0002-preview-features.md
@@ -1,0 +1,162 @@
+---
+status: proposed
+date: 2025-12-23
+decision-makers: Operaton Core Maintainers
+consulted: Operaton Engine Contributors, API Maintainers
+informed: Operaton Community, Extension Authors
+---
+
+# Introduce Preview Features in Operaton
+
+## Context and Problem Statement
+
+Operaton evolves rapidly and follows a high-frequency release cadence with multiple releases per year. New functionality often benefits from early integration and real-world feedback, but committing prematurely to permanent compatibility guarantees increases long-term risk and integration effort.
+
+At the same time, Operaton implements standards such as BPMN 2.0 and DMN. Changes in those areas must be introduced carefully to avoid inconsistent semantics, unexpected persistence effects, or partial standard coverage becoming de facto permanent. This creates a natural tension with fast releases: we want rapid delivery but must avoid locking in brittle or underspecified interpretations of BPMN/DMN constructs.
+
+The question to be answered is: **Should Operaton introduce a formal concept of Preview Features to allow early integration of new functionality without committing to permanent compatibility guarantees?**
+
+In particular, we want to align the semantics of preview features with the principles defined in JEP 12 for Java language preview features: clearly scoped, explicitly opt-in, time-boxed, and without long-term compatibility guarantees.
+
+### Definition and Scope
+
+A **Preview Feature** is functionality that is available to users for early evaluation but:
+
+* is **explicitly opt-in** (default is off),
+* is **time-boxed** (must be promoted or removed), and
+* carries **no compatibility guarantees** (signatures, behavior, configuration, or persistence details may change without deprecation; database state may not be migrated automatically on a new release or when switching the feature on/off).
+
+The preview concept applies to public surfaces such as Java APIs, REST APIs, configuration properties, and end-user behavior.
+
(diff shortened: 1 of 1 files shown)


docs/decisions/0002-preview-features.md
```diff
@@ -0,0 +1,162 @@
...
+* Avoid fragmentation of the Operaton platform
+
+## Considered Options
+
+* Introduce Preview Features
+* Preview via isolated modules/extensions only (new artifacts, no in-place changes)
+* Label-only preview (documentation/release notes only, no opt-in)
+* Do not introduce Preview Features
+
+## Decision Outcome
```
kthoms, line 49: General question: Should be decision outcome already be part of a _proposed_ ADR?

docs/decisions/0002-preview-features.md
```diff
@@ -0,0 +1,162 @@
...
+### Confirmation
+
+Compliance with this ADR can be confirmed by the following fitness functions:
+
+* Every preview feature is clearly labeled as **Preview** in documentation and release notes.
+* Every preview feature has an **explicit opt-in** that is **off by default**.
+* Each preview feature documents its **time box** and its **promotion/removal criteria**.
+* Reviews verify that preview surfaces do not introduce permanent compatibility guarantees (for example: no promise of stable API/package names, stable REST contract, or stable persistence/schema behavior while in preview).
+* A periodic review (at least once per release) decides whether each preview feature is promoted, extended explicitly, or removed.
+
```
kthoms, line 70: A preview feature and all related changes have to be tracked by a GitHub issue that is labeled (e.g. "preview")

Code related to public API changes for a preview feature is marked/annotated.

docs/decisions/0002-preview-features.md
```diff
@@ -0,0 +1,162 @@
...
+### Preview via isolated modules/extensions only (new artifacts, no in-place changes)
+
+Preview functionality is shipped only in isolated modules or extensions (new artifacts) so that existing engine and API surfaces remain unchanged.
+
+* Good, because it minimizes risk of accidental adoption and protects the core engine surface.
+* Good, because it can reduce the chance of partially implemented BPMN/DMN semantics leaking into stable behavior.
+* Neutral, because it can still gather feedback from early adopters.
+* Bad, because it increases packaging and dependency complexity.
+* Bad, because it can delay integration into the real engine/runtime and create long-lived divergence.
```
kthoms, line 91: Bad, because it might be impossible to deliver a feature as separate artifact, as it requires access/changes to API in core modules

hauptmedia:
I like it! The explicit opt-in, time-boxing, and clear "no compatibility guarantees" rule make the expectations very transparent for users and contributors. Aligning the concept with Java’s preview model is also a smart move, since it gives developers an already familiar mental model.

ungerts:
I am currently considering three types of changes to better understand the consequences of this approach:

- Changing public APIs (e.g., adding methods to control `AdHocSubprocess`)
- Changing the persistence layer (e.g., optimizing table layouts)
- Changing BPMN behavior (e.g., modifying the execution behavior of `AdHocSubprocess`)

javahippie:
I have no further remarks beyond the ones that @hauptmedia and @kthoms already mentioned and like the idea. How can we move forward with this? Is this ADR good enough for us to agree on it?

javahippie:
A problem I see: What if a new feature is introduced in the Java API? A runtime flag is not able to prevent developers from accessing a method, and a startup check cannot be done. This behavior of Java Preview features we cannot reproduce, we cannot add a compile time flag that only makes the API "visible" via code completion or allows the compiler to access a method when the flag is present, at least not without relying on any tools being present in the users compile chain / build tool.

Worst case: A developer uses an API because code completion suggests it to them, doesn't realize it's a preview feature (I know, they should know better, see Deprecation warnings maybe) ships it to production and it causes a runtime error only when called.

PR #2448, opened 2026-02-22
fix(engine-cdi): ensure CDI producers are discoverable with annotated bean discovery

hauptmedia:
Add explicit `@Dependent` bean-defining annotations to engine-cdi producer/support classes that previously relied on archive-wide discovery, and add a regression test that deploys with bean-discovery-mode="annotated".

This hardens CDI 4.x/container compatibility and prevents unsatisfied dependency errors caused by missing producer beans.

This was reported in this forum thread: https://forum.operaton.org/t/weld-001408-unsatisfied-dependencies-for-engine-services/402

### DIFF
--- a/engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/BusinessProcess.java
+++ b/engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/BusinessProcess.java
@@ -20,6 +20,7 @@ import java.io.Serial;
 import java.io.Serializable;
 import java.util.Map;
 import jakarta.enterprise.context.Conversation;
+import jakarta.enterprise.context.Dependent;
 import jakarta.enterprise.context.RequestScoped;
 import jakarta.enterprise.inject.Instance;
 import jakarta.inject.Inject;
@@ -82,6 +83,7 @@ import org.operaton.bpm.engine.variable.value.TypedValue;
  * @author Daniel Meyer
  * @author Falko Menge
  */
+@Dependent
 @Named
 public class BusinessProcess implements Serializable {
 
--- a/engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/CurrentProcessInstance.java
+++ b/engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/CurrentProcessInstance.java
@@ -16,6 +16,7 @@
  */
 package org.operaton.bpm.engine.cdi;
 
+import jakarta.enterprise.context.Dependent;
 import jakarta.enterprise.inject.Produces;
 import jakarta.enterprise.inject.Typed;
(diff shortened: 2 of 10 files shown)


hauptmedia:
Our forum user dieter12 will provide a test program where we can test against if this PR fixes the problem

kthoms:
The test does reproduce a bug with annotated discovery, which the changes repair.

Good work, @hauptmedia !

kthoms:
Related to https://github.com/operaton/operaton/issues/2473

DieterMayerOSS:
Great fix! Adding `@Dependent` to the producer classes is indeed the cleaner approach compared to `bean-discovery-mode="all"`, as it avoids unnecessary classpath scanning and is fully CDI 4.x compliant.

However, this PR only solves part of the problem for WildFly users. There is a second, related issue that is not addressed here:

**`operaton-engine-cdi` is not packaged as a WildFly module in the distribution at all.**

Currently:
- `operaton-engine-cdi` is declared as a Maven dependency in `distro/wildfly/assembly/pom.xml`, but it is not referenced in `assembly.xml` and has no `module.xml` defined
- It does **not** appear under `modules/org/operaton/bpm/` in the WildFly distribution
- The `operaton-wildfly-subsystem` `module.xml` does not list it as a dependency

This means that even after this PR is merged, WildFly users must manually add `operaton-engine-cdi` as a `compile`-scoped dependency to their own WAR — which is neither documented nor intuitive. This is how I discovered the issue while porting from Camunda 7.18 to Operaton 2.0.0-M3.

I am currently working on a PR for the WildFly 39.0.1 upgrade (`fix/wildfly-39-jakarta-11`). I would be happy to include the proper WildFly module packaging for `engine-cdi` in that PR as well — unless there is a deliberate design decision to keep it as a WAR-only dependency (in which case it should at minimum be documented).

Related forum thread: https://forum.operaton.org/t/weld-001408-unsatisfied-dependencies-for-engine-services/402

hauptmedia:
The issue that operaton-engine-cdi is not packaged as a WildFly module in the distribution at all is also adressed in this PR now.

I also added another regression Test for it. However this test is currently failing in the CI pipeline (I guess the root cause are the cached artifacts here)

hauptmedia:
I will split this in two PRs, because I think the first part can be merged now and we can better discuss the other problem with a dedicated PR

hauptmedia:
i spllited this PR to this and #2480
So this PR is only about adding explicit Dependent bean-defining annotations again

hauptmedia:
@DieterMayerOSS thank you for your feedback i think this can be merged now, let's continue the discussion about the original problem in #2480. Your feedback is very much appreciated here.

kthoms:
This has been integrated already via #2781

PR #2453, opened 2026-02-22
JReleaser Provide snapshot label

kthoms:


kthoms:
Not needed. The snapshot label is already set by env property `JRELEASER_PROJECT_SNAPSHOT_LABEL`

PR #2471, opened 2026-02-27
feat/1573 arquillian cargo replacement

### LINKED ISSUE #1573: Rework Frontend Integration tests to use Arquillian instead of the Cargo Plugin
**What needs to be done?**

The engine integration tests for Wildfly use Arquillian. We have already combined testcontainers with Arquillian to enable a broad selection of engine integration tests. The Webapp integration tests run on cargo, which is not adaptable as easy. We should consider harmonizing the test setups.

shiftyseaotter:
feat(integration-tests):
 webapps: Integration Tests are using Arquillian instead of Cargo Plugin
closes #1573


kthoms:
@shiftyseaotter Great that you managed to refactor the tests to Arquillian. This will make the maintenance easier.

There are a few resolvable minor Sonar findings. Please address them if possible.

The tests are executing, so I assume the review is just about formal stuff. I will review that soon!

parent/pom.xml
```diff
@@ -48,7 +48,7 @@
     <version.commonj>1.1.0</version.commonj>
     <!-- application servers -->
     <version.wildfly>38.0.1.Final</version.wildfly>
-    <version.tomcat>11.0.18</version.tomcat>
```
javahippie, line 51: This donwgrades tomcat for the whole platform, is there a workaround, maybe?

qa/tomcat-runtime/src/main/conf/context.xml
```diff
@@ -23,6 +23,11 @@
     <WatchedResource>WEB-INF/web.xml</WatchedResource>
     <WatchedResource>${catalina.base}/conf/web.xml</WatchedResource>
 
+    <!--
+        https://stackoverflow.com/questions/46367851/what-causes-filenotfoundexception-pdq-jar-with-db2jcc4
+        Temporary and should be removed later when a proper fix exists
```
javahippie, line 28: Can we create and link an issue for this, so we don't forget it?

javahippie (commented):
Hi, that's a huge effort and will help us a lot, thanks for doing this! I have two review comments, only the one about Tomcat is a concern. I guess that there is no arquillian-tomcat-managed-11, yet?

parent/pom.xml
javahippie: https://github.com/arquillian/arquillian-container-tomcat/issues/182 Doesn't seem like we should wait for Tomcat 11 support...

shiftyseaotter: I dont think its a problem with arquillian. Version 10 managed the container well enough. I may have it wrong, but I had an issue with the deployment of operaton artifacts, particularly engine-rest to Tomcat 11. As they are generated against 6.0.0 Jakarta version (at least 1.x did), it didn't work at runtime with Tomcat 11, resulting in  java.lang.AbstractMethodError: Receiver class org.apache.catalina.connector.ResponseFacade does not define or inherit an implementation of the resolved method 'abstract void sendRedirect(java.lang.String)' of interface jakarta.servlet.http.HttpServletResponse
Exception. Apparently, this is a collision with the newer Jakarta version Tomcat 11 has

qa/tomcat-runtime/src/main/conf/context.xml
shiftyseaotter: Yes, i will open an issue for this

shiftyseaotter:
> @shiftyseaotter Great that you managed to refactor the tests to Arquillian. This will make the maintenance easier.
> 
> There are a few resolvable minor Sonar findings. Please address them if possible.
> 
> The tests are executing, so I assume the review is just about formal stuff. I will review that soon!

Sure, will adress all but the "ip" one :)

parent/pom.xml
kthoms: This needs a discussion elsewhere (forum/GH issue). A downgrade should be avoided when possible, or otherwise decided separate.

kthoms: Is the downgrade required for the whole application, or could it be limited to the integration tests? The Tomcat distro should be IMO on the latest version. I would not like to have splitted versions, but if it is mandatory for getting tests running on Arquilliam right now then I could live with that.

qa/integration-tests-webapps/src/test/java/org/operaton/bpm/integrationtest/AbstractWebIntegrationTest.java
```diff
@@ -47,15 +58,37 @@ public abstract class AbstractWebIntegrationTest {
...
-  protected String appUrl;
-  protected TestProperties testProperties;
-
   protected static ChromeDriverService service;
 
   protected String csrfToken;
   protected String sessionId;
 
+  @Deployment(name = "engine-rest", testable = false, order = 1)
+  public static WebArchive createRestEngineDeployment() {
```
kthoms, line 64: reduce visibility of these methods
suppress "unused" warnings

kthoms: When you are on it: Maybe reduce the visibility of other methods here, too, to what is minimal needed by the current code base. Optional, as not related to the story.

qa/integration-tests-webapps/src/test/resources/arquillian.xml
```diff
@@ -0,0 +1,35 @@
...
+            xsi:schemaLocation="http://jboss.org/schema/arquillian http://jboss.org/schema/arquillian/arquillian-1.0.xsd">
+
+  <engine>
+        <property name="deploymentExportPath">target/deployments</property>
+  </engine>
+
+  <container qualifier="wildfly">
+    <configuration>
+      <property name="jbossHome">${wildfly.runtime.location}</property>
+      <property name="managementPort">39990</property>
```
kthoms, line 12: Could you try to configure a port by property and make it use a random port?
In `clients/java/client/pom.xml` the `build-helper-maven-plugin` is used to bind a random port to a property. 

Same for Tomcat.

qa/integration-tests-webapps/src/test/resources/arquillian.xml
```diff
@@ -0,0 +1,35 @@
...
+      <property name="serverConfig">standalone.xml</property>
+      <property name="javaVmArguments">-Xmx1024m</property>
+      <!-- <property name="javaVmArguments">-Xmx1024m -Xrunjdwp:transport=dt_socket,address=8787,server=y,suspend=y</property> -->
```
kthoms, line 15: Would it make sense to define the javaVmArguments by a property and introduce a `debug-arquillian` profile that activates the debug option?

qa/integration-tests-webapps/pom.xml
```diff
@@ -97,47 +88,280 @@
...
     </dependency>
+    <dependency>
+      <groupId>org.operaton.bpm</groupId>
+      <artifactId>operaton-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.operaton.bpm</groupId>
+      <artifactId>operaton-engine</artifactId>
+      <classifier>junit6</classifier>
```
kthoms: I have removed the "junit6" classifier meanwhile again. Please use "junit5".

Main reason: The "junit6" jar contains classes with "junit5" package name. This feels inconsistent and causes confusion. It is easier to tell "just use junit5, it is compatible also for junit6". But what we really need is a "junit-jupiter" test classes artifact with classes without "junit5" in their package name.

qa/integration-tests-webapps/pom.xml
```diff
@@ -97,47 +88,280 @@
...
+        </dependency>
+      </dependencies>
+
+      <build>
+        <plugins>
+          <plugin>
+            <!-- Copy relevant wars to target so that they can be accessed from Unittests directly -->
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-dependency-plugin</artifactId>
+            <version>3.9.0</version>
```
kthoms: remove version, it is managed by root pom.xml, also for other occurrences in the changes.

In general, do not define plugin versions in modules. Manage them in /pom.xml

qa/integration-tests-webapps/README.md
```diff
@@ -1,7 +1,9 @@
 # Webapp Integration Tests
 
-This file contains integration tests for the operaton webapps.
-
+This module contains integration tests for the operaton webapps.
+Tests are executed with managed arquillian container. To run tests individually add -Darquillian.launch=tomcat/wildfly and -Dtomcat/wildfly.runtime.location to your run configuration.
```
kthoms, line 5: add empty lines between paragraphs, otherwise it will be rendered as a single paragraph

parent/pom.xml
shiftyseaotter: The main issue was, to my knowledge (I checked it in SBOM), the servlet.api dependency 6.0.0. I checked it now against latest main branch and it's 6.1.0 now, so I've updated Tomcat version back to 11.x. Local build and integration test run was successful. If GHA Pipeline run is also successful, this should not be an issue anymore.

qa/integration-tests-webapps/src/test/resources/arquillian.xml
shiftyseaotter: This is sadly a bigger and more impactful as I initially assumed. I'd would like to create a separate issue for this

kthoms:
@shiftyseaotter What is the current state? Please resolve the trivial conversations yourself when they are addressed.

When we are back on Tomcat 11 and everything is working as expected, we should be pretty close to finish this one.

qa/integration-tests-webapps/src/test/resources/arquillian.xml
kthoms: Just discussed with @javahippie : Create a follow-up issue.

qa/integration-tests-webapps/src/test/java/org/operaton/bpm/integrationtest/AbstractWebIntegrationTest.java
shiftyseaotter: I've reduced all I could sofar. Arquillian injections need to stay public to my knowledge

qa/integration-tests-webapps/src/test/resources/arquillian.xml
shiftyseaotter: I will create a followup issue then

qa/integration-tests-webapps/src/test/resources/arquillian.xml
shiftyseaotter: I've introduced a environment property for this

qa/integration-tests-webapps/pom.xml
shiftyseaotter: done

qa/integration-tests-webapps/pom.xml
shiftyseaotter: Done

shiftyseaotter:
> @shiftyseaotter What is the current state? Please resolve the trivial conversations yourself when they are addressed.
> 
> When we are back on Tomcat 11 and everything is working as expected, we should be pretty close to finish this one.

Can you please review it again? Everything should be in order now. Except for random ports, which will be handled in separate issue

qa/integration-tests-webapps/src/test/java/org/operaton/bpm/integrationtest/AbstractWebIntegrationTest.java
kthoms: ok

kthoms:
Looks good. I have squashed and rebased your changes. Planning to merge them.

kthoms (approved):
Thank you @shiftyseaotter for resolving this work

Issue #2473, opened 2026-03-01
Improve CDI integration for the WildFly distribution


kthoms:
**Describe the bug**

In [this forum thread](https://forum.operaton.org/t/weld-001408-unsatisfied-dependencies-for-engine-services/402/10) it is reported that CDI injection is not properly working on Wildfly.

The user reports that the operaton-engine-cdi module missing in the Wildfly distribution is missing.

**To Reproduce**

The user has provided a [reproducing example](https://github.com/DieterMayerOSS/operaton-cdi-test).

**Expected behavior**

The given can be deployed without error.

An integration test exists that reproduces the given example.

**Additional context**

The CDI module is missing.

<img width="395" height="771" alt="Image" src="https://github.com/user-attachments/assets/c5a2db44-8204-413d-b6de-0849a7aa9b43" />

hauptmedia:
After looking into this topic in more detail, it became clear that we first need to clarify what the intended default behavior should be. Until we have a clear agreement on that, I will put this issue on hold from my side.

Tim suggested scheduling a call so we can discuss the topic and align on the expected behavior.

javahippie:
I'd propose switching this from a bug to a feature request, as this behavior was never present in Camunda 7 and was never documented to be working

hauptmedia:
i changed the title of this issue and switched the type from bug to feature

kthoms:
Compatibility with Wildfly 39, CDI 4.1 and JakartaEE 11 has been added for 2.2.0.

PR #2480, opened 2026-03-02
fix(wildfly): provide operaton-engine-cdi as default module dependency

hauptmedia:
Splitted this PR from #2448 

_Originally posted by @DieterMayerOSS in https://github.com/operaton/operaton/issues/2448#issuecomment-3982212360_

**`operaton-engine-cdi` is not packaged as a WildFly module in the distribution at all.**

Currently:
- `operaton-engine-cdi` is declared as a Maven dependency in `distro/wildfly/assembly/pom.xml`, but it is not referenced in `assembly.xml` and has no `module.xml` defined
- It does **not** appear under `modules/org/operaton/bpm/` in the WildFly distribution
- The `operaton-wildfly-subsystem` `module.xml` does not list it as a dependency

Related forum thread: https://forum.operaton.org/t/weld-001408-unsatisfied-dependencies-for-engine-services/402

Related issue #2473

### DIFF
--- a/distro/wildfly/modules/pom.xml
+++ b/distro/wildfly/modules/pom.xml
@@ -55,6 +55,10 @@
     </dependency>
     <dependency>
       <groupId>org.operaton.bpm</groupId>
+      <artifactId>operaton-engine-cdi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.operaton.bpm</groupId>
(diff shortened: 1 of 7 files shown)


kthoms:
This PR has been replaced by https://github.com/operaton/operaton/pull/2781
The original changes are in there

Issue #2483, opened 2026-03-03
Admin-User Setup broken in 1.1.0


cCavegn:
**Describe the bug**

When starting Operaton with an empty database & opening the WebApp, the Operaton setup screen is shown. After creating the admin account, login should be possible.
However, this seems to be broken in 1.1.0 at least when using the spring boot starter.

**To Reproduce**

1. Create a new maven project with the attached pom.xml & main class (no additional classes / configuration is needed to reproduce the behavior)
2. Start the application
3. Open the web frontend
4. The setup screen is shown / the admin account can be set up. However, login doesn't work - in the network console, some 403 are shown

**Expected behavior**

Login with the newly created user should be possible

When changing the operaton version 1.1.0 -> 1.0.3 in the attached pom, the system behaves as expected

**Additional context**

pom.xml:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0"
         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
    <modelVersion>4.0.0</modelVersion>

    <parent>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter-parent</artifactId>
        <version>3.5.11</version>
        <relativePath/>
    </parent>

    <groupId>operaton-test</groupId>
    <artifactId>operaton-login-test</artifactId>
    <version>1.0-SNAPSHOT</version>

    <properties>
        <maven.compiler.source>21</maven.compiler.source>
        <maven.compiler.target>21</maven.compiler.target>
        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>

        <operaton.spring-boot.version>1.1.0</operaton.spring-boot.version>
    </properties>

    <dependencyManagement>
        <dependencies>
            <!-- Operaton -->
            <dependency>
                <groupId>org.operaton.bpm.springboot</groupId>
                <artifactId>operaton-bpm-spring-boot-starter-webapp</artifactId>
                <version>${operaton.spring-boot.version}</version>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <!-- Spring -->
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter</artifactId>
        </dependency>
        <dependency>
[...]

PR #2504, opened 2026-03-04
test(qa/webapps): add e2e regression for initial user setup login

hauptmedia:
- add browser-based webapp flow test for admin setup -> login
- place regression test in qa/integration-tests-webapps suite
- verify initial user creation and successful admin login via UI

related to #2483


javahippie:
lgtm, but I'd follow the Sonar recommendation to introduce constants for repeated string literals

PR #2505, opened 2026-03-04
chore(comments): replace legacy ticket references with behavior-focused context

hauptmedia:
- remove Camunda Jira/upstream references from code comments
- preserve relevant background as technical rationale, quirks, and expected behavior
- rewrite user-story phrasing into clear developer-facing explanations
- align wording with Operaton and Operaton (standalone)


hauptmedia:
this is currently a draft to better decide if we want to integrate the informations from the Camunda legacy ticket system directly into the code (preventing loss of this context informationen when they shut down this system anytime in the future)

PR #2527, opened 2026-03-07
fix(engine): enable process application context switch for EL express…

### LINKED ISSUE #2522: Flaky integration tests: Cross-deployment CDI bean resolution fails intermittently on Tomcat
**Describe the bug**

Several Arquillian integration tests in `qa/integration-tests-engine` fail intermittently on Tomcat in CI because CDI beans defined in one WAR deployment cannot be resolved when EL expressions are evaluated in the context of a different WAR deployment.

On Tomcat, each WAR runs its own isolated Weld CDI container. The engine resolves CDI beans via `BeanManagerLookup`, which performs a JNDI lookup (`java:comp/env/BeanManager`). The result depends on which WAR's classloader is active on the current thread. If the engine does not correctly switch to the target process application's context before evaluating the expression, the JNDI lookup returns the wrong WAR's `BeanManager`, and the bean cannot be found.

On WildFly, CDI is managed at the container level and beans are visible across deployments, so these tests pass reliably.

Affected tests:
  - `CdiBeanCaseTaskResolutionTest.testResolveBeanInCmmnCase` — `Cannot resolve identifier 'caseVariableBean'`
  - `DecisionContextSwitchTest` (all 4 test methods) — `Unable to evaluate expression for language 'juel': '${greeter.greet()}'`

 Potentially any other test that relies on cross-deployment CDI bean resolution could be affected.

 **To Reproduce**

1. Push any change that triggers the `Integration Build` workflow
2. Observe intermittent failures in the `Integration Tests (17, engine, tomcat, h2)` job
3. The failing test varies between runs but always involves a CDI bean defined in a different WAR than the one evaluating the EL expression

Recent CI failures:
- [PR #2516 (engine, tomcat, h2)](https://github.com/operaton/operaton/actions/runs/22744273860/job/65964568073?pr=2516) — CdiBeanCaseTaskResolutionTest`
- [PR #2509 (engine, tomcat, h2)](https://github.com/operaton/operaton/actions/runs/22697543068/job/65809237999) — `DecisionContextSwitchTest`

**Expected behavior**

Cross-deployment CDI bean resolution should work reliably on Tomcat, or the tests should be scoped to environments where this feature is supported (WildFly).

**Additional context**

Possible fixes:

1. **Skip affected tests on Tomcat** — These tests verify cross-deployment CDI resolution, which is an application-server feature that Tomcat with standalone Weld does not reliably support. The tests could be disabled for the Tomcat profile (e.g. via `@EnabledIfSystemProperty`) and kept for WildFly only. Least effort, most honest about what is being tested.

  2. **Make test deployments self-contained** — Add the required CDI beans to every WAR that references them in EL expressions. This eliminates the cross-deployment dependency but changes the semantics of the tests (they would no longer verify cross-deployment resolution).

  3. **Fix process application context switching in the engine** — Ensure that the engine activates the correct process application's classloader before evaluating `out` variable mappings (BPMN
  call activities, CMMN case tasks) and decision expressions. This is the correct fix but carries higher complexity and risk.

  Key files:
  - `engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/impl/util/BeanManagerLookup.java`
  - `engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/impl/el/CdiResolver.java`
  - `qa/integration-tests-engine/src/test/java/org/operaton/bpm/integrationtest/functional/cdi/CdiBeanCaseTaskResolutionTest.java`
  - `qa/integration-tests-engine/src/test/java/org/operaton/bpm/integrationtest/functional/el/DecisionContextSwitchTest.java`

javahippie:
…ion evaluation

Fixes #2522

When evaluating EL expressions via JuelExpression.getValue(VariableScope), the contextExecution was always null, preventing the DefaultDelegateInterceptor from switching to the correct process application context.

This caused cross-deployment CDI bean resolution to depend on the nondeterministic static BeanManagerLookup.localInstance on Tomcat, leading to flaky test failures (e.g. CdiBeanCaseTaskResolutionTest, DecisionContextSwitchTest).

Pass the variableScope as contextExecution when it is a BaseDelegateExecution, enabling the interceptor to determine the target process application and perform the classloader context switch before JNDI-based BeanManager lookup.


javahippie:
This fix might be dangerous, or might not be, I could really use some input here.

The fix propagates the context execution, so even on Tomcat the matching delegate can find the deployment from which the classes should be looked up (very simplified). In a previous fix, the instanceof check used too broad a class and accidentally overwrote an existing context, making the CMMN tests break. With the fix like this ALL TESTS and integration tests now reliably work, but I have a really bad gut feeling about overriding existing contexts in this use case, even if none of the many, many, many unit tests break now.

kthoms:
Actually I can't judge this really. Could we maybe use a feature switch to enable this? And activate this for our tests?

I am also wondering what could have caused this. We did not have problems with the latest release. I need to investigate this deeper. For now, I would be happy if we could fix our tests.

javahippie:
I'm pretty sure it started appearing with the new CDI version.

I wouldn't just enable this for the tests, because the tests show that a behavior was changed. All unit tests, all integration tests are green, and I was not yet able to find an edge case in tests.

kthoms:
Yes, maybe on by default. But just for the case we broke something that we were not expecting here, we could point to a switch and do not need to ship a new release. I have slight gut feelings here. Just as a fallback.

javahippie:
The root cause was fixed by https://github.com/operaton/operaton/pull/2781 already, not needed anymore.

PR #2533, opened 2026-03-09
refactor(various): replace singletons

LimpBizgit:
related to #511


LimpBizgit:
There are two errors in ArchRulesTest becuase Connectors.getInstance() was removed and the field availableConnectors is now static. Should I undo this change or can I fix it some other way?

kthoms:
I need to carefully think about this. The failure rgd. Connectors is right. The instance is public and therefore potentially public API and could be used by clients. I doubt we can refactor that.
Could take a bit time until I come back.

Issue #2187, opened 2026-01-25
BoundaryErrorEventTest: Handling of uncaught errors


kthoms:
**Describe the bug**

The test `BoundaryErrorEventTest` has some test cases for handling of uncaught errors. The test was wrong, because it did not fail when _no_ exception is thrown:

```
  void testUncaughtError() {
    runtimeService.startProcessInstanceByKey("simpleSubProcess");
    Task task = taskService.createTaskQuery().singleResult();
    assertThat(task.getName()).isEqualTo("Task in subprocess");

    try {
      // Completing the task will reach the end error event,
      // which is never caught in the process
      taskService.complete(task.getId());
    } catch (BpmnError e) {
      testRule.assertTextPresent("No catching boundary event found for error with errorCode 'myError', neither in same process nor in parent process", e.getMessage());
    }
  }
```

The error was uncovered by refactoring of the assertion to AssertJ:
```
    assertThatThrownBy(() -> taskService.complete(task.getId()))
        .isInstanceOf(BpmnError.class)
        .hasMessageContaining("No catching boundary event found for error with errorCode 'myError', neither in same process nor in parent process");
```

The test fails, because actually no exception was thrown.

The affected tests have been disabled. It should be clarified if it is expected that no exception is thrown, or if that uncovered a bug.

Fejbien:
Due to isEnableExceptionsAfterUnhandledBpmnError being false by default, the errors were only logged and not thrown in the testUncaughtError and testUncaughtErrorOnCallActivity tests. For testUncaughtErrorThrownByJavaDelegateOnCallActivity and testUncaughtErrorThrownByJavaDelegateOnServiceTask, an exception was thrown, but the error message and instance type were different.

PR #2604, opened 2026-03-18
test(engine): fix assertions for uncaught BPMN error tests

Fejbien:
- re-enables tests
- updates assertion

related to #2187


Fejbien:
I will look into these errors and fix them tomorrow.

kthoms:
@Fejbien I do not see why the build would fail because of your change, it is unrelated. However, we have to make sure tests pass.
Maybe it is just flaky? I just triggered the build again.

Fejbien:
@kthoms The failure was related to my change, enableExceptionsAfterUnhandledBpmnError was never reset after the test, leaking into ExecutionListenerTest which expected the default false. Fixed by resetting it back to false at the end of each affected test.

kthoms (approved):
LGTM; thank you for resolving the test failure

PR #2781, opened 2026-04-20
Wildfly 39 Upgrade & CDI 4.1 compatibility

DieterMayerOSS:
# fix(engine-cdi): make BeanManager lookup and CdiResolver WildFly-module-safe

## Summary

This PR builds on top of Julian Haupt's `fix/wildfly-missing-cdi-dependency` and fixes two follow-up issues that surface when `operaton-engine-cdi` is provided as a WildFly system module.

When deploying a process application on WildFly 39 with the new `operaton-engine-cdi` module, EL expressions like `${dateTime().plusWeeks(1).toDate()}` failed with `WELD-001408 / Could not lookup beanmanager in jndi`, even though `dateTime()` is not a CDI bean.

## Root Cause

1. The `operaton-engine-cdi` module runs in the WildFly system module classloader, where `java:comp/BeanManager` JNDI lookup is not available.
2. `CdiResolver` is part of a `CompositeELResolver` chain but throws a hard exception when no BeanManager is found, breaking the entire chain instead of delegating to the next resolver.

### `BeanManagerLookup.java`

Reordered lookup strategy to use the Jakarta CDI 4.1 standard API as primary path:

1. Custom JNDI name (if configured)
2. **`CDI.current().getBeanManager()`** ← Jakarta CDI 4.1 standard (works in WildFly module classloader)
3. `java:comp/BeanManager` (legacy JNDI fallback for app servers)
4. `java:comp/env/BeanManager` (legacy JNDI fallback for servlet containers)
5. `localInstance` (manually configured)

### `CdiResolver.java`

Made all `ELResolver` methods null-safe. When no BeanManager is available, methods return `null` (or sensible defaults) and the exception is logged at `FINE` level. This allows the `CompositeELResolver` chain to delegate to the next resolver, so non-CDI expressions like `dateTime()` resolve correctly.

### `module.xml` (`operaton-engine-cdi`)

Marked APIs as `optional="true"` that are not used by all deployments:

- `jakarta.faces.api` (only used by `TaskForm`)
- `jakarta.enterprise.concurrent.api` (only used by `ManagedJobExecutor`)
- `jakarta.transaction.api` (not directly imported)

## Result

- `@Inject IdentityService` works in servlets without manual configuration
- EL expressions like `${dateTime().plusWeeks(1).toDate()}` work correctly
- **No `jboss-deployment-structure.xml` is required** in process application WARs — the `ModuleDependencyProcessor` handles everything automatically


DieterMayerOSS:

kthoms:
closes https://github.com/operaton/operaton/pull/2448

closes https://github.com/operaton/operaton/issues/2473

distro/wildfly/modules/src/main/modules/org/operaton/bpm/operaton-engine-cdi/main/module.xml
kthoms, line 1: Must not use CRLF. Please fix your IDE settings for future changes.

I'll update the file.

engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/impl/util/BeanManagerLookup.java
```diff
@@ -44,40 +45,47 @@ public static BeanManager getLocalInstance() {
...
       } catch (NamingException e) {
-        throw new ProcessEngineException("Could not lookup beanmanager in jndi using name: '%s'.".formatted(jndiName), e);
+        throw new ProcessEngineException("Could not lookup BeanManager in JNDI using name: '%s'.".formatted(jndiName), e);
       }
     }
 
+    // CDI 1.1+ standard programmatic lookup (Jakarta CDI 4.1)
+    try {
+      return CDI.current().getBeanManager();
+    } catch (IllegalStateException e) {
```
kthoms: Should catch RuntimeException. Other CDI implementations might not throw ISE.

I'll fix that.

kthoms:
@DieterMayerOSS Integration tests are failing. Execute

```
.devenv/scripts/build/build-and-run-integration-tests.sh --distro=wildfly
```

for reproduction

DieterMayerOSS:
Hi Karsten,

Pushed another commit (8b0b778) that refactors the integration tests to align with operaton-engine-cdi being provided as a WildFly system module. Locally all 250 tests now pass (9 skipped — 4 legacy test classes that exercise patterns no longer carried by the Jakarta EE 11 migration, documented in the new qa/integration-tests-engine/README.md).

Could you kick off the CI when you get a chance? Thanks!

qa/integration-tests-engine/src/test/java/org/operaton/bpm/integrationtest/deployment/cfg/TestCustomProcessesXmlFileLocation.java
```diff
@@ -41,14 +42,13 @@ public class TestCustomProcessesXmlFileLocation extends AbstractFoxPlatformInteg
   @Deployment
   public static WebArchive processArchive() {
 
-    return ShrinkWrap.create(WebArchive.class, "test.war")
+    return TestContainer.addEngineCdiLib(ShrinkWrap.create(WebArchive.class, "test.war")
```
kthoms: IMO the pattern to integrate engine-cdi or not should be changed. Instead of wrapping the created archive, create the archive first, and then apply adding CDI depending on the container. Several tests add `TestContainer.addXXX` afterwards (e.g. TestWarDeploymentWithBrokenBpmnXml). I think we should follow the pattern here.

Means: Store the archive as variable, then call `addEngineCdiLib`:

```
var deployment = ShrinkWrap.create( ... );

TestContainer.addEngineCdiLib(deployment);

return deployment;
```

kthoms:
@DieterMayerOSS Your last commit adds a lot of noise.

Please check out this PR as alternative: https://github.com/operaton/operaton/pull/2810/changes
It is based on this one, but reverts your last commit and provides an alternative implementation to consider on Wildfly that engine-cdi is available as system module.

qa/integration-tests-engine/src/test/java/org/operaton/bpm/integrationtest/deployment/cfg/TestCustomProcessesXmlFileLocation.java
kthoms: This became obsolete, the changes to tests have been reverted

kthoms:
I'm planning to merge when build is green

DieterMayerOSS:
my build is green. Many thanks Karten

PR #2810, opened 2026-04-22
Fix/wildfly 39 upgrade 2

kthoms:


kthoms:
replaced by https://github.com/operaton/operaton/pull/2781

PR #3090, opened 2026-06-09
fix(engine-rest): handle task OR query candidates

hauptmedia:
## Description

Backports the EximeeBPMS fix for REST task OR queries with candidate user and candidate group filters.

`TaskQueryDto` previously translated every REST `orQueries` entry into a detached `TaskQueryImpl` and attached it via `addOrQuery`. That bypasses the normal `query.or()` / `endOr()` path used by the engine query API and can mishandle combinations such as separate REST OR blocks for `candidateUser` and `candidateGroups`.

This change skips empty OR DTOs and applies all non-empty REST OR filters through the real `TaskQuery` by opening one `or()` block, applying the nested filters, and closing it with `endOr()`.

Operaton already has engine-level coverage for candidate-user/group OR query behavior in `TaskQueryOrTest`, so this backport keeps the focused REST-layer regression coverage and adapts the filter interaction test to the resulting single OR query structure.

## References

EximeeBPMS:
- PR: https://github.com/EximeeBPMS/eximeebpms/pull/116
- Commit: https://github.com/EximeeBPMS/eximeebpms/commit/d483358f8c1cbd3c72d2f8878777f3f5b78dc331
- Ticket reference: BPMS-151

## Attribution

Backported commit `d483358f8c1cbd3c72d2f8878777f3f5b78dc331` from the `eximeebpms` repository.

Original author: Robert Mastalerek <robert.mastalerek@gmail.com>

Backport change unit: 628

## Testing

```bash
./mvnw -pl engine-rest/engine-rest -Dskip.frontend.build=true -Dtest=TaskRestServiceQueryTest#testOrQuery+testOrQueryWithCandidateUserAndCandidateGroups+testOrQueryWithEmptyOrQueryAndCandidateUser,FilterRestServiceInteractionTest#testExecuteSingleResultWithExtendingOrQuery test
```

```bash
./mvnw -pl engine-rest/engine-rest -DskipTests -Dskip.frontend.build=true package
```

### DIFF
--- a/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/dto/task/TaskQueryDto.java
+++ b/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/dto/task/TaskQueryDto.java
@@ -1118,14 +1118,8 @@ public class TaskQueryDto extends AbstractQueryDto<TaskQuery> {
 
   @Override
   protected void applyFilters(TaskQuery query) {
-    if (orQueries != null) {
-      for (TaskQueryDto orQueryDto: orQueries) {
-        TaskQueryImpl orQuery = new TaskQueryImpl();
-        orQuery.setOrQueryActive();
-        orQueryDto.applyFilters(orQuery);
-        ((TaskQueryImpl) query).addOrQuery(orQuery);
-      }
-    }
+    applyOrQueries(query);
+
     if (processInstanceBusinessKey != null) {
       query.processInstanceBusinessKey(processInstanceBusinessKey);
     }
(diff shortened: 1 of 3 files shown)


javahippie:
This changes the current behavior and needs to be considered as a breaking change.

Today each entry becomes its own OR group and the groups are AND'd together (CNF) — that's what the TaskQueryDto docs describe ("a task must fulfill at least one predicate of each query") and what the multi-entry examples in post.ftl show. With this change all entries are merged into a single or() block, so the result becomes one big disjunction instead of an AND of groups.

The adapted FilterRestServiceInteractionTest seems to confirm it — the assertions move from two OR groups to one, which also affects extended stored filters.

So for single-entry orQueries (the common case) everything's fine, but multi-entry queries would return different results than before and than the docs state. Could you confirm  whether that semantic change is intended? If so, it'd probably be worth updating the docs, calling it out as breaking and scheduling it for 3.0; if not, keeping a separate OR block per entry would preserve the current CNF behavior.

hauptmedia:
After looking at this again, I do not think we should merge this PR as-is.

The behavior enabled by this patch is already expressible with the existing REST API. If the intended query is:

```text
candidateGroups contains "testGroup" OR candidateUser = "testUser"
```

clients can already send a single nested OR query:

```json
{
  "orQueries": [
    {
      "candidateGroups": ["testGroup"],
      "candidateUser": "testUser"
    }
  ]
}
```

Using two separate entries currently has a different documented meaning:

```json
{
  "orQueries": [
    { "candidateGroups": ["testGroup"] },
    { "candidateUser": "testUser" }
  ]
}
```

This means CNF semantics: a task must match at least one predicate from each nested query. The REST docs and examples describe this as:

```text
(A OR B) AND (C OR D)
```

The patch collapses all nested entries into one `query.or()` block and therefore changes the meaning to:

```text
A OR B OR C OR D
```

That is a breaking change for existing clients and stored filters. The adapted filter interaction test also confirms this semantic change by reducing two OR groups to one.

So I recommend rejecting this PR in its current form. If there is a real bug around candidate user/group handling, it should be fixed while preserving the existing CNF semantics. If changing the semantics is intentional, it should be treated as a breaking change, documented accordingly, and targeted for a major release instead.

kthoms:
Following your reasoning I would not want this behavioral change.

What I would consider is the basic refactoring done in the change. I mean, factor out the filter application logic into a separate method and checking if any filter is set. But maybe this is adding too much code for a likely very small gain.

hauptmedia:
i reject this PR for now, if you have a need for improvements here in the future we have to open a new PR

PR #3091, opened 2026-06-09
docs: generalize backport agent guidance

hauptmedia:


AGENTS.md
```diff
@@ -330,19 +338,26 @@ feat(engine): Support BPEL execution
 related to #123
 ```
 
-### Backporting from Camunda 7
+### Backporting from Camunda 7-compatible Sources
 
-For issues labeled `backport:c7`:
-- Attribute the original commit properly
+For issues labeled `backport:c7`, treat the source as the Camunda 7 compatibility family. The source may be upstream
```
kthoms: Also other labels, actually all "backport:" starting labels except for "backport:patch-release"

AGENTS.md
```diff
@@ -330,19 +338,26 @@ feat(engine): Support BPEL execution
...
+Camunda 7 or a Camunda 7-compatible fork.
+
```
kthoms, line 345: Also: Copyright headers for new files MUST be copied without change

kthoms (approved):

PR #3098, opened 2026-06-09
fix(springboot): validate trailing slash redirect target

hauptmedia:
## Summary
- Backports OrqueIO commit https://github.com/OrqueIO/OrqueIO/commit/5f12e717c13fa4b8638b758f7c2f345b16dd67d6 to harden `AppendTrailingSlashFilter`.
- Change unit: `4854`
- Keeps normal trailing-slash redirects for local absolute paths such as `/operaton/app`.
- Passes unsafe or non-local redirect targets down the filter chain instead of sending a redirect.
- Adds regression coverage for local paths, absolute URLs, scheme-relative URLs, relative paths, and missing request URIs.

## Reviewer Notes
Previously, `AppendTrailingSlashFilter` used `request.getRequestURI()` directly as the redirect target after appending `/`. If a servlet container or proxy exposed a non-local URI here, the filter could emit an external redirect. This change only redirects local paths and rejects values such as `https://example.org/app` and `//example.org/app`, while preserving the intended Spring Boot 2-style trailing-slash behavior for normal Operaton app routes.

No original OrqueIO PR was recorded for this change unit; the source attribution is the commit above.

## Tests
- `.devenv/scripts/maintenance/code-cleanup.sh`
- `git diff --check`
- `./mvnw -pl spring-boot-starter/starter-webapp-core -Dtest=AppendTrailingSlashFilterTest -Dskip.frontend.build=true test`

### DIFF
--- a/spring-boot-starter/starter-webapp-core/src/main/java/org/operaton/bpm/spring/boot/starter/webapp/filter/AppendTrailingSlashFilter.java
+++ b/spring-boot-starter/starter-webapp-core/src/main/java/org/operaton/bpm/spring/boot/starter/webapp/filter/AppendTrailingSlashFilter.java
@@ -17,6 +17,7 @@
 package org.operaton.bpm.spring.boot.starter.webapp.filter;
 
 import java.io.IOException;
+
 import jakarta.servlet.Filter;
 import jakarta.servlet.FilterChain;
 import jakarta.servlet.ServletException;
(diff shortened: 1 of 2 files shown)


spring-boot-starter/starter-webapp-core/src/test/java/org/operaton/bpm/spring/boot/starter/webapp/filter/AppendTrailingSlashFilterTest.java
```diff
@@ -0,0 +1,77 @@
...
+    when(request.getRequestURI()).thenReturn("/operaton/app");
+
+    filter.doFilter(request, response, chain);
+
+    verify(response).sendRedirect("/operaton/app/");
+    verify(chain, never()).doFilter(request, response);
+  }
+
+  @Test
+  void shouldContinueFilterChainForUnsafeRedirectTargets() throws Exception {
```
kthoms: Change to ParametrizedTest with ValueSource

spring-boot-starter/starter-webapp-core/src/main/java/org/operaton/bpm/spring/boot/starter/webapp/filter/AppendTrailingSlashFilter.java
kthoms: This issue has been closed June 2022. Is this filter really needed anymore? Is it still required for SpringBoot 4?

hauptmedia:
Context for this change: the trailing-slash filter is still needed for the registered webapp entry paths (`/app`, `/app/cockpit`, `/app/admin`, `/app/tasklist`, `/app/welcome`). Transparent trailing-slash matching was deprecated in Spring Framework 6 and removed in Spring Framework 7, so these legacy entry URLs need an explicit redirect to their canonical trailing-slash form.

The redirect target validation is a small defensive hardening around that compatibility filter: only local absolute request paths are reflected into the `Location` header; absolute or protocol-relative values continue through the filter chain instead of becoming redirect targets.

kthoms:
Thank you for the clarification

kthoms (approved):
LGTM

kthoms:
Cherry-picked for 2.1.2

Issue #78, opened 2024-10-29
Add Code Coverage to the Project


javahippie:
**Is your feature request related to a problem? Please describe.**
We would like to visualize code coverage in our pipeline / sonarqube analysis to identify untested coce

**Describe the solution you'd like**
We should add the JaCoCo Maven plugin

javahippie:
Code coverage appears in Sonarcloud, it seems to be working: https://sonarcloud.io/code?id=operaton_operaton&selected=operaton_operaton%3Aengine%2Fsrc%2Fmain%2Fjava%2Forg%2Foperaton%2Fbpm

javahippie:
@kthoms I don't think there is anything left to do here, would propose closing this issue?

kthoms:
Yes, can be closed.

PR #3104, opened 2026-06-09
chore(run): Disable default SSL in production profile

hauptmedia:
## Summary

Backports CIBSeven change unit 884 to disable the bundled SSL sample configuration in the Operaton Run `production.yml` profile.

The production profile still documents the HTTPS settings, but the `server.ssl` block and `server.port: 8443` are now commented out. This avoids starting the production profile with the bundled demo keystore and default password unless an operator explicitly enables and configures HTTPS.

## Reviewer notes

Before this change, `distro/run/assembly/resources/production.yml` actively configured:

- `server.ssl.key-store: classpath:keystore.p12`
- default sample keystore passwords
- `server.port: 8443`

That is not a good production default because the profile enables a shipped sample certificate by default. After this backport, the values remain as an example in the file but are inactive. A production deployment that wants HTTPS should provide its own SSL configuration explicitly.

Operaton adaptation:

- Kept Operaton documentation links.
- Kept Operaton-specific sample values (`operaton`, `localhost`) in the commented example.
- Applied the same behavioral change as CIBSeven: disable the SSL sample block and HTTPS port in the production profile.

## Attribution

Backported from CIBSeven:

- Source repository: https://github.com/cibseven/cibseven
- Source commit: `ba4653e5e8ee3fc99b26594f68740cdb893a4c06`
- Source title: `chore(spring-boot-distro): remove ssl config in production (#78)`
- Original author: fernando-m-g `<fernandom@cib.de>`
- Backport change unit: 884

## Verification

```bash
ruby -e 'require "yaml"; YAML.load_file("distro/run/assembly/resources/production.yml"); puts "yaml ok"'
```

Result: `yaml ok`.

```bash
./mvnw -pl distro/run -DskipTests -Dskip.frontend.build=true validate
```

Result: `BUILD SUCCESS`.

```bash
./mvnw -pl distro/run/assembly -am -DskipTests -Dskip.frontend.build=true validate
```

Result: `BUILD SUCCESS`.

```bash
git diff --check
```

Result: no whitespace issues.

### DIFF
--- a/distro/run/assembly/resources/production.yml
+++ b/distro/run/assembly/resources/production.yml
@@ -68,10 +68,10 @@ server:
 # do not use the provided certificate in production
 #  ssl:
 #    key-store: classpath:keystore.p12
-#    key-store-password: operaton
 #    key-store-type: pkcs12
-#    key-alias: localhost
-#    key-password: operaton
+#    key-store-password: REPLACE_ME
+#    key-alias: REPLACE_ME
+#    key-password: REPLACE_ME
 #  port: 8443
 
 # https://docs.operaton.org/manual/latest/user-guide/security/#http-header-security-in-webapps
--- a/distro/run/assembly/resources/production.yml
+++ b/distro/run/assembly/resources/production.yml
@@ -1,10 +1,10 @@
 # This configuration is intended for production use and was created with respect to the security guide.
 # The provided links to documentation guides for each property will give more information about the purpose of each property.
-# security guide: https://docs.operaton.org/manual/latest/user-guide/security/
+# security guide: https://docs.operaton.org/docs/documentation/user-guide/security
 
 operaton.bpm:
-# https://docs.operaton.org/manual/latest/user-guide/security/#http-header-security-in-webapps
-# https://docs.operaton.org/manual/latest/webapps/shared-options/header-security/
+# https://docs.operaton.org/docs/documentation/user-guide/security/#http-header-security-in-webapps
+# https://docs.operaton.org/docs/documentation/webapps/shared-options/header-security/
   webapp:
     csrf:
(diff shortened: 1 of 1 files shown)


distro/run/assembly/resources/production.yml
```diff
@@ -66,13 +66,13 @@ operaton.bpm:
...
-  ssl:
-    key-store: classpath:keystore.p12
-    key-store-password: operaton
-    key-store-type: pkcs12
-    key-alias: localhost
-    key-password: operaton
-  port: 8443
+#  ssl:
+#    key-store: classpath:keystore.p12
+#    key-store-password: operaton
```
kthoms: if already commented out, replace passwords and alias as placeholder
I'd place key-store and keystore-type as top values without placeholders. This is most likely not to be exchanged. And if needed, configurators know.

distro/run/assembly/resources/production.yml
```diff
@@ -66,13 +66,13 @@ operaton.bpm:
 server:
 # https://docs.operaton.org/manual/latest/user-guide/operaton-bpm-run/#https
```
kthoms:
```suggestion
# https://docs.operaton.org/docs/documentation/user-guide/security#http-header-security-in-webapps
```

kthoms (approved):

distro/run/assembly/resources/production.yml
kthoms: done

distro/run/assembly/resources/production.yml
kthoms: done, fixed also all other links

kthoms:
Cherry-picked for 2.1.2

PR #3105, opened 2026-06-09
feat(engine-rest): Configure Jackson string constraints

hauptmedia:
## Summary

Backports the CIBSeven Jackson string-length configuration feature as one combined Operaton PR.

The REST `JacksonConfigurator` now applies Jackson `StreamReadConstraints` to the `ObjectMapper`. By default it keeps Jackson's own `DEFAULT_MAX_STRING_LEN`; deployments can override it by placing a `jackson.properties` file on the classpath with:

```properties
jackson.maxStringLength=<max string length>
```

The PR also adds a JUnit 5 test for the configured limit and folds in the CIBSeven follow-up that expects Jackson's `StreamConstraintsException` when a payload exceeds the configured limit.

## Identified CIBSeven commits

This PR intentionally combines these related CIBSeven commits:

- `c7296100ac81b394cb6e10a9b37496fb46ef4603` / change unit 801: introduces configurable Jackson max string length and the original test/resource.
- `1c4bcc64ea80c867ccf93dc35e7141d4f18f2ffa` / change unit 803: fixes the test to expect `StreamConstraintsException` instead of swallowing exceptions.
- `81a8d2b2074d596e1e441996482fe6fbcbc27311` / change unit 1026: guards `setStreamReadConstraints` for mixed/older Jackson runtime modules.

Related commits that were checked but not included as separate sources:

- `200764e3d0e61f99726cdcaa0c0e034439ad9530` only changes the file mode of `jackson.properties` and mostly restructures unrelated identity DB tests.
- `7e72cec8789a69b25860f3362b2bd4befe3dc39b` is a branch-only JUnit 5 migration. Operaton already uses JUnit 5, so the test was adapted directly to Operaton style instead of attributing that branch-only commit as a mainline source.

## Reviewer notes

Operaton adaptation:

- Applied the change to `org.operaton` package paths and the Jakarta REST module.
- Added an Operaton-header JUnit 5/AssertJ test instead of the source JUnit 4 test.
- Kept `jackson.properties` optional: if it is absent, Jackson's default max string length is used without a startup warning.
- If a present `jackson.properties` cannot be read, Operaton fails using the existing `EngineUtilLogger.exceptionWhileReadingFile` path instead of silently ignoring the configuration.

No production `jackson.properties` file is shipped by this PR. The file added here is test-only and sets `jackson.maxStringLength=50000000` so the test can verify both the default and configured limits.

## Attribution

Backported from CIBSeven:

- Source repository: https://github.com/cibseven/cibseven
- Original author: Oleg Skrypnyuk `<oleg.skrypnyuk@cib.de>`
- Co-authored source contribution: Serge Krot `<serge.krot@cib.de>`
- Backport change units: 801, 803, 1026

Source commits:

- `c7296100ac81b394cb6e10a9b37496fb46ef4603`
- `1c4bcc64ea80c867ccf93dc35e7141d4f18f2ffa`
- `81a8d2b2074d596e1e441996482fe6fbcbc27311`

## Verification

```bash
./mvnw -pl engine -am -DskipTests -Dskip.frontend.build=true install
```

Result: `BUILD SUCCESS`.

```bash
./mvnw -pl engine-rest/engine-rest -Dskip.frontend.build=true -Dtest=JacksonConfiguratorTest -Dsurefire.failIfNoSpecifiedTests=false -DfailIfNoTests=false test
```

Result: `BUILD SUCCESS`; `JacksonConfiguratorTest` ran twice through the module's Surefire executions, 12 total test invocations, 0 failures/errors.

```bash
git diff --check
```

Result: no whitespace issues.

### DIFF
--- a/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/mapper/JacksonConfiguratorTest.java
+++ b/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/mapper/JacksonConfiguratorTest.java
@@ -1,5 +1,10 @@
 /*
- * Copyright 2026 the Operaton contributors.
+ * Copyright CIB software GmbH and/or licensed to CIB software GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. CIB software licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * Modifications Copyright 2026 the Operaton contributors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/mapper/JacksonConfiguratorTest.java
+++ b/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/mapper/JacksonConfiguratorTest.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2026 the Operaton contributors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at:
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.operaton.bpm.engine.rest.mapper;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.StreamReadConstraints;
+import com.fasterxml.jackson.core.exc.StreamConstraintsException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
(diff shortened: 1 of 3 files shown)


engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/mapper/JacksonConfiguratorTest.java
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright 2026 the Operaton contributors.
```
kthoms: Backported new files must preserve copyright headers

hauptmedia: Added in 5ab6ad00c3. I preserved the CIBSeven copyright attribution using the same `Copyright CIB software GmbH and/or licensed to CIB software GmbH` wording that is used by other CIBSeven-owned files in the source repository, and kept the Operaton modification notice on the backported files.

javahippie:
Should we not only read from a properties file on the classpath, but also allow an environment variable as a fallback? I agree that the property does not belong into the engine config, or matches any properties files we already have, but a file on the classpath couples the behavior to a compiled artifact and prevents runtime-specific configurations in different stages. 

I'd also suggest a different name. We don't own the name "jackson", and if the Jackson maintainers decided to suddenly introduce a "jackson.properties" config file, we'd run into a problem. operaton-jackson.properties or operaton-serialization.properties come to mind as alternatives – if we decide to keep the file, and not just user environment variables entirely. @kthoms @hauptmedia what do you think?

hauptmedia:
I understand the use case behind this feature, and I agree that it may solve a real problem for specific customers.

However, I do not think we should generally encourage users to store very large payloads as process data. The current Jackson default already allows a 20MB payload, which is a generous limit for process variables. If an application needs to handle larger data, I would recommend storing that data outside the process engine and keeping only a reference in the process data.

For that reason, my suggestion would be to reject this feature.

javahippie:
Agree with the post by  @hauptmedia above, closing

PR #3106, opened 2026-06-09
feat(engine): add fork class name compatibility mapping

hauptmedia:
## Summary

Adds an opt-in compatibility mode for migrated process applications and runtime data that still reference Camunda 7 fork package names.

When `enableForkClassNameCompatibilityMapping` is enabled, Operaton maps known fork engine package prefixes to `org.operaton.bpm` in these paths:

- reflective class loading via `ReflectUtil`
- Java-serialized object type-name fallback in `AbstractObjectValueSerializer`
- Java serialization class descriptor loading, including object array descriptors
- Java serialization proxy interface resolution
- script source compile/eval in `SourceExecutableScript`
- script engine creation through a mapping context classloader
- process-application script engine lookup via `OperatonScriptEngineManager`

Known source prefixes covered by this PR:

- `org.camunda.bpm`
- `org.cibseven.bpm`
- `org.eximeebpms.bpm`
- `org.finos.fluxnova.bpm`
- `io.orqueio.bpm`

## Problem

After migrating from Camunda 7 or another fork, deployed processes, scripts, serialized variables, or process-application configuration can still contain class names from the previous namespace. Without a compatibility layer, these references fail even when the equivalent class exists in Operaton under `org.operaton.bpm`.

The CIBSeven implementation solved this for selected `org.camunda` to `org.cibseven` cases. Fluxnova added a Java serialization stream-level remapping helper for migrated serialized variables. This PR generalizes the useful parts for Operaton and keeps the behavior fork-neutral.

## Implementation Notes

The new behavior is disabled by default via `enableForkClassNameCompatibilityMapping = false` to avoid changing class loading semantics for existing installations.

The mapping is boundary-aware. For example, `org.cibseven.bpm.engine.delegate.BpmnError` maps to `org.operaton.bpm.engine.delegate.BpmnError`, but unrelated packages such as `org.camunda.community`, `org.cibseven.bpmfoo`, or `io.orqueio.bpmfoo` are left unchanged.

For Java serialization, `ClassloaderAwareObjectInputStream` continues to resolve through Operaton's class loading utility, but now also benefits from descriptor mapping for object arrays such as `[Lio.orqueio.bpm...;`. Serialized proxy interfaces are resolved through the same mapping path so migrated proxy descriptors can be loaded when their Operaton interface equivalent exists.

For scripting, Operaton normalizes known fork package prefixes in script source before compile/eval when the flag is enabled. This covers Groovy and also supports string-based class lookups such as `Java.type('org.cibseven.bpm...')` without adding production compile-time dependencies on Groovy or GraalJS internals.

## Validation

Executed:

```bash
./mvnw -pl engine -Dskip.frontend.build=true -Dtest='ClassNameUtilTest,ReflectUtilTest,AbstractObjectValueSerializerTest,JavaObjectSerializerTest,OperatonScriptEngineManagerTest,ScriptTaskTest#testGroovyCanUseKnownForkClassNameWhenCompatibilityMappingIsEnabled' test
```

Result: 43 tests passed.

Coverage added for:

- package-prefix boundary mapping
- OrqueIO `io.orqueio.bpm` prefix mapping
- object array descriptor mapping
- disabled/enabled `ReflectUtil` behavior
- custom classloader fallback
- serializer retry behavior with original then mapped object type name
- Java serialization proxy interface mapping
- script engine context classloader mapping
- real Groovy ScriptTask using `org.cibseven.bpm.engine.delegate.BpmnError`

## Attribution

Inspired by and generalized from these CIBSeven namespace compatibility commits:

- https://github.com/cibseven/cibseven/commit/73b3da20f1556512113bd6afcedf011586a6ab51
- https://github.com/cibseven/cibseven/commit/379d7c71537fc2a958118a68baecbe61d909e518
- https://github.com/cibseven/cibseven/commit/91d7b8641dbcb18fd1c95d5c56faa8bfff4e0b3c

Original CIBSeven authors include:

- vladimirg <vladimir.gribko@cib.de>
- Oleg Skrypnyuk <oleg.skrypnyuk@cib.de>

Serialization edge-case coverage is also inspired by Fluxnova change 355:

- https://github.com/finos/fluxnova-bpm-platform/commit/06f6bb0f5ba9213ea05e9f37f12bdd84d185133c
- https://github.com/finos/fluxnova-bpm-platform/commit/c7d53ababb98a48b435cff82650c7c3928c8af5e

Original Fluxnova author:

- Bhandari, Prajwol <prajwol.bhandari@fmr.com>

This is not a straight CIBSeven or Fluxnova backport. The implementation is adapted for Operaton and generalized from fork-specific mappings to known fork engine prefixes -> `org.operaton.bpm`.

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/application/impl/ProcessApplicationScriptEnvironment.java
+++ b/engine/src/main/java/org/operaton/bpm/application/impl/ProcessApplicationScriptEnvironment.java
@@ -21,11 +21,11 @@ import java.util.List;
 import java.util.Map;
 
 import javax.script.ScriptEngine;
-import javax.script.ScriptEngineManager;
 
 import org.operaton.bpm.application.ProcessApplicationInterface;
 import org.operaton.bpm.engine.impl.scripting.ExecutableScript;
 import org.operaton.bpm.engine.impl.scripting.engine.DefaultScriptEngineResolver;
+import org.operaton.bpm.engine.impl.scripting.engine.OperatonScriptEngineManager;
 import org.operaton.bpm.engine.impl.scripting.engine.ScriptEngineResolver;
 
 /**
@@ -57,7 +57,8 @@ public class ProcessApplicationScriptEnvironment {
    */
   public synchronized ScriptEngine getScriptEngineForName(String scriptEngineName, boolean cache) {
     if(processApplicationScriptEngineResolver == null) {
-      processApplicationScriptEngineResolver = new DefaultScriptEngineResolver(new ScriptEngineManager(getProcessApplicationClassloader()));
+      processApplicationScriptEngineResolver = new DefaultScriptEngineResolver(
+          new OperatonScriptEngineManager(getProcessApplicationClassloader()));
     }
     return processApplicationScriptEngineResolver.getScriptEngine(scriptEngineName, cache);
   }
--- a/engine/src/main/java/org/operaton/bpm/engine/ProcessEngineConfiguration.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/ProcessEngineConfiguration.java
@@ -423,6 +423,11 @@ public abstract class ProcessEngineConfiguration {
    */
   protected boolean skipOutputMappingOnCanceledActivities;
 
+  /**
+   * Enables compatibility class name mapping from known Camunda 7 fork package prefixes to the Operaton package prefix.
+   */
+  protected boolean enableForkClassNameCompatibilityMapping;
+
   protected ProcessEngineConfiguration() {
   }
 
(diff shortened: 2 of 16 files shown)


kthoms:
We really need to be sure that we want this. It is potentially in danger of breaking, especially when the forks drift away more from each other.

hauptmedia:
We also started this alternative approach: https://github.com/operaton/operaton/pull/1424
My idea for the moment would be to let AI go through both PRs and propose a new solution out of it. This PR is a hacky solution for the problem which should be avoided to be integrated in the core (just think about it when this code is still active in 10 years!!). We also had no complaints from our community that they didn't manage to migrate their data/models to Operaton from C7.

hauptmedia:
I think the solution should be a clean data migration process, like we already proposed in #1424, which results in a clean state after it is finished. It's of course tricky because models with embedded scripts for example needs to be adjusted and migrated. It should be possible to do for known deployed models but the user has of course the burden to adapt their source models, too.

kthoms:
I would also prefer an external migration tool. Could this potentially create new versions of a process definition and then use process instance migration? But this kind of discussion should be done elsewhere outside of the scope of a PR.

I tend to close this PR as rejected.

hauptmedia:
After discussing the approach, I think we should not merge this PR in its current form. The concern is less about the implementation itself and more about the long-term architectural trade-off: a runtime compatibility layer for C7 and fork-specific class names would introduce implicit mapping behavior into the engine core. Even if it is opt-in, this is behavior we would need to maintain, explain, and keep correct over time. The risk grows as Operaton continue to diverge, because names that are compatible today may no longer represent equivalent behavior in the future.

Our preferred direction is a clean, explicit migration path that leaves installations in a valid state after completion, instead of keeping legacy or fork-specific references alive at runtime. This is especially important for deployed models, embedded scripts, serialized variables, and process application configuration, where hidden runtime rewriting can make problems harder to reason about later.

Given that we currently do not have strong community demand showing that users are blocked without this compatibility mode, I do not think the additional complexity in the engine core is justified.

I therefore close this PR as rejected / not planned. The work is still valuable as input for the broader migration discussion, especially around the edge cases it identifies, but I think that discussion should continue in the context of a dedicated migration approach such as #1424 or a follow-up design issue.

PR #3108, opened 2026-06-09
fix(engine-rest): whitelist identity verification endpoint

hauptmedia:
## Summary

Whitelists the REST identity verification endpoint in `ProcessEngineAuthenticationFilter`:

- `/identity/verify`
- `/engine/{engineName}/identity/verify`

This keeps the login/credential verification endpoint reachable without requiring the same REST authentication filter to authenticate the request first.

## Problem

`IdentityRestService#verifyUser` is the endpoint that checks submitted username/password credentials and returns an `AuthenticationResult`. If the global REST authentication filter treats that endpoint like a normal protected engine resource, the request can be challenged before the endpoint itself gets a chance to validate the submitted credentials.

CIBSeven fixed this as part of a larger REST filter configuration simplification. In Operaton, most of that larger diff is not applicable as-is:

- Operaton's assembly `web.xml` files already map the REST auth filter broadly with `/*`.
- Operaton Run registers the Jersey application URL mapping directly.
- The missing part is the narrow `identity/verify` whitelist and its path-matching coverage.

## Implementation Notes

The whitelist is intentionally exact:

- default-engine verification: `^/identity/verify$`
- named-engine verification: `^/engine/[^/]+/identity/verify$`

Other identity endpoints, such as `/identity/groups` or password-policy endpoints, continue to require the normal REST authentication filter behavior.

## Validation

Executed:

```bash
./mvnw -pl engine-rest/engine-rest -Dskip.frontend.build=true -Dtest=AuthenticationFilterPathMatchingTest test
```

Result:

- `AuthenticationFilterPathMatchingTest` ran 15 parameterized cases twice via the module's configured Surefire executions.
- No failures or errors.

## Attribution

Backported and narrowed from CIBSeven:

- https://github.com/cibseven/cibseven/commit/171f12e124a39c0944d72cea60f4706fad9c982a

Original author:

- Copilot <198982749+Copilot@users.noreply.github.com>

Original co-authors include:

- patrickCIB <204714531+patrickCIB@users.noreply.github.com>
- patrickCIB <patrick.fincke@cib.de>

### DIFF
--- a/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/security/auth/ProcessEngineAuthenticationFilter.java
+++ b/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/security/auth/ProcessEngineAuthenticationFilter.java
@@ -69,7 +69,9 @@ public class ProcessEngineAuthenticationFilter implements Filter {
 
   // regexes for urls that may be accessed unauthorized
   protected static final Pattern[] WHITE_LISTED_URL_PATTERNS = new Pattern[] {
-    Pattern.compile("^" + NamedProcessEngineRestServiceImpl.PATH + "/?")
+    Pattern.compile("^" + NamedProcessEngineRestServiceImpl.PATH + "/?"),
+    Pattern.compile("^" + NamedProcessEngineRestServiceImpl.PATH + "/[^/]+/identity/verify$"),
+    Pattern.compile("^/identity/verify$")
   };
 
   protected static final Pattern ENGINE_REQUEST_URL_PATTERN = Pattern.compile("^" + NamedProcessEngineRestServiceImpl.PATH + "/(.*?)(/|$)");
--- a/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/standalone/AuthenticationFilterPathMatchingTest.java
+++ b/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/standalone/AuthenticationFilterPathMatchingTest.java
@@ -109,6 +109,9 @@ public class AuthenticationFilterPathMatchingTest extends AbstractRestServiceTes
         {"", "/process-definition", "default", true},
         {"", "/engine", null, false},
         {"", "/engine/", null, false},
+        {"", "/identity/verify", null, false},
+        {"", "/engine/default/identity/verify", null, false},
+        {"", "/engine/someOther/identity/verify", null, false},
         {"", "/", "default", true},
         {"", "", "default", true},
         {"/someservlet", "/engine/someengine/process-definition", "someengine", true}


javahippie:
Thinking about whether this actually benefits us: as far as I can tell, nothing we ship uses /identity/verify — the webapps authenticate through their own UserAuthenticationResource, not this endpoint. So for any Operaton-delivered configuration, enabling it unblocks nothing. The only case it helps is an external consumer that enabled the REST auth filter and built a custom login on top of verify and even there it's largely redundant under HTTP Basic (you can just authenticate) and contradictory under SSO/token providers (where engine-password verification isn't what you'd use). Unless we know consumers actually rely on the verify-as-login pattern, the only real upside I see is staying in sync with the upstream backports. Happy to be corrected if there's a concrete use case I'm missing.

hauptmedia:
I dug a bit deeper into the CIBSeven history and Tim's concern.

The reason CIBSeven added the `identity/verify` whitelist was not primarily a new concrete webapp use case. It came from cibseven/cibseven#156, which simplified the engine-rest authentication filter configuration: instead of registering `ProcessEngineAuthenticationFilter` on a long list of explicit REST endpoint patterns (`/process-definition/*`, `/task/*`, `/identity/groups`, etc.), the filter was changed to match `/*` and rely on its internal whitelist for exclusions.

That changed the responsibility boundary. Before that PR, `/identity/verify` was effectively excluded by omission: it was not part of the explicit authenticated URL patterns. The old Tomcat/Wildfly web.xml snippets even had a commented note saying that `/identity/verify` should not be filtered so login remains possible. Once CIBSeven moved to a global `/*` mapping, that old exclusion had to move into `ProcessEngineAuthenticationFilter.WHITE_LISTED_URL_PATTERNS`; otherwise `/identity/verify` would be caught by the auth filter before it could verify the submitted credentials itself.

So in CIBSeven, this whitelist was the compensating change for the broader filter-mapping simplification.

For Operaton, the context is narrower:

- The delivered webapps do not use `/identity/verify` for login. They authenticate through `UserAuthenticationResource` at `/api/admin/auth/user/{engine}/login/{app}`, which also handles app authorization and session authentication.
- In the embedded webapp API, `/api/engine/engine/{engine}/identity/verify` would not become anonymously usable from this change alone, because the webapp `SecurityFilter` rules still protect `/api/engine/.*`; the generic `/api/engine/engine/{engine}/.*` allow rule goes through `EngineRequestAuthorizer` and requires an existing webapp authentication.
- The change can still matter for standalone engine-rest / Operaton Run / custom deployments where `ProcessEngineAuthenticationFilter` is enabled broadly and an external client intentionally uses `/identity/verify` as a credential-check endpoint.

That makes Tim's assessment mostly right: this does not unblock an Operaton-delivered webapp flow. The strongest reason to keep the PR is compatibility / preserving standalone engine-rest behavior for consumers that use the verify-as-login pattern. If we do keep it, I think the PR should be framed that way rather than as a fix for the shipped webapps.

kthoms:
Following your discussion I tend to reject this change. I have to admit, security is not my expert topic. Being in sync with forks with behavior Operaton has no benefit from is not justifying this.
We can start this discussion and remember when multiple integrators have this requirement.

javahippie:
Following our agreement to reject backport proposals if two or more committers agree, I'm closing this. See comments by me and @hauptmedia

PR #3109, opened 2026-06-09
fix(engine): use stored process definition key for MDC

hauptmedia:
## Summary

This backports a small Fluxnova engine fix for `ProcessDataContext`: when the MDC logging context needs the process definition key, the engine now reads the already stored execution property via `ExecutionEntity#getProcessDefinitionKey()` instead of resolving the full process definition and then calling `getKey()` on it.

## Problem

Before this change, `ProcessDataContext#pushSection` populated the configured `processDefinitionKey` MDC property through:

```java
execution.getProcessDefinition().getKey()
```

That makes MDC population depend on resolving/loading the process definition even though the execution already carries the process definition key directly. For logging-context setup this is unnecessary work and can cause avoidable process-definition access on a path that should only copy execution metadata into MDC.

## Fix

Use:

```java
execution.getProcessDefinitionKey()
```

This keeps the logging-context behavior the same from the caller's perspective, but avoids resolving the full process definition for the key.

## Test coverage

Added `ProcessDataContextTest#pushSectionUsesStoredProcessDefinitionKeyForMdc`.

The test is intentionally not self-referential: it configures only the `processDefinitionKey` MDC property, stubs `execution.getProcessDefinitionKey()` with a concrete value, asserts that the value is written to MDC, and verifies that `execution.getProcessDefinition()` is never called.

## Backport attribution

- Source commit: https://github.com/finos/fluxnova-bpm-platform/commit/705979e55288c506eb37e30408910d2ee9ee47ad
- Also seen as the same fix in Fluxnova commit: https://github.com/finos/fluxnova-bpm-platform/commit/9cc0f792c8901f91969c272f599721fa4219f68f
- Original author: Mannuru, ReddyKishore <ReddyKishore.Mannuru@fmr.com>

Adaptation notes:

- Namespace adapted from Fluxnova to Operaton.
- Added an Operaton-owned JUnit 5 unit test with AssertJ/Mockito.

## Verification

- `git diff --check`
- `./mvnw -pl engine -Dskip.frontend.build=true -Dtest=ProcessDataContextTest test`

### DIFF
--- a/engine/src/test/java/org/operaton/bpm/engine/impl/interceptor/ProcessDataContextTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/impl/interceptor/ProcessDataContextTest.java
@@ -22,7 +22,6 @@ import org.mockito.junit.jupiter.MockitoExtension;
 
 import org.operaton.bpm.engine.ProcessEngine;
 import org.operaton.bpm.engine.impl.cfg.ProcessEngineConfigurationImpl;
-import org.operaton.bpm.engine.impl.cfg.StandaloneInMemProcessEngineConfiguration;
 import org.operaton.bpm.engine.impl.persistence.entity.ExecutionEntity;
 import org.operaton.commons.logging.MdcAccess;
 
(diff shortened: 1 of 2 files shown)


engine/src/test/java/org/operaton/bpm/engine/impl/interceptor/ProcessDataContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+  private static final String PROCESS_DEFINITION_KEY = "invoice";
+
+  @AfterEach
+  void clearMdc() {
+    MdcAccess.remove(PROCESS_DEFINITION_KEY_MDC_PROPERTY);
+  }
+
+  @Test
+  void pushSectionUsesStoredProcessDefinitionKeyForMdc() {
+    ProcessEngineConfigurationImpl configuration = new StandaloneInMemProcessEngineConfiguration()
```
kthoms: Do we really need to build a real ProcessEngineConfigurationImpl with so many null initializations, or could `configuration` be a Mock
```
ProcessEngineConfigurationImpl configuration = mock(ProcessEngineConfigurationImpl.class);
doReturn(xxxxx).when(configuration).getXXX();
```

kthoms (approved):
LGTM

kthoms:
Cherry-picked for 2.1.2

PR #3110, opened 2026-06-09
fix(engine): handle form key evaluation failures

hauptmedia:
## Summary

This backports a Fluxnova engine fix for `TaskQuery#initializeFormKeys()`: if evaluating a task `formKey` expression fails, task retrieval continues and the task's form key is initialized as `null`.

## Problem

`TaskEntity#initializeFormKeyFromTaskDefinition` currently evaluates the BPMN user task form key expression directly:

```java
this.formKey = (String) taskDefFormKey.getValue(this);
```

That means a task query using `initializeFormKeys()` can fail while retrieving tasks if one task's form key expression references missing data, for example `${nonExistentVariable}`. This is especially awkward for task-list style queries: the task itself exists and is queryable, but asking the query to initialize form keys can make the whole retrieval fail.

## Fix

Wrap form key expression evaluation in `try/catch`:

- successful expressions keep the existing behavior and set the evaluated form key,
- failed expressions are logged as a warning,
- the task is still returned with `formKey == null`.

The change is intentionally limited to `formKey` expression initialization. Operaton Form Ref handling remains unchanged.

## Test coverage

Added a BPMN process with `operaton:formKey="${nonExistentVariable}"` and two focused `TaskQueryTest` cases:

- `testInitializeFormKeyWithInvalidExpressionReturnsNullForSingleTask`
- `testInitializeFormKeyWithInvalidExpressionDoesNotBlockTaskListRetrieval`

These tests are not self-referential: they deploy an actual BPMN user task whose `operaton:formKey` expression fails at runtime because the referenced variable is absent. The assertions prove that `initializeFormKeys()` returns the task(s) and that `getFormKey()` is initialized to `null` instead of propagating the expression evaluation exception.

Existing positive behavior was also checked with the current form-key/form-ref tests.

## Backport attribution

- Source commit: https://github.com/finos/fluxnova-bpm-platform/commit/0e8e2ec390e322d8c93bf4e1ef7224fb38f721eb
- Original author: Sowmya, HL <Sowmya.HL@fmr.com>

Adaptation notes:

- Namespace adapted from Fluxnova to Operaton.
- BPMN extension namespace adapted to `operaton:formKey`.
- Tests adapted to Operaton's JUnit 5 + AssertJ style.
- Added an Operaton header to the new BPMN test resource.

## Verification

- `git diff --check`
- `./mvnw -pl engine -Dskip.frontend.build=true -Dtest=TaskQueryTest#testInitializeFormKeyWithInvalidExpressionReturnsNullForSingleTask+testInitializeFormKeyWithInvalidExpressionDoesNotBlockTaskListRetrieval test`
- `./mvnw -pl engine -Dskip.frontend.build=true -Dtest=TaskQueryTest#testInitializeFormKeys+testInitializeFormKeysOperatonFormRef+testInitializeFormKeysForCaseInstance test`

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/db/EnginePersistenceLogger.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/db/EnginePersistenceLogger.java
@@ -807,6 +807,16 @@ public class EnginePersistenceLogger extends ProcessEngineLogger {
     );
   }
 
+  public void logFormKeyExpressionEvaluationException(String taskId, String expression, Exception cause) {
+    logWarn(
+      "111",
+      "Failed to evaluate form key expression '{}' for task '{}'. The form key will be set to null. Cause: {}",
+      expression,
+      taskId,
+      cause.getMessage()
+    );
+  }
+
   // exception code 110 is already taken. See requiredOperatonAdminOrPermissionException() for details.
 
   public static List<SQLException> findRelatedSqlExceptions(Throwable exception) {
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/persistence/entity/TaskEntity.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/persistence/entity/TaskEntity.java
@@ -1451,7 +1451,12 @@ public class TaskEntity extends AbstractVariableScope implements Task, DelegateT
   private void initializeFormKeyFromTaskDefinition(TaskDefinition taskDef) {
     Expression taskDefFormKey = taskDef.getFormKey();
     if (taskDefFormKey != null) {
-      this.formKey = (String) taskDefFormKey.getValue(this);
+      try {
+        this.formKey = (String) taskDefFormKey.getValue(this);
+      } catch (Exception e) {
+        LOG.logFormKeyExpressionEvaluationException(id, taskDefFormKey.getExpressionText(), e);
+        this.formKey = null;
+      }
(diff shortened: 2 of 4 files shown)


javahippie:
One concern about backward compatibility: this changes the observable behavior of initializeFormKeys(). Today a failing form-key expression propagates a ProcessEngineException  to the caller; after this change it's swallowed and the task is returned with formKey == null.

Anyone currently catching that exception to detect a broken expression (and e.g. fall back) would silently lose that hook. It also overloads null — it now means both "no form key" and "form-key evaluation failed," so callers can no longer distinguish the two from the API (only from a WARN log).

The intent is reasonable (one bad task shouldn't fail a whole list query), but I'd treat this as a behavioral change rather than a pure fix: ideally make it opt-in (strict vs. lenient), and at minimum call it out in the changelog so users relying on the current throw aren't surprised on upgrade.

hauptmedia:
I think we should reject this. It's breaking and you cannot evaluate between "no form key" and "form-key evaluation failed" anymore with this patch. I also think it's a better default behavior to fail so that you know for sure something is broken and not just "swallow" the problem

kthoms:
I understand that swallowing might be a behavior that users might want. Although as said, this would break others who rely on the existing behavior.
Positive is
a) additional logging by the engine
b) less need to handle these errors in client code

I would like to propose that this could be an opt-in behavior with a feature toggle parameter.

PR #3113, opened 2026-06-09
chore: clean up removed telemetry sending remnants

hauptmedia:
## Summary
- remove remaining telemetry sending configuration remnants after the sender was removed
- mark the legacy telemetry configuration REST endpoint as deprecated/no-op and document local diagnostics data accurately
- remove stale telemetry request timeout configuration and old test telemetry endpoint wiring

## Backport context
- Camunda 7 backport based on https://github.com/camunda/camunda-bpm-platform/issues/4483
- Original upstream PR: https://github.com/camunda/camunda-bpm-platform/pull/4486
- Upstream commit: https://github.com/camunda/camunda-bpm-platform/commit/aa1f15bddbf56b1ed9c79851c7d664beb5994565
- Backport inventory change unit: 4765

## Tests
- `git diff --check`
- `./mvnw -pl engine -DskipTests -Dskip.frontend.build=true -DskipChecks compile -q`
- `./mvnw -pl engine -Dtest=ManagementServiceTest#shouldAlwaysReturnFalseWhenFetchingIsTelemetryEnabled+shouldReturnFalseWhenToggleTelemetry,ManagementServiceGetTelemetryDataTest test -Dskip.frontend.build=true -q`
- `./mvnw -pl engine-rest/engine-rest -Dtest=TelemetryRestServiceTest test -Dskip.frontend.build=true -q`

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/cfg/ProcessEngineConfigurationImpl.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/cfg/ProcessEngineConfigurationImpl.java
@@ -956,7 +956,6 @@ public abstract class ProcessEngineConfigurationImpl extends ProcessEngineConfig
   // OLEs for foreign key constraint violations on databases that rollback on SQL exceptions, e.g. PostgreSQL
   protected boolean enableOptimisticLockingOnForeignKeyViolation = true;
 
-  protected int telemetryRequestTimeout;
   protected TelemetryDataImpl telemetryData;
   protected DiagnosticsRegistry diagnosticsRegistry;
   protected DiagnosticsCollector diagnosticsCollector;
@@ -5242,15 +5241,6 @@ public abstract class ProcessEngineConfigurationImpl extends ProcessEngineConfig
 
   public ProcessEngineConfigurationImpl setTelemetryData(TelemetryDataImpl telemetryData) {
     this.telemetryData = telemetryData;
-    return this;
-  }
-
-  public int getTelemetryRequestTimeout() {
-    return telemetryRequestTimeout;
-  }
-
-  public ProcessEngineConfigurationImpl setTelemetryRequestTimeout(int telemetryRequestTimeout) {
-    this.telemetryRequestTimeout = telemetryRequestTimeout;
     return this;
   }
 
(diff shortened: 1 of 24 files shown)


engine/src/main/resources/org/operaton/bpm/engine/db/liquibase/baseline/liquibase.db2.create.engine.sql
```diff
@@ -40,9 +40,6 @@ values ('history.cleanup.job.lock', '0', 1);
 insert into ACT_GE_PROPERTY
 values ('startup.lock', '0', 1);
 
-insert into ACT_GE_PROPERTY
```
kthoms, line 43: If this is removed, a migration should also remove it. This would require schema >7.24

hauptmedia: I think it doesn't hurt for instances that are being migrated if they still have this property because there is no code anymore that is evaluating it. It's only a matter of having a clean database when initializing a new database. I would not increase the database version for this.

hauptmedia: I just saw that this is actually not changed in the camunda repo, i will remove it here

hauptmedia:
we can perhaps also merge this PR https://github.com/operaton/operaton/pull/3151 when this is merged, which is partly related

engine/src/main/java/org/operaton/bpm/engine/impl/cfg/ProcessEngineConfigurationImpl.java
```diff
@@ -5245,15 +5244,6 @@ public ProcessEngineConfigurationImpl setTelemetryData(TelemetryDataImpl telemet
     return this;
   }
 
-  public int getTelemetryRequestTimeout() {
```
kthoms, line 5248: This removal would be a breaking change. We cannot do this now. These methods have to be annotated deprecated and marked for removal.

engine/src/main/java/org/operaton/bpm/engine/telemetry/Command.java
```diff
@@ -20,18 +20,18 @@
...
 /**
  * This class represents the data structure used for collecting information
- * about executed commands for telemetry data. A command is an action usually
+ * about executed commands for diagnostics data. A command is an action usually
  * triggered by a user and performed by the engine. This class counts the number
  * of executions per command.
  *
  * <p>
- * This information is sent to Operaton when telemetry is enabled.
+ * This information is exposed as local diagnostics data through the legacy
```
kthoms: Since telemetry is never enabled, can this paragraph be removed?
Same for other occurences.

javahippie (approved):
lgtm, the Sonar issues are just the deprecations (maybe we could adapt sonar here?)

PR #3114, opened 2026-06-09
feat(engine): add root process instance id to MDC

hauptmedia:
## Summary

Adds `rootProcessInstanceId` to the process-data MDC logging context so engine activity logs can be correlated across call-activity process hierarchies.

## Problem

Operaton already writes process data such as process instance id, process definition id, activity id, business key, tenant id, and engine name into the MDC while process execution logs are emitted. For executions that are part of a called process hierarchy, the local `processInstanceId` alone is not enough to correlate all related log lines back to the top-level process instance.

## Implementation Notes

This backport adds a configurable MDC property `loggingContextRootProcessInstanceId`, defaulting to `rootProcessInstanceId`, and writes `execution.getRootProcessInstanceId()` into the process-data logging context.

The Operaton adaptation also parks and restores externally provided MDC values for the new key, matching the existing behavior for the other process-data MDC properties. Setting the property to `null` disables it in the same way as the existing configurable MDC fields.

## Validation

Executed:

```bash
./mvnw -pl engine -Dskip.frontend.build=true -Dtest="ProcessDataLoggingContextTest#shouldLogRootProcessInstanceIdInActivityContext+shouldLogCustomRootProcessInstanceIdMdcProperty" test
```

Result: 2 tests passed.

Coverage added for:

- default `rootProcessInstanceId` MDC key
- custom MDC key via `setLoggingContextRootProcessInstanceId(...)`
- absence of the default key when a custom key is configured

## Attribution

- https://github.com/finos/fluxnova-bpm-platform/commit/241c3e1422390c62fcabb1ef4b99f2eb61efbee3
- https://github.com/finos/fluxnova-bpm-platform/commit/2442d26ecad8bc4860ab1f5b27ac7b686e41bd95

Original author:

- Nandan <Nandan.shenoy@fmr.com>

This is adapted for Operaton and is not a byte-for-byte cherry-pick.

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/cfg/ProcessEngineConfigurationImpl.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/cfg/ProcessEngineConfigurationImpl.java
@@ -947,6 +947,9 @@ public abstract class ProcessEngineConfigurationImpl extends ProcessEngineConfig
   protected String loggingContextProcessDefinitionId = "processDefinitionId";
   protected String loggingContextProcessDefinitionKey;// default == null => disabled by default
   protected String loggingContextProcessInstanceId = "processInstanceId";
+  /**
+   * @since 2.2
+   */
   protected String loggingContextRootProcessInstanceId = "rootProcessInstanceId";
   protected String loggingContextTenantId = "tenantId";
   protected String loggingContextEngineName = "engineName";
@@ -1743,6 +1746,7 @@ public abstract class ProcessEngineConfigurationImpl extends ProcessEngineConfig
         return MARIA_DB_PRODUCT_NAME;
       }
     } catch (SQLException ignore) {
+      // ignored
     }
 
     try {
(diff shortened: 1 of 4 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/cfg/ProcessEngineConfigurationImpl.java
```diff
@@ -5166,6 +5167,15 @@ public ProcessEngineConfigurationImpl setLoggingContextProcessInstanceId(String
     return this;
   }
 
+  public String getLoggingContextRootProcessInstanceId() {
```
kthoms, line 5178: New API => since tag

engine/src/test/java/org/operaton/bpm/engine/test/logging/ProcessDataLoggingContextTest.java
kthoms: Combin 2 AfterEach methods to one

engine/src/main/java/org/operaton/bpm/engine/impl/cfg/ProcessEngineConfigurationImpl.java
kthoms: done

engine/src/test/java/org/operaton/bpm/engine/test/logging/ProcessDataLoggingContextTest.java
kthoms: done

kthoms (approved):

PR #3125, opened 2026-06-09
feat(external-task-client): support pooled task execution

hauptmedia:
## Summary

This replaces the previous placeholder draft with an Operaton-adapted code backport of the EximeeBPMS external task client threading/statistics change.

The PR adds configurable pooled execution for external task handlers, optional handler-specific executors, per-executor acquisition runners, isolated backoff state per runner, and lightweight execution statistics logging.

This intentionally stays **Draft** because it changes public external-task-client API and concurrency behavior.

## What Changed

- Added `ExternalTaskClientBuilder.threadPoolSize(int)`, `maxFetchedTasksMultiplier(double)`, and `statsSchedulerEnabled(boolean)`.
- Added `ExternalTaskHandlerWithSpecificExecutor` so selected topic handlers can provide a dedicated `ThreadPoolExecutor`.
- Reworked `TopicSubscriptionManager` to route subscriptions to one `ExecutorRunner` per executor instead of executing handlers in a single acquisition loop.
- Added `ExecutorRunner`, which calculates fetch capacity from executor capacity minus active/queued work and dispatches fetched tasks asynchronously.
- Added `ExternalTaskExecutionStats` and periodic stats logging; the scheduler can be disabled through the builder.
- Added `BackoffStrategy.copy()` as a default method, with real copies for the built-in exponential strategies, so runners do not share mutable backoff state.
- Added an `EngineClient.fetchAndLock(..., maxTasks)` overload; the effective value is capped by the builder-level `maxTasks`.
- Skips already-expired task locks before invoking the handler and does not count those skipped tasks as execution stats.
- Replaced the old single-loop manager test with focused JUnit 5 coverage for executor routing, fetch capacity, backoff isolation, expired locks, and stats.

## Attribution

Backport source:

- Fork: EximeeBPMS
- PR: https://github.com/EximeeBPMS/eximeebpms/pull/115
- Commit: https://github.com/EximeeBPMS/eximeebpms/commit/4b8338305c898acc7793e910433057f01318d9cd
- Original author: Andrzej Dengusiak <andrzej.dengusiak@consdata.com>

Operaton adaptation notes:

- Adapted packages and names to `org.operaton`.
- Exposed the new builder options on the public `ExternalTaskClientBuilder` interface; the source fork only added them on the implementation.
- Kept existing custom `BackoffStrategy` implementations source-compatible by making `copy()` a default method. Stateful custom strategies should override it for full runner isolation.
- Did not import EximeeBPMS parent/internal dependency/build churn; Operaton already has the test dependencies needed here.
- Converted the test coverage to focused JUnit 5 tests instead of importing the large JUnit 4 source test set verbatim.

## Reviewer Notes

This is a behavior/API PR, not a small bugfix. The main review questions are:

- Is the public builder API (`threadPoolSize`, `maxFetchedTasksMultiplier`, `statsSchedulerEnabled`) the shape Operaton wants?
- Should handler-specific executors be part of the external-task-client API, or should executor selection be configured elsewhere?
- `stop()` stops acquisition runners but does not forcibly interrupt already queued/running handler jobs. That preserves normal handler execution but should be reviewed as part of client lifecycle semantics.
- The default builder executor allows core threads to time out after 60 seconds to avoid idle thread retention while keeping client restart semantics.
- The stats scheduler logs every five minutes on a daemon thread and is enabled by default, matching the source behavior. Review whether Operaton wants this default or a quieter opt-in.
- Custom backoff strategies that rely on mutable state should override `copy()`; otherwise the default returns `this` for backward compatibility.

## Verification

- `./mvnw -pl clients/java/client test -Dskip.frontend.build=true`

Result: client module tests passed (`27` tests, `0` failures/errors/skips).


javahippie (changes requested):
1. Undocumented default behavior change — handler execution moves from inline on the single acquisition thread to async dispatch on a ThreadPoolExecutor. Even at the default threadPoolSize=1, handlers now run on a different thread (breaking ThreadLocal/context-classloader/MDC assumptions), and with threadPoolSize>1 they run concurrently where they were previously sequential. Affects every existing user; needs a prominent migration note.
2. Breaking API change — threadPoolSize, maxFetchedTasksMultiplier, and statsSchedulerEnabled are added as abstract methods on the public ExternalTaskClientBuilder interface, breaking any third-party implementer/mocker. (The source fork kept these impl-only.) 
3. stop() can hang — ExecutorRunner.stop() does an unbounded thread.join(); if the acquisition thread is parked in the "all threads busy" CompletableFuture…join(), shutdown blocks until in-flight/queued handlers drain. manager.stop() stops runners sequentially, so one stuck runner blocks the rest and the stats scheduler.
4. Task executor is never shut down — client.stop() stops runners and the stats scheduler but never shuts down the builder/default ThreadPoolExecutor. Queued handler tasks still execute after stop() returns, and there is no close()/AutoCloseable for deterministic cleanup.
5. Silent regression for custom stateful BackoffStrategy — copy() defaults to returning this, so existing stateful strategies are shared across runners in multi-executor setups, reintroducing the mutable-state race with no compile-time signal.
6. Lost-runner-start race — subscribe() (computeIfAbsent → reads isRunning) racing manager.start() (CAS isRunning, then iterates runners) can leave a newly created runner never started, silently dropping that subscription.
7. Possible deadlock calling stop() from a handler on a small pool — handler holds the only worker, acquisition thread is parked waiting for a free slot, stop() joins the acquisition thread. Needs confirmation/guarding.

javahippie:
Forgot: This only touches the core client, not the Spring Boot starter. It won't break compilation, since the starter just calls the builder. But Spring workers silently inherit the new defaults: a stats thread logging every 5 minutes with no way to disable it, and handlers running on a different thread than before. Shutdown also gets riskier, since bean destruction calls client.stop(), which can hang on a stuck handler and leaves the executor running. Meanwhile none of the new options are exposed in application.yaml, so users get the downsides without the feature. The starter's properties and lifecycle need to be part of this PR, not a follow-up.

javahippie:
My suggestion is to reject this PR entirely. If we decide that we need a better external task worker with it's own thread handling for Java, we should create a new version that utizilizes all the shiny new concurrency features in the latest Java releases. I don't think it is possible to the change that was proposed here without changing the runtime behavior (from execution in the same thread to multi threaded) in a sane way

hauptmedia:
Thank you for your in-depth review, I also tend to reject this PR it but what i would like to do is to benchmark it so that we can use it as reference for future work

javahippie:
Rejected, see my comments

PR #3130, opened 2026-06-09
feat(engine): support transient BPMN IO parameters

hauptmedia:
## Summary

This PR backports the Fluxnova transient BPMN IO parameter support to Operaton.

It adds `operaton:isTransient="true"` handling for `operaton:inputParameter` and `operaton:outputParameter` mappings. When the attribute is present, the mapped value is set through `Variables.untypedValue(value, true)`, so the created variable is transient and is not persisted to runtime variable storage or history.

The BPMN model API is updated as well: `OperatonInputParameter` and `OperatonOutputParameter` now expose `isOperatonTransient()` / `setOperatonTransient(boolean)` for the same `operaton:isTransient` attribute.

## Change unit

- `361` [fluxnova:a1b817ba818e] add support for transient input and output parameters in BPMN model

## Attribution

Backported-adapted from Fluxnova commit `a1b817ba818ecf2bff618a134602a29e23704fab`.

Original author: Sowmya, HL <Sowmya.HL@fmr.com>

## Reviewer Notes

- Fluxnova parses the attribute from the Camunda namespace; this Operaton adaptation uses `BpmnParse.OPERATON_BPMN_EXTENSIONS_NS` and the `operaton:isTransient` model constant.
- The implementation keeps the existing non-transient IO mapping path unchanged. Only parameters explicitly marked with `operaton:isTransient="true"` are wrapped as transient typed values.
- `Variables.untypedValue(value, true)` preserves existing `TypedValue` instances by updating their transient flag rather than nesting them as plain Java objects.
- Runtime tests cover both directions:
  - transient input parameter: the service task receives a transient variable created by the BPMN input parameter.
  - transient output parameter: the following service task receives a transient variable created by the BPMN output parameter.
- Both runtime tests also assert that the mapped variable is not present in the persisted variable-instance query after execution reaches the wait state.
- Model API tests cover the new attribute default (`false`) and setter/getter behavior on input and output parameters.

## Verification

- `./mvnw -pl engine -Dtest=InputOutputTest#testTransientInputParameter+testTransientOutputParameter test -Dskip.frontend.build=true`
- `./mvnw -pl model-api/bpmn-model -Dtest=OperatonInputParameterTest,OperatonOutputParameterTest test -Dskip.frontend.build=true`

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParseUtil.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParseUtil.java
@@ -135,6 +135,8 @@ public final class BpmnParseUtil {
    */
   public static void parseInputParameterElement(Element inputParameterElement, IoMapping ioMapping) {
     String nameAttribute = inputParameterElement.attribute("name");
+    boolean isTransient = Boolean.parseBoolean(
+        inputParameterElement.attributeNS(BpmnParse.OPERATON_BPMN_EXTENSIONS_NS, "isTransient"));
     if(nameAttribute == null || nameAttribute.isEmpty()) {
       throw new BpmnParseException("Missing attribute 'name' for inputParameter", inputParameterElement);
     }
@@ -142,7 +144,7 @@ public final class BpmnParseUtil {
     ParameterValueProvider valueProvider = parseNestedParamValueProvider(inputParameterElement);
 
     // add parameter
-    ioMapping.addInputParameter(new InputParameter(nameAttribute, valueProvider));
+    ioMapping.addInputParameter(new InputParameter(nameAttribute, valueProvider, isTransient));
   }
 
   /**
@@ -154,6 +156,8 @@ public final class BpmnParseUtil {
    */
   public static void parseOutputParameterElement(Element outputParameterElement, IoMapping ioMapping) {
     String nameAttribute = outputParameterElement.attribute("name");
+    boolean isTransient = Boolean.parseBoolean(
+        outputParameterElement.attributeNS(BpmnParse.OPERATON_BPMN_EXTENSIONS_NS, "isTransient"));
     if(nameAttribute == null || nameAttribute.isEmpty()) {
       throw new BpmnParseException("Missing attribute 'name' for outputParameter", outputParameterElement);
     }
@@ -161,7 +165,7 @@ public final class BpmnParseUtil {
     ParameterValueProvider valueProvider = parseNestedParamValueProvider(outputParameterElement);
 
     // add parameter
-    ioMapping.addOutputParameter(new OutputParameter(nameAttribute, valueProvider));
+    ioMapping.addOutputParameter(new OutputParameter(nameAttribute, valueProvider, isTransient));
   }
 
   /**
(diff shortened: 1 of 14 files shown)


engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParseUtil.java
```diff
@@ -135,14 +135,16 @@ public static void parseOperatonOutputParameters(Element inputOutputElement, IoM
    */
   public static void parseInputParameterElement(Element inputParameterElement, IoMapping ioMapping) {
     String nameAttribute = inputParameterElement.attribute("name");
+    boolean isTransient = Boolean.parseBoolean(
+        inputParameterElement.attributeNS(BpmnParse.OPERATON_BPMN_EXTENSIONS_NS, "isTransient"));
```
javahippie, line 139: Should be called `transient`, not `isTransient` to match the parameters `asyncBefore`, `asyncAfter` and `exclusive`, none of them carries the `is`.

javahippie (changes requested):
I'm a bit concerned about what would happen if you define a transient input parameter on an activity that is marked as async before or an output parameter from an activity that is marked as asyncAfter, or is an event, or an external task. That behavior is hard to predict for users, and we would need to document this *really* well, ideally as a tooltip in the modeler itself. 

The go-to way to avoid passing large payloads currently is to only pass references, and then look up theses payloads. That would be more robust than to add property that whole classes of BPMN elements can never use, because they have wait states on their own.

engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/parser/BpmnParseUtil.java
hauptmedia: yes it we should decide to backport this feature it should be named "transient" and not "isTransient"

hauptmedia:
Large payloads are one use case, but another important one is avoiding persistence of sensitive data, for example secrets, in runtime or historical data.

That said, I agree with your concern: from a developer’s perspective it may not be obvious how long a transient variable actually lives, especially once wait states or async continuations are involved. That makes the feature easy to misunderstand unless it is documented and surfaced very clearly.

I’m still undecided whether this is something we should backport as-is.

a-m-zill:
Hello,
I’ve taken a look at the whole thing and thought about it at length. This feature is really more geared toward modelers, since transient variables can already be accessed directly by developers in the Operation API—for example, in the delegate. The same applies to listeners, which operate in a similar context.

You’ve already discussed the technical behavior behind this and recognize that it’s complex for a BPMN modeler, who may not even be deeply involved in automation themselves.

So from a developer’s perspective, I find it interesting but also a bit daunting. It’s just very complex—figuring out which data to use and how to handle it.

I also believe that this feature only really makes sense once there are dedicated modelers that support it. Are there any available yet, or does the Fluxnova Modeler already support this?

Should we perhaps propose this for discussion and a vote in the forum if we can’t decide?

hauptmedia:
Hi @a-m-zill , interesting thoughts that this is (mainly) a modeling feature. Could be used for configuration and secrets from the model to be accessed in the java code then. Question would be: is it in general a good idea to keep such data in a model? I would always think there should be some kind of secrets/config system from which you can access such data. It's also very simple to add something like this yourself in the engine. So I cannot still imagine use-cases where you exactly benefit using this transient variables mechanic

PR #3140, opened 2026-06-09
feat(engine): use random UUID v4 for strong IDs

hauptmedia:
## Summary

This PR backports the EximeeBPMS switch from time/MAC-based UUID v1 generation to random UUID v4 generation.

It changes `StrongUuidGenerator` to use `UUID.randomUUID()` and adapts tests that implicitly relied on UUID-v1 ordering by querying jobs/activity results explicitly or order-independently. The existing UUID concurrency test is also updated so it now exercises `StrongUuidGenerator` directly and asserts that generated IDs are UUID v4 values.

## Change unit

- `627` [eximeebpms:751bd7925cae] Replaced the current time-based UUID (v1) generation in StrongUuidGenerator with random UUID v4 (EximeeBPMS/eximeebpms#117)

## Attribution

Backported-adapted from EximeeBPMS commit `751bd7925cae3a4c5efa4598b59d113589810fcc`.

Original PR: https://github.com/EximeeBPMS/eximeebpms/pull/117

Original author: Robert Mastalerek <robert.mastalerek@gmail.com>

## Reviewer notes

This changes runtime ID semantics from time-based UUIDs to random UUIDs. The dependency cleanup for `com.fasterxml.uuid:java-uuid-generator` is intentionally left for explicit review, because the dependency is still referenced in several modules beyond the engine class changed here.

## Verification

- `./mvnw -pl engine -Dtest=UuidGeneratorTest,QueryByIdAfterTest test -Dskip.frontend.build=true`
- `./mvnw -pl test-utils/assert/core -Dtest=ProcessInstanceAssertJobTest test -Dskip.frontend.build=true`
- `./mvnw -pl webapps/assembly -Dtest=ProcessInstanceResourceTest test -Dskip.frontend.build=true`

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/persistence/StrongUuidGenerator.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/persistence/StrongUuidGenerator.java
@@ -16,36 +16,18 @@
  */
 package org.operaton.bpm.engine.impl.persistence;
 
-import com.fasterxml.uuid.EthernetAddress;
-import com.fasterxml.uuid.Generators;
-import com.fasterxml.uuid.impl.TimeBasedGenerator;
+import java.util.UUID;
 
 import org.operaton.bpm.engine.impl.cfg.IdGenerator;
 
 /**
- * {@link IdGenerator} implementation based on the current time and the ethernet
- * address of the machine it is running on.
- *
- * @author Daniel Meyer
+ * Secure random UUID (v4) based {@link IdGenerator}.
  */
 public class StrongUuidGenerator implements IdGenerator {
 
-  // different ProcessEngines on the same classloader share one generator.
-  protected static TimeBasedGenerator timeBasedGenerator;
-
-  public StrongUuidGenerator() {
-    ensureGeneratorInitialized();
-  }
-
-  protected synchronized void ensureGeneratorInitialized() {
-    if (timeBasedGenerator == null) {
-      timeBasedGenerator = Generators.timeBasedGenerator(EthernetAddress.fromInterface());
-    }
-  }
-
   @Override
   public String getNextId() {
-    return timeBasedGenerator.generate().toString();
+    return UUID.randomUUID().toString();
   }
 
 }
--- a/engine/src/test/java/org/operaton/bpm/engine/test/concurrency/UuidGeneratorTest.java
+++ b/engine/src/test/java/org/operaton/bpm/engine/test/concurrency/UuidGeneratorTest.java
@@ -18,12 +18,13 @@ package org.operaton.bpm.engine.test.concurrency;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentSkipListSet;
 
-import com.fasterxml.uuid.EthernetAddress;
-import com.fasterxml.uuid.Generators;
-import com.fasterxml.uuid.impl.TimeBasedGenerator;
 import org.junit.jupiter.api.Test;
+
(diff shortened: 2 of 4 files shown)


kthoms:
Combine the 2 AfterEach methods in ProcessInstanceResourceTest to a single one as we are on it to resolve the Sonar warning

javahippie:
I wonder what the reasoning for this change is?

UUID v1 is partially ordered, because it's time based. Using UUID v4 could hit the database performance, because they are completely random. I understand that this made the UUID (somewhat) predictable, but this change has performance considerations. They are now randomly inserted into the b tree, and writes into the index might become more expensive. This could increase the write times for history tables measureabily. As these are transactional, it might affect all writes.

If we really see the predictability of v1 UUIDs as an issue (id shouldn't be exposed / publicly accessible anyways), we should consider if v4 is the way to go, or other UUID schemas would match better. UUIDv7 seems to be a strong contender to solve both issues

hauptmedia:
I ran a focused PostgreSQL benchmark for the UUID v1 -> UUID v4 concern raised above.

Benchmark setup:

- PostgreSQL 17.10 via Testcontainers / Docker Desktop
- Operaton-like primary key shape: `ID_ varchar(64) not null primary key`
- Same table payload for both runs, same JDBC batch size (`1,000`)
- UUIDs were pre-generated outside the timed section, so the timing mostly measures the database insert / PK index behavior
- This is a synthetic benchmark that isolates the B-tree primary-key effect; it is not an end-to-end engine workload benchmark

Result with 250,000 inserted rows per strategy:

```text
ID pattern        duration   rows/s    table     pk index   leaf density   leaf fragmentation
UUID v1 ordered    1.894s   132,028   48.8 MiB   14.1 MiB        90.0%              0.0%
UUID v4 random     2.391s   104,550   48.8 MiB   18.5 MiB        68.9%             50.0%
```

Ratios from that run:

- UUID v4 insert duration was `1.26x` UUID v1
- UUID v4 primary-key index size was `1.31x` UUID v1

So the benchmark supports the concern: random UUID v4 inserts produced a visibly less dense and more fragmented PostgreSQL B-tree index, and insert throughput was lower in this isolated workload. The effect is measurable even though the table data size is identical.

Command used:

```bash
./mvnw -pl engine -Ppostgresql -Dtest=PostgreSqlUuidPrimaryKeyInsertBenchmarkTest \
  -Doperaton.uuidBenchmark.enabled=true -Doperaton.uuidBenchmark.rows=250000 \
  test -Dskip.frontend.build=true
```

I also verified that the benchmark is opt-in only; without `-Doperaton.uuidBenchmark.enabled=true` it is skipped.

hauptmedia:
Based on the PostgreSQL benchmark results, I would suggest rejecting this change. The switch from time-based UUID v1 to random UUID v4 introduces a measurable performance trade-off for primary-key inserts: in the isolated PostgreSQL benchmark, UUID v4 inserts were ~1.26x slower and produced a ~1.31x larger primary-key index with significantly higher B-tree fragmentation.

I also do not see a sufficiently strong security argument in the PR description to justify that trade-off. Predictable IDs alone should not grant access to protected resources; in correctly configured deployments, authorization checks should prevent unauthorized access.

If avoiding UUID v1 predictability is a goal, I think we should consider a time-ordered alternative such as UUIDv7 instead of moving directly to fully random UUIDv4.

javahippie:
Thanks for running the benchmarks, that's really helpful! I'd agree with your conclusions 100%

hauptmedia:
Reject, see remarks from https://github.com/operaton/operaton/pull/3140#issuecomment-4728031352

Issue #274, opened 2024-11-27
(webapps) Update Screenshots on landing page


kthoms:
**Describe the bug**
The Operaton landing page contains screenshots from Tasklist, Cockpit etc. which still show "Camunda".

**To Reproduce**
Start Operaton Tomcat distro.

**Expected behavior**
The screenshots should contain "Operaton" instead of "Camunda".

**Screenshots**
![Screenshot 320](https://github.com/user-attachments/assets/1961b6c6-e2d3-4424-a1ae-322582aa9323)

hauptmedia:
It seems like there is also a Camunda logo still in the upper left corner next to "Operaton" which must be removed.

AyalaFridman:
I would love to work on this, could you assign me this task?

javahippie:
@AyalaFridman thanks a lot, assigned it to you!

javahippie:
@AyalaFridman I unassigned you again. Feel free to reach out and pick up work in this issue, but I don't want it to appear blocked for others

javahippie:
The logo was not found during the first replacement beacuse it is SVG: "M13,0 C20.1797017,0 26,5.82029825..."

Pinging @lambdaschmied2 could you create a replacement for that? After that I would take care of the new screenshots

PR #3142, opened 2026-06-09
feat(engine-rest): add case-insensitive LIKE query option

hauptmedia:
## Summary

Backports CIB Seven's opt-in `likePatternIgnoreCase` support to Operaton's Task and User query APIs.

This adds the new boolean option to:

- `GET /task`
- `POST /task`
- `GET /user`
- `TaskQuery` and `UserQuery`
- stored Task filter query JSON serialization
- REST OpenAPI templates

The default is `false`, so existing query behavior is unchanged. When callers enable `likePatternIgnoreCase`, the SQL mappings use `UPPER(column) LIKE UPPER(pattern)` for LIKE fields that are case-sensitive by default.

## Reviewer Notes

This PR intentionally groups the Task and User changes because both CIB Seven commits introduce the same explicit API concept and the same opt-in SQL behavior.

Affected Task LIKE fields:

- `assigneeLike`
- `taskDefinitionKeyLike`
- `processInstanceBusinessKeyLike`
- `caseInstanceBusinessKeyLike`
- `processDefinitionNameLike`
- `caseDefinitionNameLike`
- `candidateGroupLike`

Affected User LIKE fields:

- `firstNameLike`
- `lastNameLike`
- `emailLike`

Task `nameLike`, `nameNotLike`, and `descriptionLike` are intentionally not changed here because Operaton already routes those through the existing case-insensitive task name/description handling. The new flag is additive and opt-in for the remaining LIKE comparisons.

The stored Task filter query path is covered as well: `JsonTaskQueryConverter` persists `likePatternIgnoreCase` only when enabled and restores it when filters are loaded or extended. This prevents saved filters from silently losing the new matching behavior.

Because this introduces `UPPER(...)` around selected columns, reviewers may want to consider DB/index implications for large installations. The behavior is only applied when the new option is explicitly set.

This PR does not include the separate CIB Seven case-insensitive identity/user-id query changes. Those change user-id semantics more broadly and remain a design topic outside this backport.

## Attribution

Source fork: CIB Seven

Backported and adapted from CIB Seven PRs:

- https://github.com/cibseven/cibseven/pull/266
- https://github.com/cibseven/cibseven/pull/274

Backported commits:

- `1daabccb83e597203f0eb38e819046f256379a6a` - `chore(rest): new option likePatternIgnoreCase for the TASK query to make LIKE operators case insensitive (#266)`
- `59f76b18d64e299a21937240ceaa66c7383c6e96` - `new(rest): add likePatternIgnoreCase parameter to User query (#274)`

Original authors:

- vladgrind `<156691737+vladgrind@users.noreply.github.com>`
- Vladimir Gribko `<156691737+vladgrind@users.noreply.github.com>`

Backport database:

- Change unit `1078` marked `pr_opened` for this PR.
- Change unit `1081` marked `pr_opened` for this PR.

## Verification

- `git diff --check`
- `./mvnw -pl engine -Dtest=TaskQueryTest,UserQueryTest,FilterTaskQueryTest test`
  - `Tests run: 353, Failures: 0, Errors: 0, Skipped: 4`
- `./mvnw -pl engine -DskipTests install`
- `./mvnw -f engine-rest/engine-rest/pom.xml -Dtest=TaskRestServiceQueryTest,UserRestServiceQueryTest test`
  - REST tests run twice by the module's standard/encoding Surefire executions.
  - Each execution: `Tests run: 98, Failures: 0, Errors: 0, Skipped: 0`
- Backport database: `PRAGMA integrity_check` returned `ok`.

### DIFF
--- a/engine-rest/engine-rest-openapi/src/main/templates/lib/commons/task-query-params.ftl
+++ b/engine-rest/engine-rest-openapi/src/main/templates/lib/commons/task-query-params.ftl
@@ -670,6 +670,16 @@
       desc = "Match all variable values in this query case-insensitively. If set
               `variableValue` and `variablevalue` are treated as equal." />
 
+  <@lib.parameter name = "likePatternIgnoreCase"
+      location = "query"
+      type = "boolean"
+      defaultValue = "false"
+      desc = "Makes LIKE comparisons case-insensitive for fields that are case-sensitive by default.
+              This affects: `assigneeLike`, `taskDefinitionKeyLike`, `processInstanceBusinessKeyLike`,
+              `caseInstanceBusinessKeyLike`, `processDefinitionNameLike`, `caseDefinitionNameLike`,
+              and `candidateGroupLike`. Note: `nameLike`, `nameNotLike`, and `descriptionLike`
+              are already case-insensitive by default." />
+
   <@lib.parameter name = "parentTaskId"
       location = "query"
       type = "string"
--- a/engine-rest/engine-rest-openapi/src/main/templates/lib/commons/user-query-params.ftl
+++ b/engine-rest/engine-rest-openapi/src/main/templates/lib/commons/user-query-params.ftl
@@ -62,5 +62,13 @@
     name = "potentialStarter"
     location = "query"
     type = "string"
-    last = last
     desc = "Only select Users that are potential starter for the given process definition."/>
+
+<@lib.parameter
+    name = "likePatternIgnoreCase"
+    location = "query"
+    type = "boolean"
+    defaultValue = "false"
+    last = last
+    desc = "Makes LIKE comparisons case-insensitive for fields that are case-sensitive by default.
+            This affects: `firstNameLike`, `lastNameLike`, and `emailLike`."/>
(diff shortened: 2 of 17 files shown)


javahippie:
I'm not a fan of this one. The like queries will be significantly slower, because the database has not indexed the affected columns for this usecase (including an UPPER function). Task Definitions, Assignees, Business Keys are all values that should be accessible in the correct case. Querying for them via "like" is not very helpful in my opinion, and I have never met anybody who wanted to do this with the engine.  If we decide that the benefit is there, we would either need to properly index the colums case sensitively, or accept that these queries could become really(!) slow and memory-heavy for big databases as they all will lead to full table scans. 

So I propose closing this, as it bloats the API and either introduces new DB indexes (and increase write time) for a function which will be used very, very rarely or decrease database performance during reads.

kthoms:
I second @javahippie arguments. Please close.

hauptmedia:
Rejected, please see Tims remarks https://github.com/operaton/operaton/pull/3142#issuecomment-4705932718

PR #3154, opened 2026-06-10
feat(engine-rest): add bulk job operations

hauptmedia:
## Summary

This draft PR backports the Fluxnova bulk job REST API and adapts it to Operaton's current REST stack.

It adds two job-level bulk endpoints:

- `POST /job/delete` deletes up to 200 jobs by id.
- `POST /job/suspended` activates or suspends up to 200 jobs by id.

Each endpoint processes the submitted job ids independently. If all operations succeed, the endpoint returns `200 OK` with a per-job result list. If at least one operation fails, the endpoint returns `207 Multi-Status` and includes success/failure details for each processed job. Empty, null, blank, or oversized input is rejected with `400 Bad Request`.

The OpenAPI templates are included for the new request and response DTOs. I also added the response schema for the delete endpoint in the Operaton adaptation, so both new endpoints document their per-job result payloads.

## Scope

This PR intentionally implements only the Bulk Job API from change unit 357.

The related Fluxnova Bulk Task API work is intentionally not included here. That part is much larger, introduces several task endpoints and partial-success semantics, and should stay as a separate API/product design review before code is proposed.

## Attribution

Backport of Fluxnova commit:

- `29cc3ec42dbf6a04ff97443ec10a149373b5356b` - `Bulk job api - delete, suspend & activate`

Original author:

- `jyotisahu9 <30759880+jyotisahu9@users.noreply.github.com>`

Operaton adaptation notes:

- Migrated package names to `org.operaton`.
- Used Jakarta REST imports and current Operaton REST test style.
- Added Operaton headers for newly created DTO/template files where the source files did not carry original copyright headers.
- Kept the implementation as a thin REST layer over existing single-job operations.
- Corrected the suspend/activate error log message so it describes suspension-state updates instead of deletion.
- Added OpenAPI response schemas for both new bulk result DTOs.

## Reviewer Notes

The implementation deduplicates job ids before processing, matching the set-based behavior in the source backport. This means duplicate ids in one request do not produce duplicate response entries.

The per-item failure handling deliberately catches exceptions around each individual job operation and continues processing the remaining ids. This is what makes the `207 Multi-Status` response meaningful for partial failures.

## Verification

- `./mvnw -pl engine-rest/engine-rest -Dtest=JobRestServiceInteractionTest test -Dskip.frontend.build=true`
- `./mvnw -pl engine-rest/engine-rest-openapi test -Dskip.frontend.build=true`

### DIFF
--- a/engine-rest/engine-rest-openapi/src/main/templates/models/org/operaton/bpm/engine/rest/dto/runtime/JobActivateSuspendDto.ftl
+++ b/engine-rest/engine-rest-openapi/src/main/templates/models/org/operaton/bpm/engine/rest/dto/runtime/JobActivateSuspendDto.ftl
@@ -1,0 +1,19 @@
+<#macro dto_macro docsUrl="">
+  <@lib.dto desc = "List of job ids and the requested suspension state.">
+
+    <@lib.property
+        name = "jobIds"
+        type = "array"
+        itemType = "string"
+        desc = "List of job ids."
+    />
+
+    <@lib.property
+        name = "suspended"
+        type = "boolean"
+        last = true
+        desc = "Supply `true` to suspend the jobs and `false` to activate them."
+    />
+
+  </@lib.dto>
+</#macro>
--- a/engine-rest/engine-rest-openapi/src/main/templates/models/org/operaton/bpm/engine/rest/dto/runtime/JobDeletionDto.ftl
+++ b/engine-rest/engine-rest-openapi/src/main/templates/models/org/operaton/bpm/engine/rest/dto/runtime/JobDeletionDto.ftl
@@ -1,0 +1,13 @@
+<#macro dto_macro docsUrl="">
+  <@lib.dto desc = "List of job ids to delete.">
+
+    <@lib.property
+        name = "jobIds"
+        type = "array"
+        itemType = "string"
+        last = true
+        desc = "List of job ids for deletion."
+    />
+
+  </@lib.dto>
+</#macro>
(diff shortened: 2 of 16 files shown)


hauptmedia:
I think the Operaton-native way to implement this should be via the batch system rather than synchronous per-job REST handling.

The current backport processes each job id directly and returns per-job success/failure responses, but Operaton already models larger asynchronous bulk operations through `BatchDto` and batch job handlers, for example job retries. For delete/suspend/activate jobs there does not seem to be an existing async batch API in `ManagementService`, so doing this properly would likely require adding the corresponding engine-level batch support first instead of only implementing it in the REST layer.

I am moving this PR back to draft for now so we can decide whether to keep the narrow Fluxnova backport semantics or rework this into a proper Operaton batch-based implementation.

javahippie:
I fail to imagine any usecase where I'd want to bulk delete jobs. If you bulk delete timers or async continuations, the related process instances are effectively broken? 

Adding a new HTTP verb on an existing REST path could be confusing, I think. Effectively, the new endpoint is just a loop over the logic of the old one. I agree with you @hauptmedia that thisshould be handled by the batch operation framework in Operaton, like [the retries endpoint](https://docs.operaton.org/reference/latest/rest-api/#tag/Job/operation/setJobRetriesAsyncOperation)

hauptmedia:
I have seen cases in Operaton where orphaned jobs were stuck and could not be cleaned up through the usual operational paths. For that kind of recovery scenario, a job delete operation can be genuinely useful.

At the same time, this is a very sharp tool. Deleting jobs can have severe side effects, for example removing timers or async continuations from running process instances. That may leave instances in an inconsistent operational state and can also undermine the audit trail, because the modeled process behavior no longer matches what actually happened.

So I can see a valid use case for this as an administrative recovery operation, but I would be careful about exposing it as a regular bulk API. My guess is that this risk is also one reason why the API historically has not encouraged job deletion beyond very controlled scenarios.

My overall thinking is that we should reject this feature because it does not use the batch system and could undermine the audit trail.

javahippie:
@hauptmedia 100% with you. I think your comment can be used as a "closing" comment on this PR to document the rationale perfectly, do you want to use that and close?

hauptmedia:
Rejected for now, see my remarks from https://github.com/operaton/operaton/pull/3154#issuecomment-4727566127

PR #3157, opened 2026-06-10
feat(engine-rest): expose deprecated camundaFormRef alias

hauptmedia:
## Summary

Backports the Fluxnova compatibility behavior for deprecated `camundaFormRef` response fields.

- Adds `camundaFormRef` as a deprecated alias for `operatonFormRef` in task/form REST DTOs.
- Includes the alias in HAL task output.
- Documents the deprecated alias in the OpenAPI templates and examples.
- Adds interaction-test assertions for start-form, task GET and task-form responses.

## Operaton adaptation

The source fork added a new Camunda-named Engine API type. This PR keeps the compatibility surface narrower for Operaton:

- No new `org.operaton.bpm.engine.form.CamundaFormRef` core API is introduced.
- The legacy object is represented as REST-local `CamundaFormRefDto`.
- `operatonFormRef` remains the canonical field and `camundaFormRef` is output-only/deprecated compatibility.

## Reviewer notes

- This intentionally reintroduces a Camunda-named REST response field for migration/client compatibility.
- The PR remains Draft because the source item came from the `needs_design` bucket and maintainers should explicitly confirm whether Operaton wants this compatibility contract.
- The legacy field mirrors `key`, `binding` and `version` from `operatonFormRef`; it is `null` whenever `operatonFormRef` is `null`.

## Attribution

Backported from Fluxnova commit `7311b89100e23c6bc4d0e90c7b8663341b693e29`.

Original author: `Mannuru, ReddyKishore <ReddyKishore.Mannuru@fmr.com>`

Tracked change unit: `465`.

## Verification

```bash
./mvnw -pl engine-rest/engine-rest -Dtest=TaskRestServiceInteractionTest,ProcessDefinitionRestServiceInteractionTest test -Dskip.frontend.build=true
```

Result: both Surefire executions passed, `Tests run: 387, Failures: 0, Errors: 0, Skipped: 0` per execution.

### DIFF
--- a/engine-rest/engine-rest-openapi/src/main/templates/models/org/operaton/bpm/engine/rest/dto/CamundaFormRef.ftl
+++ b/engine-rest/engine-rest-openapi/src/main/templates/models/org/operaton/bpm/engine/rest/dto/CamundaFormRef.ftl
@@ -1,0 +1,25 @@
+<#macro dto_macro docsUrl="">
+<@lib.dto>
+
+    <@lib.property
+        name = "key"
+        type = "string"
+        deprecated = true
+        desc = "The key of the Camunda Form. Deprecated compatibility alias for `operatonFormRef.key`." />
+
+    <@lib.property
+        name = "binding"
+        type = "string"
+        deprecated = true
+        desc = "The binding of the Camunda Form. Deprecated compatibility alias for `operatonFormRef.binding`." />
+
+    <@lib.property
+        name = "version"
+        type = "integer"
+        format = "int32"
+        deprecated = true
+        last = true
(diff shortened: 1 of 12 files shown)


hauptmedia:
This makes sense for me because it would be consistent to our other behavior where we are accepting old camunda namespaces. WDYT?

kthoms:
It is one thing that we allow to deploy models with Camunda namespace. It is a different thing what we answer on the API.
I am against such a feature.

hauptmedia:
You're right, we already migrated our own webapps to operatonFormRef. Would only make sense to be compatible with other frontends, who still use camundaFormRef. Let's reject this PR.

PR #3164, opened 2026-06-11
Fix mysql ignored test

### LINKED ISSUE #2526: Check CompetingMessageCorrelationTest#testConcurrentMixedCorrelationCase2
## What needs to be done?

The test should be checked and enabled, fixed or removed.

## Additional context

This relates to https://jira.camunda.com/browse/CAM-3636 which was fixed in 7.3.0.

We should enable the test again *and run it with MySQL, not H2* to verify that it works again. If not, we should try and make it work with all of our databases.

aounhaider1:
Fixes https://github.com/operaton/operaton/issues/2526
 The fix (DbSqlSession.java): Added deadlock detection at the top of isConcurrentModificationException using the existing ExceptionUtil.unwrapException + ExceptionUtil.checkDeadlockException utilities. When a
  deadlock is detected, return true → FAILED_CONCURRENT_MODIFICATION state → OptimisticLockingException thrown.


kthoms:
Thank you for providing the fix. Please remove the Draft mode once you are finished.

kthoms (approved):
LGTM; Thank you for providing this fix!

Issue #1010, opened 2025-07-24
Create Implementation Concept for AdHocSubProcess


ungerts:
**What needs to be done? Please describe.**

Create a detailed implementation concept for supporting the BPMN 2.0 `AdHocSubProcess` element in Operaton (pre-1.0 release). The concept should specify model structure, execution semantics, command interface, completion condition logic, and relevant API additions. The goal is to enable manual activation of subprocess activities and condition-based completion.

**Additional context**

While Operaton already supports core BPMN constructs and dynamic commands, it currently lacks support for BPMN 2.0’s `<adHocSubProcess>`. Adding this capability will enable:

- Dynamic, user-driven workflows.
- Better alignment with BPMN 2.0 spec compliance.
- Foundation for agentic/AI-assisted orchestration patterns.

hauptmedia:
Really great that you're bringing up this topic.

There are three open source implementations we can have a look at:

- Kogito - a cloud native BPM solution has support for Ad-Hoc processes, see 

https://blog.kie.org/2020/07/flexible-processes-in-kogito-part-1.html
https://github.com/apache/incubator-kie-kogito-runtimes

- Flowable has support for Ad-Hoc sub processes, see https://github.com/flowable/flowable-engine

- JBPM https://github.com/kiegroup/jbpm

hauptmedia:
I’m particularly excited about this feature! We’re currently building a custom case management solution, and our vision is to tightly bind BPM processes to individual cases. What Kogito refers to as "flexible processes" aligns exactly with what we’re aiming for.

Unfortunately, Kogito comes with other trade-offs that make it less suitable for our needs. That’s why bringing this capability into Operaton is something I’m genuinely enthusiastic about. Definitely looking forward to it!

ungerts:
@hauptmedia As you  said I think the interesting part in Kogito is the concept of an adhoc process at the root level ([link](https://blog.kie.org/2020/07/flexible-processes-in-kogito-part-1.html)). That’s something I’ve been looking into as well. Implementation-wise, Kogito is very different from Operaton. They always serialize the whole process instance using protobuf, so the engine behaves in a completely different way. I had to look into that in a previous project.

[...]

PR #3207, opened 2026-06-18
feat(engine): support for ad-hoc subprocesses

hauptmedia:
## Summary
- Backports CU 4872 from Fluxnova commit db14f6e71fba096408c560672cf96475c6c5516b.
- Adds BPMN model/parser support for ad-hoc subprocesses plus runtime commands to trigger activities and complete the subprocess.
- Adds REST DTOs/endpoints, OpenAPI templates, and engine/REST/integration test coverage adapted for Operaton namespaces and Jakarta REST.

## Verification
- ./mvnw -pl model-api/bpmn-model -DskipTests compile
- ./mvnw -pl model-api/bpmn-model -DskipTests install
- ./mvnw -pl engine -DskipTests test-compile
- ./mvnw -pl engine -DskipTests install
- ./mvnw -pl engine-rest/engine-rest -DskipTests test-compile
- ./mvnw -pl engine -Dtest=AdHocSubProcessTest test
- ./mvnw -pl engine-rest/engine-rest -Dtest=ExecutionRestServiceInteractionTest test
- git diff --check HEAD~1..HEAD
- rg check for Fluxnova/Finos leftovers and conflict markers in the backport files

Draft because this is a larger backport and should get focused review before marking ready.


hauptmedia:
## Review context: relation to #1010

This PR should be reviewed as a **Phase 1 / MVP implementation** for standard BPMN ad-hoc subprocess support, not as the full implementation concept discussed in #1010.

What aligns well with #1010:

- Adds BPMN model and parser support for `<adHocSubProcess>`.
- Adds engine behavior for entering an ad-hoc subprocess, manually starting contained activities, and completing the scope.
- Supports `completionCondition` and respects `cancelRemainingInstances` in the core completion path.
- Uses normal child executions, so standard BPMN propagation and boundary handling can work through the existing engine mechanisms.
- Does not attempt root-level ad-hoc processes, which matches the proposed step-1 direction from #1010.
- Adds Java RuntimeService operations and REST endpoints for triggering activities and completing the ad-hoc subprocess.

Important gaps compared to the #1010 concept:

- No discovery/introspection API yet. The concept proposed operations like `getAdHocScopes(processInstanceId)` and `getStartableActivities(scopeId)`. This PR only adds mutating operations.
- `ordering="Sequential"` is not enforced. The model API documents it as metadata only, while runtime activation is currently parallel-only.
- Startability is simplified to a type whitelist plus “no incoming sequence flow from inside the ad-hoc scope”. The #1010 concept expected BPMN-style startability, including activities with satisfied token conditions / `startQuantity`.
- No expression helpers are provided for completed/active/started/cancelled counts, startability checks, completion-candidate checks, or timing helpers.
- No scheduler SPI is included. The proposed `AdHocStartProposal` / `AdHocContext` mechanism remains future work.
- `activeTasksCollection` and `autoComplete` are Operaton extensions and should be documented explicitly as such.
- `completionCondition` is evaluated on subprocess entry as well as after child activity completion; #1010 described evaluation after each inner activity completes, so this semantic should be reviewed deliberately.
- Manual completion bypasses `completionCondition`; this may be acceptable, but should be an explicit API semantic.

Suggested review framing:

- Treat this PR as a pragmatic foundation for BPMN ad-hoc subprocesses.
- Do not close #1010 as fully implemented by this PR alone.
- Before marking ready, decide whether `ordering="Sequential"` should be enforced now or tracked as a follow-up.
- Strongly consider a follow-up for a non-mutating discovery API (`getAdHocScopes`, `getStartableActivities`) before exposing this as a complete user-facing feature, because UI/case-management/agent use cases otherwise need to know execution ids and activity ids upfront.
- Track expression helpers and scheduler SPI as separate follow-up work.

hauptmedia:
### Additional review context: comparison with Flowable ad-hoc subprocess support

I compared this PR against Flowable's current ad-hoc subprocess implementation (`flowable-engine` at `238a0d6d5c8876c6aa9644f864d29c638f504edd`) to validate the architecture choices and the remaining #1010 topics.

#### What Flowable does

Flowable exposes ad-hoc subprocess support as RuntimeService operations:

- `getAdhocSubProcessExecutions(processInstanceId)`
- `getEnabledActivitiesFromAdhocSubProcess(executionId)`
- `executeActivityInAdhocSubProcess(executionId, activityId)`
- `completeAdhocSubProcess(executionId)`

See Flowable's RuntimeService API and implementation:

- https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/RuntimeService.java#L1497-L1531
- https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/RuntimeServiceImpl.java#L828-L844

Flowable keeps `AdhocSubProcessActivityBehavior` itself very small: entering the subprocess marks the execution as scope and initializes data objects, but it does not auto-start children.

- https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/bpmn/behavior/AdhocSubProcessActivityBehavior.java#L35-L45

The runtime commands perform the key decisions:

- Discovery returns `FlowNode`s inside the ad-hoc subprocess that have no incoming flows. For sequential ordering it returns no enabled activities while any child execution exists.
  https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/GetEnabledActivitiesForAdhocSubProcessCmd.java#L49-L68
- Execution validates the same “no incoming flow” rule, creates a child execution, sets the requested node as current flow element, then continues the agenda.
  https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/ExecuteActivityForAdhocSubProcessCmd.java#L50-L78
- Explicit completion refuses to complete while child executions exist.
  https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/CompleteAdhocSubProcessCmd.java#L49-L64
- Completion condition handling is wired into the outgoing-sequence-flow agenda for activities whose parent container is an ad-hoc subprocess.
  https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/TakeOutgoingSequenceFlowsOperation.java#L147-L155
  https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/TakeOutgoingSequenceFlowsOperation.java#L293-L324

Their tests cover simple execution/discovery, sequential ordering, flows within the ad-hoc subprocess, completion conditions, parallel execution, and `cancelRemainingInstances=false`.

- https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/subprocess/adhoc/AdhocSubProcessTest.java#L41-L68
- https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/subprocess/adhoc/AdhocSubProcessTest.java#L199-L220
- https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/subprocess/adhoc/AdhocSubProcessTest.java#L304-L340
- https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/test/java/org/flowable/engine/test/bpmn/subprocess/adhoc/AdhocSubProcessTest.java#L345-L442

#### Comparison with this PR

This PR currently differs from Flowable in a few important ways:

1. **Startability is more conservative in Operaton.**  
   Flowable allows any `FlowNode` with no incoming flows. This PR currently allows task-like activities, call activities, subprocesses, and transactions only, also requiring no incoming transition and no already-active child execution. That is stricter than Flowable and should be treated as MVP scope, not as the final BPMN-complete interpretation.

   Current central point in this PR:
   https://github.com/operaton/operaton/blob/fd8f009efb157649bb3ec12c8434cc75be44340d/engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/behavior/AdHocStartability.java#L31-L145

2. **The new `AdHocStartability` refactor is a good architectural direction.**  
   Flowable duplicates the “enabled/startable” logic across the discovery command and execute command. With this PR, Operaton now has one central component that can be reused by trigger, future `getStartableActivities`, and later scheduler/startQuantity logic. This fits the #1010 direction better than copying Flowable literally.

3. **Sequential ordering is intentionally unsupported here.**  
   Flowable implements sequential ordering at runtime: discovery returns no enabled activities while one child exists, and execute throws if a child is already active. This PR rejects `ordering="Sequential"` at parse time. That is acceptable as an honest MVP boundary, but if #1010 later wants sequential support, the Flowable model suggests implementing the runtime check in the same central startability/support component rather than spreading it into each command.

4. **Initial activation is an Operaton/Fluxnova extension, not Flowable behavior.**  
   Flowable does not auto-start child activities on entering the ad-hoc subprocess. This PR supports `activeTasksCollection` for initial activation and validates it against startable activities. That is a product-specific extension and should stay clearly separated from generic BPMN ad-hoc semantics.

5. **Completion handling is different but reasonable for Operaton's PVM model.**  
   Flowable evaluates ad-hoc completion when an inner `FlowNode` leaves via agenda logic. This PR handles child completion through `CompositeActivityBehavior#concurrentChildExecutionEnded` and explicit completion through `CompleteAdHocSubProcessCmd`. The behavior is not a direct port, but it is consistent with the current Operaton architecture as long as we keep coverage for:
   - inner activities with outgoing flows inside the ad-hoc subprocess,
   - completion condition becoming true while downstream child executions may still exist,
   - `cancelRemainingInstances=true/false`,
   - explicit completion with and without active children.

   Relevant Operaton paths:
   https://github.com/operaton/operaton/blob/fd8f009efb157649bb3ec12c8434cc75be44340d/engine/src/main/java/org/operaton/bpm/engine/impl/bpmn/behavior/AdHocSubProcessActivityBehavior.java#L86-L162
   https://github.com/operaton/operaton/blob/fd8f009efb157649bb3ec12c8434cc75be44340d/engine/src/main/java/org/operaton/bpm/engine/impl/cmd/CompleteAdHocSubProcessCmd.java#L72-L101

#### Recommendation for this PR

I would not copy Flowable's implementation directly. The current PR should keep the conservative MVP behavior, but the review context should explicitly record that Flowable is broader and API-driven. The important architectural action item has already been addressed by the latest refactor: keep all “is this ad-hoc activity startable/enabled now?” decisions behind `AdHocStartability`, so #1010 follow-ups such as discovery API, scheduler integration, sequential ordering, and startQuantity/token semantics can evolve from one shared decision point.

Before marking this ready, I would keep the remaining open topics as follow-up scope unless we decide to expand this PR beyond the MVP:

- public discovery API equivalent to Flowable's `getEnabledActivitiesFromAdhocSubProcess`, backed by `AdHocStartability#getStartableActivities`,
- broader FlowNode support versus the current whitelist,
- real runtime support for `ordering="Sequential"`, if we want it instead of parse-time rejection,
- scheduler/startQuantity semantics,
- additional flow-through completion tests inspired by Flowable's `testFlowsInAdhocSubProcess` and `testParallelFlowsWithKeepRemainingInstancesAdhocSubProcess`.

hauptmedia:
### Additional review context: BPMN 2.0.2 conformance vs. Flowable

I also checked the PR against the official BPMN 2.0.2 specification, especially the ad-hoc subprocess descriptions in sections 10.3.5 and 13.3.5.

Specification reference:

- https://www.omg.org/spec/BPMN/2.0.2/PDF

#### What BPMN 2.0.2 requires conceptually

For an ad-hoc subprocess, BPMN 2.0.2 describes a flexible execution model rather than a normal start-to-end subprocess flow:

- An ad-hoc subprocess contains embedded inner Activities. Sequence Flows, Gateways and Intermediate Events may be used to express dependencies. Start Events and End Events are not allowed inside the ad-hoc subprocess.
- Initially, activities without incoming Sequence Flows are enabled.
- Activities may be executed multiple times. With `ordering="Parallel"`, the same inner Activity can even have multiple parallel instances.
- With `ordering="Sequential"`, only one inner Activity may execute at a time.
- After each inner Activity completion, `completionCondition` is evaluated.
- If the completion condition becomes true and `cancelRemainingInstances="true"`, remaining running inner activities are canceled. If `cancelRemainingInstances="false"`, the subprocess completes only after remaining inner instances have completed or terminated.
- If inner Activities have outgoing Sequence Flows, the standard describes token propagation. `completionQuantity` controls how many tokens an Activity produces and `startQuantity` controls when downstream Activities become enabled.

#### Current PR vs. BPMN 2.0.2

This PR implements a useful MVP, but it should not be described as full BPMN 2.0.2 ad-hoc subprocess conformance yet.

Implemented or partially implemented:

- Model support for `adHocSubProcess`, `ordering`, `cancelRemainingInstances`, and `completionCondition`.
- Runtime triggering of selected ad-hoc activities.
- Completion condition evaluation after inner activity completion.
- `cancelRemainingInstances` behavior.
- Initial/triggerable activity filtering based on “no incoming Sequence Flow”.
- Conservative parse-time rejection of `ordering="Sequential"`, which is honest but not standard-complete.

Important gaps versus BPMN 2.0.2:

- `ordering="Sequential"` is standard BPMN but currently rejected instead of executed with runtime sequencing.
- Parallel mode should allow multiple parallel instances of the same inner Activity. The current `already active` check prevents that.
- The standard’s enabled-set semantics are token-based once Sequence Flows are present. The PR currently treats activities with incoming Sequence Flows as non-startable entry points, but it does not model `completionQuantity`, token propagation, or `startQuantity` for later enabling.
- There is no public discovery API yet for the current enabled activities.
- The current startability whitelist is narrower than the standard’s “Activities” wording. That is acceptable as an MVP boundary, but it is stricter than BPMN itself.
- We should verify/cover validation for elements that BPMN excludes from ad-hoc subprocesses, especially Start Events and End Events inside the ad-hoc scope.

#### Flowable comparison

Flowable is closer to the BPMN 2.0.2 operational model in a few places:

- It exposes discovery and execution APIs for ad-hoc subprocesses:
  - https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/RuntimeService.java#L1497-L1531
- It implements `ordering="Sequential"` at runtime by returning no enabled activities / throwing while a child execution exists:
  - https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/GetEnabledActivitiesForAdhocSubProcessCmd.java#L55-L68
  - https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/cmd/ExecuteActivityForAdhocSubProcessCmd.java#L56-L78
- It handles ad-hoc completion after outgoing Sequence Flow handling in agenda logic:
  - https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/TakeOutgoingSequenceFlowsOperation.java#L147-L155
  - https://github.com/flowable/flowable-engine/blob/238a0d6d5c8876c6aa9644f864d29c638f504edd/modules/flowable-engine/src/main/java/org/flowable/engine/impl/agenda/TakeOutgoingSequenceFlowsOperation.java#L293-L324

But Flowable also does not appear to implement the full BPMN token semantics for discovery. Its enabled activity command mainly returns `FlowNode`s without incoming flows, and it does not appear to compute a complete token-based enabled set using `startQuantity` / `completionQuantity`.

#### Recommendation

I would keep this PR framed as an MVP / first runtime support slice, not as full BPMN 2.0.2 ad-hoc subprocess support.

For #1010 follow-ups, the standard-driven roadmap should be:

1. Add a public discovery API backed by `AdHocStartability`.
2. Decide whether the current whitelist should expand toward standard “Activities”.
3. Implement runtime `ordering="Sequential"` instead of rejecting it.
4. Allow multiple parallel instances of the same inner Activity where BPMN permits it.
5. Add token/startQuantity/completionQuantity semantics for activities connected by Sequence Flows.
6. Add explicit validation/tests for BPMN-prohibited elements inside ad-hoc subprocesses, especially Start Events and End Events.

The latest `AdHocStartability` refactor is therefore still the right direction: it gives us a single place to evolve the MVP rules into the standard-enabled-set rules later.

hauptmedia:
### Additional review context: jBPM ad-hoc subprocess implementation

I also compared this PR with the ad-hoc subprocess implementation in `kiegroup/jbpm` (`main` at `d518cacb3957a203a906a8ae1525aec12728a0d2`). jBPM is useful as a third reference because it does **not** follow the same shape as Flowable. Instead of a direct `getEnabledActivities` / `executeActivity` API, jBPM maps BPMN ad-hoc subprocesses to its internal `DynamicNode` runtime model and drives activation through signals and Drools rule-flow groups.

Reference repository:

- https://github.com/kiegroup/jbpm/tree/d518cacb3957a203a906a8ae1525aec12728a0d2

#### How jBPM represents ad-hoc subprocesses

The BPMN parser maps `<adHocSubProcess>` to `DynamicNode`:

- `AdHocSubProcessHandler#createNode` creates a `DynamicNode` and a variable scope.
- `handleNode` reads `cancelRemainingInstances`, defaults ad-hoc subprocesses to `autoComplete=false`, reads a custom activation expression, and parses `completionCondition` into either jBPM's special `autocomplete` mode or a completion expression.
- The dumper writes `ordering="Parallel"` unconditionally.

Sources:

- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/AdHocSubProcessHandler.java#L37-L95
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/AdHocSubProcessHandler.java#L98-L117

`DynamicNode` itself accepts events for nodes without incoming connections. That mirrors the BPMN “initially enabled = no incoming Sequence Flow” idea, but it is signal/event based rather than exposed as a discovery list.

- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/DynamicNode.java#L41-L70
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/DynamicNode.java#L104-L120

The runtime behavior is implemented in `DynamicNodeInstance`:

- entering the node activates the rule-flow group,
- auto-start nodes can be triggered from metadata,
- signal events with a matching inner node name trigger that node if it has no incoming connections,
- completion is evaluated after child completion via terminate end events, auto-complete/no active children, or a MVEL completion expression.

Sources:

- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java#L63-L97
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java#L119-L145
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java#L154-L199

Completion/cancel behavior reuses composite node completion support:

- `CompositeNode` defaults `cancelRemainingInstances=true`.
- `CompositeNodeInstance#triggerCompleted` cancels remaining child instances if that flag is true.

Sources:

- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/CompositeNode.java#L44-L47
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/core/node/CompositeNode.java#L369-L383
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java#L168-L178

jBPM also supports dynamically adding work items and subprocesses into a running dynamic/ad-hoc context, which is a different capability from both Flowable and this PR:

- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicUtils.java#L60-L177
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicUtils.java#L189-L300

#### jBPM tests worth noting

jBPM has tests for:

- signaling named inner activities,
- auto-completion after active children finish,
- MVEL completion expressions,
- dynamic work item insertion,
- dynamic subprocess insertion,
- terminate-end-event behavior.

Sources:

- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/subprocess/AdHocSubProcessTest.java#L76-L105
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/subprocess/AdHocSubProcessTest.java#L139-L173
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java#L996-L1028
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java#L1030-L1128
- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java#L1856-L1890

One caveat: some jBPM test models include an End Event inside an ad-hoc subprocess, which BPMN 2.0.2 says should not be used inside ad-hoc subprocesses. Example:

- https://github.com/kiegroup/jbpm/blob/d518cacb3957a203a906a8ae1525aec12728a0d2/jbpm-test-coverage/src/test/resources/org/jbpm/test/functional/subprocess/AdHocSubProcess.bpmn#L6-L12

#### Comparison with this PR and Flowable

Compared with this PR:

- jBPM is more dynamic: it can activate existing inner nodes by signal name and can dynamically add work items/subprocesses to a running ad-hoc context.
- This PR is more explicit/API-oriented for Operaton: it adds `triggerAdHocActivities` and explicit completion commands, but only for existing modeled activities.
- jBPM and this PR both use “no incoming connections/flows” as the first startability boundary.
- This PR has the new `AdHocStartability` centralization, which jBPM does not have as a separate discoverable support component.
- jBPM does not appear to provide a Flowable-like public “list currently enabled activities” API.

Compared with Flowable:

- Flowable is closer to a simple BPMN runtime API: `getEnabledActivitiesFromAdhocSubProcess`, `executeActivityInAdhocSubProcess`, `completeAdhocSubProcess`.
- jBPM is closer to a rule/signal-driven dynamic work context.
- Flowable implements runtime sequential gating; jBPM's parser/schema recognizes `Sequential`, but the inspected runtime path does not show an equivalent sequential scheduling rule, and the dumper always emits `ordering="Parallel"`.

Compared with BPMN 2.0.2:

- jBPM supports flexible, performer/signal-driven selection of inner activities and dynamic insertion, which aligns well with the ad-hoc spirit.
- jBPM does not appear to implement full token-based enabled-set semantics with `startQuantity` / `completionQuantity` for ad-hoc subprocesses.
- jBPM's acceptance/test usage of End Events inside ad-hoc subprocesses is not BPMN-conformant.
- Sequential ordering does not appear complete from the inspected code path.

#### Recommendation for Operaton

jBPM reinforces the same conclusion as Flowable and the BPMN spec review: this PR should stay framed as an MVP slice, not full BPMN 2.0.2 ad-hoc subprocess conformance.

For #1010 follow-ups, jBPM adds two useful ideas to the roadmap:

1. Consider an activation/discovery model that can be used both by explicit API calls and by event/scheduler-style activation.
2. Keep the door open for dynamic ad-hoc work item/subprocess insertion, even if that is outside the current PR scope.

The current `AdHocStartability` component is still the right anchor point for Operaton. It can evolve from today’s conservative “existing modeled starter activities” rule into a broader standard-aware enabled-set component without copying either Flowable or jBPM directly.

Issue #1559, opened 2025-11-12
Version tag filtering for event subscriptions


marijnritense:
**Is your feature request related to a problem? Please describe.**
For both process definitions and decision definitions, we use version tags to bundle them together. Everything with the same version tag belongs to the same bundle, which in our own domain we refer to as a "case definition". 

When we use message correlation, we often want to correlate these to process definitions and process instances with the same version tag. Operaton currently keeps track of event subscriptions for the latest version of a process definition. When a process definition with the same key is saved, the event subscriptions for the old version are removed, and the event subscriptions for the new version are stored. 
If we have three different versions of this process definition, and they all have their own version tag, there is no way for us to correlate messages to the correct process definition, as (like mentioned before) only the event subscriptions for the latest version are stored. Additionally, if someone makes a process definition with the same key as an already existing process definition, but for a different use-case (e.g. because it is relevant for a different case definition) this can result in process definitions that are not accessible via message correlation at all, leaving those case definitions in a broken state.

**Describe the solution you'd like**
Functionally:
- Message correlation can be done based on version tag, where the current behavior (only correlating to the latest) is the default.
- Event subscriptions for older versions of a process definition are not removed (or this is behavior that can be toggled)

From a technical perspective:
- If a process definition already has subscriptions in `act_ru_event_subscr`, do not delete these.
- Add `version_tag_` column to `act_ru_event_subscr` table that holds the version tag of the corresponding process definition.
- Add support in the message correlation builder (and related classes) to provide this version tag as a way to further filter the subscriptions (not providing a version tag would result in the current behavior of correlating the event to the latest subscription with that event name)

**Additional context**

We considered solving it outside of Operaton, but replacing the existing message correlation with our own implementation would mean significantly deviating from what Operaton provides. This brings a risk with future version upgrades.

kthoms:
Thank you @marijnritense for this detailed description of your problem. We understand that this is a real issue and would like to work on a fix. Or could you imagine to work on one? Note that you are proposing a change affecting the DB schema. The first change of that kind would require the initialization of a database version to `7.25` using the script `.devenv/scripts/maintenance/init-database-version.py`

javahippie:
Hi @marijnritense 

thanks for bringing this up and taking the time for the writeup! Direct feedback and feature requests out of daily use are really important for us. Please find some questions for clarification below.

[...]

PR #3275, opened 2026-07-01
Create ADR: Version tag filtering for start Events

javahippie:
This ADR addresses issue #1559 

Draft version: 
https://github.com/operaton/operaton/blob/feature/version-tag-filtering-for-events/docs/decisions/0002-version-tag-filtering-for-events.md

### DIFF
--- a/docs/decisions/0002-version-tag-filtering-for-events.md
+++ b/docs/decisions/0002-version-tag-filtering-for-events.md
@@ -68,27 +68,6 @@ an earlier version that carries the **same** tag, and only retains earlier versi
 carry a **different**, non-empty tag. Correlation without a version tag continues to resolve
 to the latest version.
 
-### Consequences
-
-- Good, because messages can be routed to a specific, non-latest version selected by its
-  version tag, enabling parallel, tag-bundled versions.
-- Good, because the version tag lives directly on the subscription row, so correlation
-  filtering needs no additional join and can be indexed.
-- Good, because it reuses the established version-tag model and the existing correlation
-  code path rather than introducing a new resolution mechanism.
-- Bad, because it requires a database schema change (a new column on `ACT_RU_EVENT_SUBSCR`)
-  with create scripts and upgrade scripts for all supported database vendors.
-- Bad, because retaining subscriptions for multiple versions changes long-standing
-  invariants (e.g. the "one message start subscription per name and tenant" guard) that must
-  be relaxed carefully to avoid regressions.
-- Bad, because the deploy-time and undeploy-time subscription lifecycle must be kept
-  symmetric: undeployment has to restore the same-tag predecessor without duplicating an
-  already-retained subscription, adding complexity to the existing restore logic in
-  `DeleteProcessDefinitionsByIdsCmd`.
-- Neutral, because the version tag is denormalised onto the subscription row and must be
-  kept consistent with the process definition; because subscriptions are immutable for a
-  deployed version, the value is written once at subscription creation.
-
 ### Confirmation
 
 - Integration tests that deploy multiple tagged versions of the same process, assert that
@@ -120,6 +99,10 @@ process definition's version tag on the subscription row, and filter correlation
   filtering requires no join and can be indexed.
 - Good, because it matches the approach proposed in the originating issue and keeps the
   correlation query close to its current form.
+- Good, because it follows the current practices in most tables (duplicating column values
+  for read performance)
+- Good, because it enables us to add a constraint to verify only one process definition is
+  deployed with unique version tags
 - Neutral, because the tag is denormalised and must be written consistently at subscription
   creation.
 - Bad, because it requires a new column with create and upgrade scripts for all supported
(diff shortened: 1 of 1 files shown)


docs/decisions/0002-version-tag-filtering-for-events.md
```diff
@@ -0,0 +1,265 @@
+---
+status: "Proposed"
+date: 2026-07-01
+decision-makers: ["@javahippie"]
```
kthoms: add me

docs/decisions/0002-version-tag-filtering-for-events.md
```diff
@@ -0,0 +1,265 @@
...
+consulted: []
+informed: []
```
kthoms: add martin and all you pinged on slack

kthoms:
In general already a very strong proposal.

Not covered is the configurability of removal of old subscriptions. (_Event subscriptions for older versions of a process definition are not removed (or this is behavior that can be toggled_)

Questions:
Could the sender know the process definition id of the version that should be correlated? The API would already allow that. I assume that only the version tag is known.

How should `correlateAll()` behave? Would it also relate other versions or only the latest?

kthoms:
I would like to challenge if Option 2 is indeed the inferior choice. It is said that queries and filtering becomes more complex. I'd consider a simple join not as extraordinary more complex (even using mybatis) to the cost of a required db schema change.
Performance wise (index) this would need to be measured. My feeling is that the additional join should not be that expensive.

javahippie:
Thanks for the comments, @kthoms 

I'm not quite sure I understand your question about a toggle for removing older process instances. This is currently the default and if we added tagged versions, we would keep this behavior for now: If a process definition has no tag, the old one will be removed. If the definition has a tag, an older one with the same tag will be removed. If we don't remove these, we will have two subscriptions with the same correlation attributes. How should the toggle behave in this setting? If this is a runtime toggle, what happens if you keep the latest versions, change the toggle and restart the application? We would need to check the setting change on startup and create the correct state in the DB, that sounds challenging

The sender could theoretically look up the latest process definition version for a specific version tag upfront and correlate on the process instance ID. If the correlation comes from "outside", e.g. per REST API, this assumes that the caller has authorization to call the other endpoints, too. Using Process Definition IDs and Process Instance IDs outside of the engine is not generally advised. 

I would keep the behavior of correlateAll the same as before: It finds all valid subscriptions. In the past, there was only one, in the future there would be only one per definition and version tag. Having mutiple Process Definitions callable by the same version tag sounds like something that happens more by mistake than by purpose, I cannot think of a Usecase.

I think you are right to challenge performance concerns, it's a little early to fixate on them. I tried to follow the established behavior in the tables, and duplication seems to have been the favored solution for Camunda. Even more bonus points for the joins: We would not need to care about migrating existing subscriptions, because the version tags are already in the joined table, even retroactively. The only Drawback I see is that we cannot make sure that no "duplicate subscriptions" happen with a unique index, if we look up the tag via join. That might be a minor detail, and I'm still torn after your suggestion for the join. On the other hand, we have an open bug for duplicate process definitions due to a missing unique index.

docs/decisions/0002-version-tag-filtering-for-events.md
```diff
@@ -0,0 +1,265 @@
...
+
+To keep the current behaviour intact, previous versions do not accumulate subscriptions
+without bound. The retained set is **at most one active start subscription per
+`(key, tenant, version tag)` — the latest version bearing that tag — plus the global latest
+version**. Deploying a new version therefore still supersedes (deletes) the subscription of
+an earlier version that carries the **same** tag, and only retains earlier versions that
+carry a **different**, non-empty tag. Correlation without a version tag continues to resolve
+to the latest version.
+
+### Consequences
```
javahippie: Mostly duplicated, remove

kthoms:
Regarding the potential toggle, I was assuming this would be a configuration level option, thus on start. I was just wondering if a user would like to have influence on the behavior, but I'm lacking a use case.

Relating on the process def id to use existing API we can forget. You are right, when users call via REST, they might have restricted access and maybe not know the id to use this.

correlateAll: Fine for me, would be also my interpretation. Maybe worth to add a comment in the method's documentation then.

Regarding Join: You know the consequences better than me. My point is just: Why changing the schema when the information is already there at (propably) no additional runtime cost? Following the Camunda style is valid, but not obligatory. Please add "Camunda style" as pro.
A decision does not be taken now, just stating the options and consequences.

a-m-zill:
Hi @javahippie,

I've now read your ADR, and first of all, thank you for your work. Overall, I really like it. 

I’ve been thinking for a long time about what the best approach is. Now I’ve looked into this from the caller’s perspective using the code `runtimeService.correlateMessage`. Karsten already asked about this in relation to the version. 

Questions:
* For the join, is it sufficient to provide just the process version as an integer (version_tag_ of process), or does the caller need to provide the full ProcessDefinitionID?
* From the modeler’s perspective, we should use the process’s version tag. Will that allow the join to work smoothly and easily?

a-m-zill:
Hello,

after a call with @javahippie i have an further update for this ADR.
We found different topics where we have to think over.
Here the list:

- How we want to handle tenant ID in this scope?
- What is with version tag for signal and conditional events?
- The intermediate event occurs also in the table **act_ru_event_subscr**, we have to handle this.

I vote for the idea to add the version_tag to the tabel **act_ru_event_subscr**.

PR #3296, opened 2026-07-07
Feature/1069 integrate new web apps

lambdaschmied2:
I'd like to make the first step integrating the web apps in the main repo.

This merge request is quite large, yet most of it is a duplication of existing Java code used in the legacy web apps and another big part is just the new files of the Preact application (which does not need to be checked).

Yet, if some active Java developers could do a sanity check (#it runs on my machine), this would be great.


kthoms:
@lambdaschmied2 I'll check it. When accepted I would squash the commits if you do not object. I will do the necessary work to make the build green again and maybe polish.

lambdaschmied2:
Squashing would be appropriate in my mind as well. Thanks for your time and effort!

kthoms:
@lambdaschmied2 I have merged the latest main and enabled the new webapps in org.operaton.bpm.spring.boot.starter.webapp.WebappTest
However, tests do not receive the expected 302. I did not investigate into deep there now.
Could you try to fix it so that we have a green build?

Further, I was wondering if `starter-webapp-neo(-core)` could be combined with their current "main" modules. There are differences, but maybe the "neo" modules could add them by overloading? Not sure if it is worth that effort, as we expect that the original modules will be just overwritten later when the old webapps are dropped.

javahippie:
Sanity-check finding from building and running the Run distro off this branch:

**The assembly never packages the new web apps** — `run.sh` advertises `--webapps-neo` and builds the loader path from `internal/webapps-neo/`, the module `operato
n-bpm-run-modules-webapps-neo` dutifully copies its jars to `target/dependency/`, but `distro/run/assembly/assembly.xml` only has `fileSet`s for `rest` and `oauth2
`. The built archive contains no `internal/webapps-neo/` (check: `tar -tzf distro/run/assembly/target/operaton-bpm-*.tar.gz | grep internal/webapps` → empty), so s
tarting with `--webapps-neo --rest` serves the Spring Boot Whitelabel 404 on `/`.

Fix that worked for me locally:

```xml
<fileSet>
  <directory>../modules/webapps/target/dependency/</directory>
  <outputDirectory>internal/webapps/</outputDirectory>
</fileSet>
<fileSet>
  <directory>../modules/webapps-neo/target/dependency/</directory>
  <outputDirectory>internal/webapps-neo/</outputDirectory>
</fileSet>
```

With that, both directories land in the archive and the new web apps are served at the root path.

Two related observations:

1. The missing `internal/webapps/` for the **legacy** webapps seems to predate this branch — `assembly.xml` is unchanged vs. `main` — so current Run snapshots may
have the same gap. Worth checking against a nightly artifact separately.
2. The test dependency of `starter-webapp-neo-core` on `operaton-webapp-neo-webjar` couples the build to the `distro-webjar-neo` profile: any build with explicit `
-P` flags (which disables `activeByDefault` profiles) fails with "Could not find artifact org.operaton.bpm.webapp-neo:operaton-webapp-neo-webjar". CI part-builds a
nd the upcoming `--webapps-only` optimization (#3340) would hit this. Suggest either moving the webjar module into the default reactor or decoupling the test depen
dency.

lambdaschmied2:
> @lambdaschmied2 I have merged the latest main and enabled the new webapps in org.operaton.bpm.spring.boot.starter.webapp.WebappTest However, tests do not receive the expected 302. I did not investigate into deep there now. Could you try to fix it so that we have a green build?

Yes, I'll have a look. I have skipped the tests till now, this is a good point in any case.

> Further, I was wondering if `starter-webapp-neo(-core)` could be combined with their current "main" modules. There are differences, but maybe the "neo" modules could add them by overloading? Not sure if it is worth that effort, as we expect that the original modules will be just overwritten later when the old webapps are dropped.

This part I don't really understand, as I probably lack context on the rest of the repo. Do you mean that I can re-use the existing legacy modules and use them as imported module in the neo apps?
If yes, looking at the goal of adding a new plugin system, I suspect that there will be less overlap sooner than later. And if we can cut out the legacy web apps by "just" deleting some folders and having no dependency toward them from the neo web apps, than I would keep it as is.

lambdaschmied2:
@kthoms @javahippie @cachescrubber all tests are green. The sonar quality gate should be ignored. The duplication stems from decoupling the legacy web apps from the new ones, so we can drop them without affecting the new ones, as soon as we are ready to do so.

kthoms (approved):
Feel free to merge. Very good work!

kthoms:
FYI: @cachescrubber is on vacation ATM

javahippie:
lgtm for the milestone. Anything that needs polishing will come up naturally :) Thanks!

PR #3340, opened 2026-07-15
[DRAFT] feat(ci): add prepare job to reduce build time for PRs

kthoms:
## Summary

- Adds a lightweight `prepare` job to `build.yml` (runs on `ubuntu-slim`) that inspects PR context and emits build optimization flags consumed by the `build` job
- Adds `--skip-engine-tests` and `--webapps-only` flags to `build.sh`, backed by a Python heuristics script and composite GitHub Action
- Removes the undocumented `skipTests.operaton-engine` property from `engine/pom.xml` in favor of the existing `test.excludes` Surefire mechanism
- Documents the new flags in `.github/copilot-instructions.md` and `AGENTS.md`

### Optimization heuristics (PR events only; push to main/release always runs full build)

| Flag | Maven effect | When triggered |
|------|-------------|----------------|
| `--skip-tests` | `-DskipTests` | `dependabot[bot]` PRs for `dependabot/github_actions` or `dependabot/npm_and_yarn` |
| `--skip-engine-tests` | `-Dtest.excludes=org/operaton/bpm/engine` | No changed files under `engine/`, `engine-dmn/`, `engine-rest/`, `engine-cdi/`, `engine-spring/`, `commons/`, `model-api/`, `juel/` |
| `--webapps-only` | `-pl webapps/assembly -am` | All changed files are under `webapps/` |

**Note:** `-pl webapps -am` is insufficient (resolves only 4 aggregator POM modules). `-pl webapps/assembly -am` builds the full 29-module reactor including engine and engine-rest as dependencies.

### Design

- Heuristics live in a pure-Python stdlib script (`prepare_build.py`), isolated from workflow YAML. New heuristics only require editing the script unless a new output variable is introduced.
- All heuristics fail safe: API failure, empty file list, or non-PR events all fall back to a full build.
- 20 unit tests covering all heuristics.
- Design doc: `docs/decisions/2026-07-15-ci-prepare-job-design.md`

## Test plan

- [ ] Verify `python3 -m unittest test_prepare_build -v` passes (20/20) in `.github/actions/prepare-build/`
- [ ] Open a dependabot GitHub Actions PR and confirm `prepare` emits `skip_tests=true`
- [ ] Open a webapps-only PR and confirm `prepare` emits `webapps_only=true` and `skip_engine_tests=true`
- [ ] Open an engine PR and confirm `prepare` emits all flags `false` (full build)
- [ ] Push to `main` (or simulate) and confirm all flags are `false`


engine/pom.xml
```diff
@@ -456,7 +455,7 @@
           <failIfNoTests>false</failIfNoTests>
           <trimStackTrace>false</trimStackTrace>
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
-          <skip>${skipTests.operaton-engine}</skip>
```
javahippie, line 459: Is this flag never used/overridden anywhere else?

.github/copilot-instructions.md
```diff
@@ -40,6 +40,24 @@
 .devenv/scripts/maintenance/code-cleanup.sh
 ```
 
+#### Faster Builds
+
+Use `.devenv/scripts/build/build.sh` flags for targeted builds:
```
javahippie, line 45: That's documentation for agents, I think we also had some docs for humans that described how the build scripts work?

.github/actions/prepare-build/prepare_build.py
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env python3
+"""CI prepare job: analyzes PR context and emits build optimization flags."""
```
javahippie: I get that Python is more platform independent than bash, but step by step we have a lot of Python files in our tooling, and for the db-schema preparation, as someone only roughly familiar with python, it took me some time to figure out dependencies, venv, etc. 

This script does not look too complex, having the logic in bash would not be an alternative?

.github/actions/prepare-build/prepare_build.py
```diff
@@ -0,0 +1,143 @@
...
+import json
+import os
+import sys
+import urllib.error
+import urllib.request
+
+ENGINE_DEPS = (
```
javahippie: This is not catching changes on parent/pom.xml, bom/pom.xml, database/*, test-utils/*.

On purpose?

.github/copilot-instructions.md
javahippie: .devenv/scripts/README.md, that was it

engine/pom.xml
kthoms: no, I have added it to allow easy disabling of the tests without knowing of tests.exclude

.github/copilot-instructions.md
kthoms: I'll add it

.github/actions/prepare-build/prepare_build.py
kthoms: I chose python because I expect that there will come more heuristics to compute other build shortcuts and assume that bash could become complex then

.github/actions/prepare-build/prepare_build.py
kthoms: nope, this is a gap, thank you

javahippie: Sorry for the second review wave: pom.xml in the root is missing, too.

.github/actions/prepare-build/prepare_build.py
```diff
@@ -0,0 +1,143 @@
...
+            headers={
+                "Authorization": f"Bearer {token}",
+                "Accept": "application/vnd.github+json",
+                "X-GitHub-Api-Version": "2022-11-28",
+            },
+        )
+        try:
+            with urllib.request.urlopen(req) as resp:
+                data = json.loads(resp.read())
+        except urllib.error.HTTPError as exc:
```
javahippie: catches HTTPError, but a URLError (connection/DNS) or JSON decode error will crash the script — and since build has needs: [prepare], a failed prepare job skips the build entirely. That inverts the fail-safe design goal: an API hiccup should mean "full build", not "no build". Suggest wrapping the whole heuristic section in a broad try/except that emits the default outputs, plus a unit test for that path.

engine/pom.xml
```diff
@@ -456,7 +455,7 @@
...
+          <skip>false</skip>
```
javahippie: Hardcoding "skip" to false would prevent developers from setting maven.test.skip and passing the behavior to surefire - so we could never skip surefire tests?

Issue #2875, opened 2026-04-30
Possible NoClassDefFoundError when loading FEEL engine


patptrs:
**Describe the bug**

Operaton release 2.1.0 contains jackson-module-scala_2.13-2.21.2.jar as a dependency originating from feel-engine-1.21.0
This was introduced to camunda-feel-engine with https://github.com/camunda/feel-scala/pull/1031

The classes in this jar file are actually shaded in feel-engine-1.21.0-scala-shaded.jar and properly reference the shaded library classes of scala-library which are prefixed with the package "camundajar.impl.". The class files in jackson-module-scala_2.13-2.21.2.jar however reference the unshaded classes from the original scala library.

Because the shaded jackson classes are not properly prefixed with a shade package and use the same package names as the original ones, if the original classes are loaded first, a NoClassDefFoundError may occur when classes from feel-engine are loaded.

I think that, similar to scala-library, jackson-module-scala should be excluded from the dependency tree of operaton-engine.

**To Reproduce**

1. Start an application using operaton-engine 2.1.0 where the classpath orders jackson-module-scala_2.13-2.21.2.jar before feel-engine-1.21.0-scala-shaded.jar
2. A NoClassDefFoundError occurs when instantiating a DmnEngine using FEEL: java.lang.NoClassDefFoundError: scala/collection/immutable/Seq

**Expected behavior**

Shaded classes should be loaded from feel-engine-1.21.0-scala-shaded.jar

**Additional context**

As a workaround I manually removed jackson-module-scala from the classpath when running operaton.

kthoms:
Thank you for your detailed explanation. This will be helpful to track down the bug.

We can see that the dependency hierarchy is
```
operaton-engine-dmn -> operaton-engine-feel-scala -> org.camunda.feel:feel-engine -> com.fasterxml.jackson.module:jackson-module-scala
```

<img width="719" height="219" alt="Image" src="https://github.com/user-attachments/assets/69317fb6-867b-4b9c-b511-c9cfcf1790b1" />

`jackson-module-scala_2.13-2.21.2.jar` **does not contain** scala classes with package prefix `scala.*`
<img width="314" height="168" alt="Image" src="https://github.com/user-attachments/assets/da52ef8d-5b8d-4eab-87c3-65176299f5c6" />

`feel-engine-1.21.0-scala-shaded.jar` contains shaded scala classes for package prefix `camundajar.impl.scala`.
[...]

PR #3342, opened 2026-07-16
deps: Switch to unshaded FEEL engine

javahippie:
This PR replaces the scala-shaded classifier of org.camunda.feel:feel-engine with the plain artifact and manages its real dependency tree instead. Closes #2654.

Why drop the shaded artifact? The shaded jar relocates Scala and fastparse to camundajar.impl.*, but bundles Jackson and paranamer without relocation. This causes two classes of problems that shading was supposed to prevent: duplicate-class conflicts with applications' own Jackson (flagged e.g. by the enforcer's banDuplicateClasses, see #2654), and classpath-order-dependent version mixing — the bundled Jackson 2.21 classes break against apps using jackson-annotations 2.22 (JsonTypeInfo$Value.getWriteTypeIdForDefaultImpl → NoSuchMethodError). With the unshaded artifact, FEEL deterministically uses the platform's Jackson, managed by our jackson-bom import.

Why list feel-engine's dependencies explicitly in the internal BOM? Maven would resolve them transitively; the explicit entries are about version control, not resolution:

- one consistent version platform-wide instead of per-module "nearest wins" mediation (e.g. feel-engine declares scala-library 2.13.18 while fastparse's POM references 2.13.14);
- jackson-module-scala must track the platform Jackson version (via the imported jackson-bom), otherwise we'd reintroduce the version mixing this PR removes;
- the Wildfly module descriptors reference exact jar filenames (fastparse_2.13-@version.fastparse@.jar etc.), which need a single version property as source of truth;
- explicit versions are visible to Renovate/Dependabot and security tooling.

Trade-off: these pins override whatever a future feel-engine release requires, so they must be bumped together with version.feel-scala — the entries are grouped under one comment in the BOM for exactly that reason. A drift in the parser/Jackson path fails loudly in the feel-scala and engine DMN test suites (precedent: the jackson-module-scala version guard is what failed #3040's CI).

Related cleanups this forces:

- camundajar.impl.scala.* imports in ScalaFeelEngine/ContextVariableWrapper (+ tests) become plain scala.* — the relocated classes no longer exist.
- engine/pom.xml drops the CAM-11822 exclusions of fastparse/geny/sourcecode/scala-library. They were harmless while those classes were shaded into the jar, but with the unshaded artifact they silently strip required runtime dependencies from every engine consumer.
- The verify-no-jackson-scala check in the Run assembly and the jackson-module-scala exclusion in the BOM (both from #2875) are removed — that workaround existed only because the shaded jar leaked Jackson service metadata; the module is now a regular, required runtime dependency.
- Wildfly gets proper modules for the new third-party jars (lihaoyi trio, paranamer, jackson-module-scala in the Jackson version slot); the Tomcat assembly bundles them into lib/, which newly includes the Jackson core jars in the shared classloader.

Compatibility note: the org.camunda.feel API packages are unchanged, but signatures that previously exposed camundajar.impl.scala.* types now expose scala.* — custom value mappers / function providers compiled against the shaded artifact need a recompile (and an import fix if they referenced camundajar types directly). Applications embedding Operaton alongside their own Scala or a divergent Jackson version should check the release notes. SBOMs are intentionally untouched; they are regenerated by the update-sbom workflow.


javahippie:
Alternatives considered

Fix the shading instead (relocate Jackson + paranamer too, upstream via camunda/camunda#56147 or self-shaded): only fully non-breaking option, but keeps ~20 MB of vendored classes invisible to scanners/SBOMs and caps Jackson upgrades — the bundled jackson-module-scala enforces < 2.22.0, which is why version.jackson is currently frozen (#3040). Shading doesn't remove the dependencies, it removes our ability to patch them. Valid 2.x fallback if this must wait for a major.

Wait for upstream (camunda/camunda#56147): same properties, timeline not ours. Worth pursuing in parallel anyway.

Dual artifacts (shaded default + unshaded opt-in, flip at next major): feasible — only two classes differ — but doubles the test/distro matrix for an interim state.

Classloader isolation: fails on the extensibility API — custom value mappers implement interfaces whose signatures reference Scala types, so Scala leaks back into the application classloader.

Document an exclusion recipe: proven broken in #2654 — our operaton-engine-feel-scala is compiled against the relocated camundajar.* types, so users can't unshade from the outside (ClassNotFoundException: camundajar.impl.scala.jdk.CollectionConverters).

Unshading is the only option that fixes the root cause — FEEL's dependencies become ordinary, visible, patchable Maven dependencies — at the cost of a documented one-time migration for custom FEEL integrations.

kthoms:
Is this even worth an ADR?

javahippie:
@kthoms Might be possible, it's a bigger change. Some kind of discussion would be adequate

kthoms:
[Integration Tests](https://github.com/operaton/operaton/actions/runs/29619914385)

kthoms:
@javahippie The integration tests are failing. Seems a simple dependency issue. Please make sure the test run succeeds. You'll have to trigger it manually.

distro/wildfly/modules/src/main/modules/org/camunda/feel/feel-engine/main/module.xml
```diff
@@ -1,15 +1,17 @@
 <module xmlns="urn:jboss:module:1.0" name="org.camunda.feel.feel-engine">
   <resources>
-    <resource-root path="feel-engine-@version.feel-scala@-scala-shaded.jar" />
+    <resource-root path="feel-engine-@version.feel-scala@.jar" />
   </resources>
 
   <dependencies>
     <module name="org.slf4j" />
-    <!-- Needed for feel-engine 1.20.0+ due to jackson-module-scala -->
     <module name="org.scala-lang.scala-library" />
```
kthoms:
```suggestion
    <module name="org.scala-lang.scala-library" export=true/>
```

Reason: The module `distro/wildfly/modules/src/main/modules/org/operaton/bpm/dmn/operaton-engine-feel-scala/main/module.xml` depends on this module and was able to load `scala.*` classes by the shaded relocated library. By unshading, the original classes have to be loaded, but are no dependency of `operaton-engine-feel-scala`. The export makes them visible to the downstream module.

kthoms (commented):
Just one issue, and the failing integration tests. Besides that make integration tests work and then ready to merge.

distro/wildfly/modules/src/main/modules/org/camunda/feel/feel-engine/main/module.xml
kthoms: done

kthoms:
Added required scala-library dependency to the db migration test.

[Integration Test](https://github.com/operaton/operaton/actions/runs/29801240395)

kthoms:
The integration tests show an error:
```
Caused by: org.jboss.modules.ModuleLoadError: com.lihaoyi.fastparse_2.13
```
see `qa/wildfly-runtime/target/server/wildfly-40.0.1.Final/standalone/log/server.log`

kthoms:
Reorganized the scala related modules
[Integration Test](https://github.com/operaton/operaton/actions/runs/29913801622)

javahippie:
Suggestion for release notes:

```
FEEL engine is no longer shaded

We now ship the plain org.camunda.feel:feel-engine artifact instead of the scala-shaded one. The shaded jar bundled unrelocated Jackson and paranamer classes, which caused duplicate-class conflicts and version mixing with application-provided Jackson (#2654, #2875). FEEL now simply uses the platform's Jackson.

What this means for you:

- Custom FEEL value mappers or function providers need a recompile. Method signatures now use plain scala.* types instead of the relocated camundajar.impl.scala.* ones. If you imported camundajar types directly, fix the imports.
- If you reference feel-engine with the scala-shaded classifier anywhere (dependency management, exclusions, enforcer rules), drop the classifier.
- The engine brings a few new transitive dependencies: the Scala 2.13 library, fastparse (with geny and sourcecode), jackson-module-scala and paranamer. If you embed Operaton next to your own Scala, check that you're on the 2.13 line.
- jackson-module-scala is a normal dependency now. If you call findAndRegisterModules() on an ObjectMapper, it will pick up DefaultScalaModule.

Distributions: Tomcat's shared lib/ now contains the Jackson and Scala jars (webapps with their own Jackson are unaffected — webapp classloading wins, but CVE scanners will now see jackson-databind in the distro). Wildfly gets proper modules for the new jars, and the feel-engine module exports the Scala library. Operaton Run bundles the new jars in its fat jar.
```

javahippie:
[Integration Tests](https://github.com/operaton/operaton/actions/runs/29990555100)

Edit: Failed due to Spring Boot tests.

PR #3344, opened 2026-07-16
fix(engine): Suppress java:S5738 for still-required deprecated-API usage

laurinbirchler:
Hi,

My first contribution. Please let me know if I'm doing anything wrong.

Three call sites use APIs that have been deprecated with forRemoval=true but are not replaceable yet, therefore the Sonar finding is suppressed instead of being worked around:

- CdiResolver#getWrappedResolver uses BeanManager#getELResolver which is not replaceable at the moment because ELAwareBeanManager from the CDI specification itself is not included in any version of jakarta.enterprise.cdi-api.
- CmmnExecution#collectAffectedSentries uses the deprecated CmmnSentryPart#getSourceCaseExecutionId because the field may actually contain information about sentry parts that were persisted prior to removing this API from the model.
- CommentEntity uses the deprecated Event interface just to remain compatible with TaskService#getTaskEvents (GetTaskEventsCmd).

### DIFF
--- a/engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/impl/el/CdiResolver.java
+++ b/engine-cdi/src/main/java/org/operaton/bpm/engine/cdi/impl/el/CdiResolver.java
@@ -47,11 +47,13 @@ public class CdiResolver extends ELResolver {
     }
   }
 
+  @SuppressWarnings("removal")
   protected ELResolver getWrappedResolver() {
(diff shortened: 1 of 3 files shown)


laurinbirchler:
#511

kthoms (approved):
Changed also the suppression to "removal" in CommentEntity, since the Sonar warnings was still shown

PR #3350, opened 2026-07-17
fix(sonar): eliminate non-linear regex backtracking (java:S8786)

laurinbirchler:
Good Evening,

I've looked at this five SonarCloud S8786 issues as part of #511. I've verified the changes with 1,000,000 randomized differential fuzz tests, checking old and new matches and capture groups, plus the current test Suites.

- **COLUMN_NAME_REGEX:** Rewritten to have deterministic matching with identical behavior. Equality proved by construction and fuzz testing (1M CAses, 0 differences). Removes the possibility of catastrophic backtracking (36 seconds → ~0ms on benchmark).
- **TaskDecorator.extractCandidates:** Greedy quantifiers replaced with possessive quantifiers (\s*+). No longer backtracks but still matches the same splits as before (1M case fuzz tests, 0 differences).
- **PropertyHelper.PROPERTY_TEMPLATE:** Rewritten using possessive quantifiers and deterministic matching. Equivalence proved by 1M fuzz tests (0 differences) and the current test suites.
- **IntervalTransformer.INTERVAL_PATTERN:** REwritten to ensure separators are unambiguous and have linear performance. Behaves identically for all valid FEEL expressions (already covered by the current test suites), except for the documented case of a malformed interval with .. in the endpoint.


kthoms (approved):
LGTM 
Added a unit test for testing the pattern in IntervalTransformer

PR #3362, opened 2026-07-20
fix(invoice-example): associate form labels with their inputs

laurinbirchler:
Hello there,

I reviewed some additional Sonar findings.

The embedded invoice forms displayed visible `<label>` elements that were not associated with their corresponding inputs. This change adds matching `id`/`for` pairs to the 16 affected input and select elements, following the existing pattern in `start-form.html`. It also corrects two stale `name` attributes (`Invoice Category` and `Approver`), which are unused because values are bound via `cam-variable-name`.

Although this only affects the invoice example not the runtime or platform itself, the example is shipped with Target distributions and commonly serves as a reference for users building embedded forms.

So It resolves 17 occurrences of `Web:InputWithoutLabelCheck`.

Have a great day!

### DIFF
--- a/examples/invoice/src/main/webapp/forms/approve-invoice.html
+++ b/examples/invoice/src/main/webapp/forms/approve-invoice.html
@@ -10,40 +10,44 @@
   </div>
 
   <div class="form-group mb-3">
-    <label class="form-label col-md-2">Creditor</label>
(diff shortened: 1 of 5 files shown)


laurinbirchler:
#511

kthoms (approved):
LGTM

PR #3363, opened 2026-07-20
fix(regexp): avoid superlinear regex scanning

laurinbirchler:
## What this PR does

Fixes the remaining four `java:S8786` (super-linear regex backtracking) findings.

### Why the previous fix wasn't enough

The earlier fix in `dc88275fa` (#3350) added possessive quantifiers (`*+`, `++`). That reduces backtracking *within* a single match attempt, but it doesn't prevent the real issue: when a match fails, the regex engine retries from the next character, rescanning the rest of the input each time. The result is still quadratic runtime. We confirmed this with worst-case inputs (up to 51 s / 13.5 s).

This PR removes the quadratic scan itself, so it supersedes the previous approach for these four cases.

### Changes

| File | Fix |
|------|-----|
| `PropertyHelper.resolveProperty` | Match only the `${...}` placeholder and rebuild the surrounding text with `appendReplacement`/`appendTail`, anchoring each match on the literal `${`. |
| `TaskDecorator.extractCandidates` | Split on `,\s*+` (anchored by the comma) and trim each part afterward. |
| `DatabaseNamingConsistencyTest` / `WebappsDatabaseNamingConsistencyTest` | Limit the unbounded character classes in `COLUMN_NAME_REGEX` to 200 characters, bounding the work per match attempt. |

**Performance:** resolving a 200k-character property value drops from **51 s to <1 ms**, and parsing a candidate list containing 120k whitespace characters drops from **13.5 s to <1 ms**. Since the second case can be triggered through a process variable, this is a DoS fix, not just a lint cleanup.

### Behavior change

`resolveProperty` no longer drops text before a stray `$`. For example:

- Before: `"$HOME/${key}/x"` → `"HOME/1234/x"`
- Now: `"$HOME/1234/x"`

This only affects values containing a literal `$` before a placeholder. The old behavior silently discarded text, so this is treated as a bug fix. A regression test covers the new behavior.

### Verification

- Ran Sonar's `SuperLinearRegexCheck` (same version as SonarCloud): all four findings disappear.
- Differential fuzzing (~27M inputs):
  - `extractCandidates`: no behavioral differences.
  - `resolveProperty`: all differences are the documented stray-`$` case.
- Compared the bounded column-name regex against all existing SQL/mapping resources and injected test cases: identical results. The longest real identifier is 74 characters, well below the 200-character limit.

## Questions @kthoms 

1. Is the `resolveProperty` behavior change acceptable as a normal bug fix, or should it be called out in the release notes?
2. Is a limit of 200 characters for `COLUMN_NAME_REGEX` reasonable, or would you prefer it as a named constant?

### DIFF
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/task/TaskDecorator.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/task/TaskDecorator.java
@@ -202,7 +202,27 @@ public class TaskDecorator {
    * Extract a candidate list from a string.
    */
   protected List<String> extractCandidates(String str) {
-    return Arrays.asList(str.split("\\s*+,\\s*+"));
+    String[] parts = str.split(",\\s*+", -1);
+    if (parts.length == 1) {
+      return List.of(str);
+    }
+    int size = 0;
+    for (int i = 0; i < parts.length; i++) {
+      int end = parts[i].length();
+      while (i < parts.length - 1 && end > 0 && isSpace(parts[i].charAt(end - 1))) {
+        end--;
+      }
+      parts[i] = parts[i].substring(0, end);
+      if (!parts[i].isEmpty()) {
+        size = i + 1; // mirrors String#split dropping trailing empty elements
+      }
+    }
+    return Arrays.asList(Arrays.copyOf(parts, size));
+  }
+
+  /** The regex {@code \s} class, which is narrower than {@link String#strip()}. */
+  private static boolean isSpace(char c) {
+    return c == ' ' || c == '\t' || c == '\n' || c == 0x0B || c == '\f' || c == '\r';
   }
 
   // getters ///////////////////////////////////////////////////////////////
--- a/engine/src/main/java/org/operaton/bpm/engine/impl/task/TaskDecorator.java
+++ b/engine/src/main/java/org/operaton/bpm/engine/impl/task/TaskDecorator.java
@@ -16,6 +16,7 @@
  */
 package org.operaton.bpm.engine.impl.task;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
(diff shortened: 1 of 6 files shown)


laurinbirchler:
#511

engine/src/main/java/org/operaton/bpm/engine/impl/task/TaskDecorator.java
```diff
@@ -202,7 +202,27 @@ protected void initializeTaskCandidateUsers(TaskEntity task, VariableScope varia
...
+      if (!parts[i].isEmpty()) {
+        size = i + 1; // mirrors String#split dropping trailing empty elements
+      }
+    }
+    return Arrays.asList(Arrays.copyOf(parts, size));
+  }
+
+  /** The regex {@code \s} class, which is narrower than {@link String#strip()}. */
+  private static boolean isSpace(char c) {
+    return c == ' ' || c == '\t' || c == '\n' || c == 0x0B || c == '\f' || c == '\r';
```
laurinbirchler, line 231: The original regex combined several behaviors: splitting on commas, consuming whitespace after delimiters, trimming whitespace before delimiters, preserving `\s` semantics, and matching `String.split()`'s handling of trailing empty elements. Once the problematic regex was removed, each of those behaviors had to be implemented explicitly.

Shorter alternatives were evaluated, including `split(",")` with `strip()`/`trim()`, but differential fuzzing found behavioral differences (over 117k cases). The current implementation was the smallest version that produced zero divergences across approximately 27 million test inputs.

If exact compatibility were not required, the implementation could be much shorter (for example, `Arrays.stream(str.split(",")).map(String::trim)`), but that changes the handling of leading/trailing whitespace and therefore is not a drop-in replacement.

kthoms:
@laurinbirchler Thank you for the PR. Sorry for delaying the review, I'm on vacation right now. Do not hesitate to file more PRs. I'll review when finding time.

Could you verify that changes to TaskDecorator are properly covered by TaskDecoratorTest, and add tests if appropriate?

Rgd. Questions:
1. this is just a regular bugfix
2. it is only related to test code, fair enough

kthoms:
Reviewed the code and fixed minor findings. Now ok for me.

kthoms (approved):
Thank you for your contribution @laurinbirchler !

PR #3367, opened 2026-07-21
test(webapps): isolate LoginUiIT per test to fix flaky logins

kthoms:
## Problem

`org.operaton.bpm.integrationtest.LoginUiIT` flakes on CI. In the failing runs, `shouldLoginToCockpit` passes while `shouldLoginToTasklist`, `shouldLoginToAdmin` and `shouldLoginToWelcome` all fail **identically** at `waiting for input[type="password"] to become invisible` (the submit never logs in) — on both WildFly and Tomcat, and the failure survives the existing 3× in-session retry (~93 s per test).

Example failures:
- https://github.com/operaton/operaton/actions/runs/29712354986/job/88258277393 (`shouldLoginToAdmin`)
- https://github.com/operaton/operaton/actions/runs/29712341650/job/88258237415 (`shouldLoginToWelcome`)

## Root cause

All four tests shared **one static `ChromeDriver`** (`@BeforeAll`/`@AfterAll`). The four webapps (cockpit/tasklist/admin/welcome) share a single `JSESSIONID`/`XSRF-TOKEN` pair, and `CsrfPreventionFilter.setCSRFToken()` only issues a new token when the session has none. Once a stale authenticated session outlived the in-browser cookie wipe, the browser never received a usable CSRF token, so every subsequent login POST was rejected. Only the first (clean) login in the shared browser reliably succeeded — which is exactly the observed "first test passes, rest fail" signature. Tests running *after* LoginUiIT passed, confirming the server was healthy and the contamination was confined to the shared browser.

Prior interaction-level fixes (`click`→`submit`, localStorage/sessionStorage clear, retries, longer timeouts) all failed with the identical pattern because they tweaked the login *inside* the contaminated browser instead of isolating the tests.

## Fix

- Create/quit a **fresh `ChromeDriver` per test** (`@BeforeEach`/`@AfterEach`), so every test starts from the clean state that reliably logs in.
- Resolve the driver lazily (`Supplier<WebDriver>`) in `SeleniumScreenshotExtension`, since it is constructed before `@BeforeEach`.
- Drop the now-redundant cookie/storage reset from `login()`. The robust waits + retries stay as defense against genuine cold-start slowness.

## Verification

Relying on CI: this PR runs the webapps integration suite on tomcat + wildfly (h2). Confirming across multiple runs to demonstrate the flakiness is gone.

### DIFF
--- a/distro/run/qa/integration-tests/src/test/java/org/operaton/bpm/run/qa/webapps/AbstractWebappUiIT.java
+++ b/distro/run/qa/integration-tests/src/test/java/org/operaton/bpm/run/qa/webapps/AbstractWebappUiIT.java
@@ -18,8 +18,7 @@ package org.operaton.bpm.run.qa.webapps;
 
 import java.net.URI;
 
-import org.junit.jupiter.api.AfterAll;
-import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.openqa.selenium.WebDriver;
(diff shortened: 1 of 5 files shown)


distro/run/qa/integration-tests/src/test/java/org/operaton/bpm/run/qa/webapps/AbstractWebappUiIT.java
```diff
@@ -37,14 +36,22 @@
...
+  /**
+   * A fresh browser is created for every test method (and disposed afterwards). The webapps share
+   * a single JSESSIONID/XSRF-TOKEN pair across cockpit/tasklist/admin/welcome, so reusing one
+   * browser across tests leaks authenticated-session and CSRF-token state: the first login
+   * succeeds, but a stale JSESSIONID that outlives an in-browser cookie wipe prevents the server
+   * from re-issuing an XSRF-TOKEN, so every subsequent login is rejected. A brand-new browser
+   * guarantees each test starts from the clean state that reliably logs in.
+   */
+  @BeforeEach
+  void createDriver() {
```
javahippie: There already is a `@BeforeEach`, maybe combine them? They are separate from each other, but the running order is not deterministic, which is a door for trouble when extending one of the methods later

javahippie (approved):
Upon retry, a `driver.manage().deleteAllCookies()` might help to restart with a clean state, instead a polluted context.

The two `@BeforeEach` should be combined into one. Rest looks fine to me!

distro/run/qa/integration-tests/src/test/java/org/operaton/bpm/run/qa/webapps/AbstractWebappUiIT.java
kthoms: done

PR #3390, opened 2026-07-23
fix(deps): Pin vulnerable transitive dependencies until the framework…

javahippie:
…s fix them

This is pinning netty and jackson because their respective Quarkus / Spring Boot versions have not shipped a version with an unaffected version, yet.

This is a proposal, it's to be discussed if we want to merge this before the release. Also to be discussed, if it should be cherry-picked into the 2.1. release branch

### DIFF
--- a/bom/internal-dependencies/pom.xml
+++ b/bom/internal-dependencies/pom.xml
@@ -64,6 +64,25 @@
         <scope>import</scope>
         <type>pom</type>
       </dependency>
+      <!-- Jackson 3 - Place before Spring Boot to override the vulnerable version
+        managed by spring-boot-dependencies (3.1.4, CVE-2026-59889). Remove once
+        Spring Boot manages jackson-bom >= 3.1.5. -->
+      <dependency>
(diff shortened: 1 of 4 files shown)


javahippie:
Manual Integration build trigger: https://github.com/operaton/operaton/actions/runs/30070722500

cachescrubber (approved):

cachescrubber:
We should cherry-pick to 2.1.x. Will take care of that.

kthoms:
Flagging per the PR description's own question: this should be cherry-picked into `release/2.1.x` ahead of the 2.1.3 release.

Context: PREPARE for the upcoming 2026-07-24 releases found several open HIGH-severity Dependabot alerts (netty, quarkus-vertx-http) that this PR resolves on `main`/2.2.0. Without a cherry-pick, 2.1.3 would ship without this fix while 2.2.0 has it.

cc maintainers — please cherry-pick to `release/2.1.x` before cutting 2.1.3, or confirm if there's a reason to defer.

PR #3406, opened 2026-07-24
fix(build): Harden prepare-deploy.sh

javahippie:


cachescrubber (approved):
a dry-run release will be the ultimate test

PR #3419, opened 2026-07-27
Chore/port integration pipeline to 2 1 x

javahippie:


.github/workflows/integration-build.yml
```diff
@@ -297,40 +279,100 @@ jobs:
           if [ "$SKIP_TESTS" = "true" ]; then
               CMD="$CMD --skip-tests"
           fi
+          # The Java 17 build stages the artifacts the integration-test matrix
+          # consumes with --no-build; integration-test-operaton-run adds the
+          # run-QA modules (example-plugin, integration-tests) the operaton
+          # distro cells resolve from the staged repository.
+          if [ "${{ matrix.java }}" = "17" ]; then
+              CMD="$CMD --extra-profiles=integration-test-operaton-run"
```
kthoms: The intention of "--profile=max" is to build with all maven profiles that could be built. Using "profile=normal" with extra profile "integration-test-operaton-run" would leave out API checks, for example.
I think what lead is to this change is that `integration-test-operaton-run` is missing on the profiles list for the "max" profile in build.sh.
This may make the "extra-profiles" switch obsolete, although maybe useful. Maybe name it `extra-maven-profiles` as the word `profile` is ambiguous.
Also, document the switch in the scripts/README.md

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,414 @@
+# Unified PR pipeline: builds and installs all artifacts ONCE (build stage),
+# then runs unit tests and — when required by PR labels — the integration test
+# matrix against the staged artifacts instead of rebuilding them per matrix cell.
+#
+# Rollout: runs alongside build.yml / integration-build.yml until validated.
```
kthoms: Remove this comment, the triggers are already removed

kthoms (approved):
LGTM

Issue #3480, opened 2026-08-13
Support calling AI agents over the A2A (Agent2Agent) protocol from BPMN


FaediMichele:
## Proposal

A **Connect SPI connector**, registered via `ServiceLoader` under connector id `a2a`, in a new opt-in module
`connect/a2a` (`org.operaton.connect:operaton-connect-a2a`).

```xml
<serviceTask id="askAgent">
  <extensionElements>
    <operaton:connector>
      <operaton:connectorId>a2a</operaton:connectorId>
      <operaton:inputOutput>
        <operaton:inputParameter name="operation">sendSync</operaton:inputParameter>
        <operaton:inputParameter name="url">${agentUrl}</operaton:inputParameter>
        <operaton:inputParameter name="message">${question}</operaton:inputParameter>
        <operaton:outputParameter name="answer">${text}</operaton:outputParameter>
      </operaton:inputOutput>
    </operaton:connector>
  </extensionElements>
</serviceTask>
```

A connector gets declarative input/output mapping in the properties panel for free, runs in the engine so there
is no extra service to operate, and leaves the activity a plain `bpmn:serviceTask` so the BPMN stays portable.
No new BPMN element type, no custom `ActivityBehavior`, no `BpmnParseListener`.

### Operations

| `operation` | Behaviour |
|---|---|
| `sendSync` | Send, then poll until the task settles. Bounded by a configurable `waitTimeout`. |
| `sendAsync` | Send with an A2A push-notification config and return immediately with task id, context id and state. Without a `callbackUrl`, it is the send half of a polling loop. |
| `getTask` | Read a known task, for a timer-driven polling fallback when the agent card lacks `capabilities.pushNotifications`. |

### Surface

Inputs, each a literal or an expression: `operation`, `url`, `headers`, `message`, `parts` (text/file/data),
`metadata`, `extensions`, `referenceTaskIds`, `contextId`, `taskId`, `messageId`, `requestMetadata`, `tenant`,
`acceptedOutputModes`, `historyLength`, `callbackUrl`, `callbackToken`, `callbackAuthScheme`,
`callbackAuthCredentials`, `connectTimeout`, `readTimeout`, `waitTimeout`, `pollInterval`, `idempotencyKey`,
`reattachOnRetry`, `maxVariableSize`, `includeHistory`.

Outputs: `text`, `statusMessage`, `artifacts` (all of them, with their parts), `artifactText`, `task`, `taskId`,
[...]

PR #3481, opened 2026-08-13
feat(connect): Add A2A protocol connector

FaediMichele:
Opened as a **draft**: this adds a new engine-side module and new third-party jars, which is a maintainer-level
call. I would rather get a verdict on the dependencies and the layering before polishing. See
#3480 for the full proposal and the open questions.

## What this adds

A Connect SPI connector, id **`a2a`**, in a new opt-in module `connect/a2a`, so a modeller can call an
[A2A v1.0](https://a2a-protocol.org/) agent from a plain `bpmn:serviceTask` without writing Java.

```xml
<operaton:connector>
  <operaton:connectorId>a2a</operaton:connectorId>
  <operaton:inputOutput>
    <operaton:inputParameter name="operation">sendSync</operaton:inputParameter>
    <operaton:inputParameter name="url">${agentUrl}</operaton:inputParameter>
    <operaton:inputParameter name="message">${question}</operaton:inputParameter>
    <operaton:outputParameter name="answer">${text}</operaton:outputParameter>
  </operaton:inputOutput>
</operaton:connector>
```

Three operations: `sendSync` (send, then poll until the task settles, bounded by `waitTimeout`), `sendAsync`
(send with a push-notification config, return the ids immediately) and `getTask` (for a timer-driven polling
fallback). Full input/output/error reference in [`connect/a2a/README.md`](connect/a2a/README.md).

## Design notes for review

- **All SDK usage sits behind `A2aAgent`**, an interface whose methods take and return only strings, maps and
  lists. `SdkA2aAgent` is the only class importing `org.a2aproject.sdk`, so an SDK or protocol bump touches it and
  `TimeoutAwareA2AHttpClient` and nothing else.
- **`TimeoutAwareA2AHttpClient` exists out of necessity.** The SDK's `JdkA2AHttpClient` never calls
  `HttpRequest.Builder.timeout(...)`, and neither `A2AHttpClient` nor its builders expose a way to set one.
  Without it an unresponsive agent would hold a job executor thread until the OS gave up on the socket.
- **Streaming is deliberately off.** `sendSync` polls instead, so no job executor thread is pinned to an open SSE
  connection for the duration of the agent's work.
- **Not in `connectors-all`, no WildFly module.** `connectors-all` shades and relocates `org.apache`, which is not
  safe for protobuf, and shipping protobuf/Guava as JBoss modules is a distribution decision. Both are called out
  as open questions in the issue.
- **Header values and request bodies are never logged**; header names are logged at debug so a missing
  `Authorization` is still diagnosable.

## Dependencies added

`org.a2aproject.sdk:a2a-java-sdk-client:1.2.0.Final` (BOM imported in `bom/internal-dependencies`, version
property in `parent/pom.xml`). Its closure adds `protobuf-java`, `protobuf-java-util` (→ Guava) and
`proto-google-common-protos`; `gson` was already managed, and the two Jakarta APIs come from the Jakarta EE BOM.
**No Netty, no Jackson conflict, no shading.** ADR: `docs/decisions/0002-a2a-connector-uses-the-a2a-java-sdk.md`.

## Verified against a live agent and inside a real engine

This has been run end to end against a real A2A v1.0 agent (google-adk with `a2a-sdk` 1.1.2), first directly and
then from a BPMN process inside an `operaton/operaton:2.1.2` Run container with the connector dropped into
`configuration/userlib`. A service task calling the agent produced:

```
answer          = 4
a2aState        = TASK_STATE_COMPLETED
a2aTaskId       = f0b204bdc8d3477a97cc14c67dddd12f
a2aContextId    = ctx-c729d7ecd9210450e0c332c4ba353288
a2aArtifactText = 4
```

A second process asks for a task id that does not exist. Its activity history is
`readMissingTask → caughtProtocolError (boundaryError) → handledError`, which is direct evidence that a terminal
A2A failure becomes a `BpmnError` an error boundary event catches, in a real engine rather than a mock.
`ConnectProcessEnginePlugin` is active by default in Operaton Run, so no extra configuration was needed.

Getting there found four real defects, all fixed here with regression tests:

1. **The reattach probe took down every send to an agent without `ListTasks`.** `ListTasks` is optional and many
   agents answer `-32601 Method not found`. That failure was classified as retryable and rethrown, so with
   `reattachOnRetry` defaulting to `true` every `sendSync` to such an agent failed permanently. The probe is now
   best-effort: only a proven transport failure (`IOException`, or HTTP 408/425/429/5xx) fails the job.
2. **`text` was empty for agents that answer with artifacts.** Some agents close a task with a status message,
   others return only artifacts and no message. `text` read only the status message, so it was blank against the
   second kind, which is the output the examples and the element template lead with. It now falls back to the
   artifact text.
3. **The example processes could not be deployed.** None set `historyTimeToLive`, so any engine with `enforceTTL`
   on, the default, rejects them. All now set it.
4. **A transport failure named the wrong URL.** The `url` input locates the agent *card*; the endpoint actually
   called is the one the card advertises in `supportedInterfaces[].url`. The message now says so and includes the
   cause, so a card advertising an unreachable address is not mistaken for an agent being down.

Also confirmed empirically: `protobuf-java` and `a2a-java-sdk-spec-grpc` are required **at runtime** on the pure
JSON-RPC path, exactly as the ADR states. The client will not construct without them.

## Verification, stated precisely

- ✅ **58 unit tests pass**, run on the JUnit console launcher against published Operaton artifacts.
- ✅ **Wire-level tests** (`SdkA2aAgentWireTest`) drive the real JSON-RPC protocol over WireMock, using payloads
  copied from a real agent: `SendMessage` wraps its result in a `task` envelope while `GetTask` does not, parts
  carry a bare `text` field with no kind discriminator, and `-32601` on `ListTasks` is covered explicitly.
- ✅ **End-to-end in an Operaton Run 2.1.2 container** against a live agent, including the error-boundary path.
- ✅ Compiles on Java 17 against both `2.2.0-M2` and `2.1.2` Operaton artifacts.
- ✅ The element template validates against `@camunda/element-templates-json-schema` with 0 errors.
- ❌ **`mvn clean install` has not been run**, and neither has `.devenv/scripts/maintenance/code-cleanup.sh`
  (OpenRewrite). There is no Maven on the machine I authored this on. Style was matched by hand to
  `connect/http-client`. Please treat CI as the first signal on the full reactor build.
- ❌ **`A2aConnectorProcessTest` has not been executed** in-process; it compiles, and the behaviour it asserts was
  instead demonstrated in the real container as described above.

## Tests included

`A2aIdsTest`, `A2aParamsTest` (literal-vs-expression coercion, plus null/empty/blank cases), `A2aResponseImplTest`
(output mapping, multi-artifact reachability, truncation, the artifact-text fallback), `SdkA2aAgentTest` (the
reattach-probe classification), `SdkA2aAgentWireTest` (the real protocol over WireMock), `A2aConnectorImplTest`
(operation dispatch, retry reattachment, polling, timeout, and each error code), and `A2aConnectorProcessTest`,
which runs the three example processes on an in-memory engine.

## Checklist

- [x] New files carry the Apache 2.0 license header
- [x] Public API has Javadoc and `@since 2.2`
- [x] Unit tests written (JUnit 5 + AssertJ)
- [x] ADR added for the new major dependency
- [x] Conventional Commits, footer references #3480
- [ ] `mvn clean install` green — needs CI, see above
- [ ] `code-cleanup.sh` run — needs a Maven environment

### DIFF
--- a/bom/connect/pom.xml
+++ b/bom/connect/pom.xml
@@ -31,6 +31,11 @@
       </dependency>
       <dependency>
         <groupId>org.operaton.connect</groupId>
+        <artifactId>operaton-connect-a2a</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.operaton.connect</groupId>
         <artifactId>operaton-connect-connectors-all</artifactId>
         <version>${project.version}</version>
       </dependency>
--- a/bom/internal-dependencies/pom.xml
+++ b/bom/internal-dependencies/pom.xml
@@ -149,6 +149,14 @@
         <scope>import</scope>
         <type>pom</type>
       </dependency>
+      <!-- A2A protocol SDK, used by the A2A connect connector -->
+      <dependency>
+        <groupId>org.a2aproject.sdk</groupId>
+        <artifactId>a2a-java-sdk-bom</artifactId>
+        <version>${version.a2a-sdk}</version>
+        <scope>import</scope>
+        <type>pom</type>
+      </dependency>
       <dependency>
         <groupId>org.seleniumhq.selenium</groupId>
         <artifactId>selenium-java</artifactId>
--- a/connect/a2a/README.md
+++ b/connect/a2a/README.md
@@ -81,7 +81,7 @@ strings, so `3000` and `"3000"` both work.
 | Input | Meaning |
 |---|---|
 | `operation` | **Required.** `sendSync`, `sendAsync` or `getTask`. |
-| `url` | **Required.** Either a service base URL such as `https://agent.example.com`, in which case the agent card is read from `https://agent.example.com/.well-known/agent-card.json`, or a full agent card URL ending in `.json`, which is fetched as given. |
+| `url` | **Required.** Either a service base URL such as `https://agent.example.com`, in which case the agent card is read from `https://agent.example.com/.well-known/agent-card.json`, or a full agent card URL ending in `.json`, which is fetched as given. **This locates the card only.** The endpoint actually called for A2A requests is the one the card advertises in `supportedInterfaces[].url`, so an agent whose card says `http://localhost:8000` is unusable from anywhere but its own host. If calls fail while the card fetch succeeds, check what the card advertises. |
 | `headers` | A map of HTTP headers sent with every A2A call, including the agent card fetch. This is where `Authorization` goes. |
 
 ### Message
--- a/connect/a2a/README.md
+++ b/connect/a2a/README.md
@@ -1,0 +1,329 @@
+# Operaton Connect - A2A
+
+Call an AI agent from a BPMN service task over the [Agent2Agent (A2A) protocol](https://a2a-protocol.org/),
+without writing Java.
+
+The activity stays a plain `bpmn:serviceTask`. Everything is configured with `operaton:inputParameter` and
+`operaton:outputParameter`, so a new agent call is a matter of dragging a task onto the canvas and filling in
+fields.
+
+```xml
+<serviceTask id="askAgent" name="Ask the agent">
+  <extensionElements>
+    <operaton:connector>
+      <operaton:connectorId>a2a</operaton:connectorId>
+      <operaton:inputOutput>
+        <operaton:inputParameter name="operation">sendSync</operaton:inputParameter>
+        <operaton:inputParameter name="url">${agentUrl}</operaton:inputParameter>
+        <operaton:inputParameter name="message">${question}</operaton:inputParameter>
+        <operaton:outputParameter name="answer">${text}</operaton:outputParameter>
+      </operaton:inputOutput>
+    </operaton:connector>
+  </extensionElements>
+</serviceTask>
+```
+
+- Protocol: **A2A v1.0**
+- Transport: **JSON-RPC 2.0** only (see [Limitations](#known-limitations))
+- Built on the [A2A Java SDK](https://github.com/a2aproject/a2a-java) `org.a2aproject.sdk:a2a-java-sdk-client`
+
+## Installation
+
+The connector is not part of `operaton-connect-connectors-all`, so it has to be added deliberately.
+
+1. Add the connect plugin to the engine, exactly as for the HTTP connector:
+
+   ```xml
+   <property name="processEnginePlugins">
+     <list>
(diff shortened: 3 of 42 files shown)


javahippie:
Hi @FaediMichele, thanks for the suggestion and sorry for the late response!

The change itself seems sound and thought out really well. We already had some discussion among maintainers about your suggestions, which are still pending but I wanted to share some of our thoughts:

* The functionality looks helpful and we think many others could benefit from this
* We are unsure about adding new integrations in the monorepo per se. We are often asked about Kafka integrations, MQTT integrations, etc. and usually our response is "our engine is extensible, this could be a library or process engine plugin"
* If we add this to the monorepo, changes in this (really still young) protocol need to be made by us. This includes breaking API changes, considerations about backwards compatibility and API versioning. Everything in this monorepo has the same version identifier and is released with the same process (this causes other challenges, too, but we won't change this in the foreseeable future)
* Agent calls can take a lot of time. synchronous integrations like this live on the main job executor threads and completely block them with IO. We usually recommend the external task pattern for longer, blocking activity (but also see the benefits of the way your proposal integrates with the engine)

We are still sorting some thoughts on this matter, because it has implications for other, similar integrations in the future. If your proposal was a process engine plugin, it was not tied to our release cadence and versioning, so we still try to understand the frequency of change that is to be expected with the protocol and find a generic approach for such integrations in the future.

FaediMichele:
Closing this out — thanks again for the thoughtful feedback and for taking the time to discuss it with the other maintainers.

Given the concerns raised (monorepo versioning/release coupling, ownership of a young/evolving protocol, and the sync-vs-external-task threading question), I think your instinct to keep this out of the core monorepo for now is right. I've since reimplemented the same functionality as a standalone external connector/process engine plugin outside this repo. It isn't tied to the engine's release cadence, and each of the concerns above is naturally resolved by living outside the monorepo.

Unfortunately, due to company policy the resulting connector has to stay private for now, so I can't share the repo directly. That said, I can confirm the external version works well in practice — no functional drawbacks compared to what was proposed here, and it sidesteps the versioning/threading concerns you raised.

Thanks again for engaging with the proposal in depth — the discussion was genuinely useful in shaping the external version, and happy to share learnings/design notes here if that's useful for whatever generic approach you land on for future protocol integrations like this.

Closing the PR accordingly.

PR #3491, opened 2026-08-14
chore(build): add opt-in Maven build cache extension (PoC)

### LINKED ISSUE #2836: Reduce size of build cache
## What needs to be done?

The build stores >2GB as build artifacts. This should be optimized.

https://github.com/operaton/operaton/actions/caches

Image

## Additional context

We observe sporadic build failures when a build caches something which is used later (e.g. documentation used by the release job in integration build). It can be that a build cache is evicted before, making the overall build fail.

There is only 10 GB available for caches.

javahippie:
Registers maven-build-cache-extension 1.2.3 with a filesystem-only cache (no remote). Disabled by default via maven.config so developer builds are unaffected; opt in with -Dmaven.build.cache.enabled=true.

Measured on a cold+warm local cycle (-DskipTests): 2:49 -> 1:23 min with 116/122 modules restored. Incompatibilities found and addressed:

- operaton-external-task-client opts out of caching: Cargo's pingURL references a runtime-reserved port and fails URL parsing when the extension replays the cached mojo list
- frontend-maven-plugin runs always: a restore only brings back Maven artifacts, not webapps/frontend/target/webapp, which the operaton-webapp WAR consumes as a resource directory across module boundaries - the WAR was silently packaged without the frontend
- dependency:unpack runs always: the forced npm build in engine-rest/docs consumes the OpenAPI spec it unpacks into target/
- .claude, node_modules and node are excluded from input checksums (identical rebuilds missed the cache otherwise)
- server distro subtrees (Run, Tomcat, WildFly, JBoss) opt out of caching: their archives were ~880 MB of a ~1 GiB cache, trading a few seconds of zip work for a lot of storage

closes #2836


kthoms:
This works quite well. Without the cache I have warm build time of ~10min, with the cache ~3min.

I also like that it is easily usable by the file system cache, which could be cached on CI and opt-in.

Please go ahead integrating this into the normal build workflow.

javahippie:
https://github.com/operaton/operaton/actions/runs/32225393454

Integration Build

.github/workflows/build.yml
```diff
@@ -138,14 +148,30 @@ jobs:
...
+          echo "cached module builds: $(find ~/.m2/build-cache -name buildinfo.xml 2>/dev/null | wc -l)"
+      - name: Save Maven build cache
+        uses: actions/cache/save@v6.1.0
+        with:
+          path: ~/.m2/build-cache
+          key: "${{ runner.os }}-build-cache-${{ github.sha }}"
       - name: Publish Test Report
         if: always()
         uses: mikepenz/action-junit-report@d9f48fc87bc235f7e214acf696ca5abc0a986f16 #v6.4.2
-        continue-on-error: true
```
kthoms, line 131: do not delete

.github/workflows/integration-build.yml
```diff
@@ -266,14 +266,18 @@ jobs:
           echo "database_version=$DATABASE_VERSION"
           SKIP_TESTS=${{ github.event.inputs.skip_tests || false }}
 
-          CMD=".devenv/scripts/build/build.sh --profile=max --reports"
+          # -Dmaven.build.cache.enabled=false: this job publishes artifacts, so
+          # everything it ships is built from source. The build cache is
+          # enabled only in build.yml and pr-build.yml. Passed explicitly
+          # rather than relying on the default in .mvn/maven.config.
+          CMD=".devenv/scripts/build/build.sh --profile=max --reports -Dmaven.build.cache.enabled=false"
```
kthoms, line 273: `maven.build.cache.enabled` is false by default, see `maven.config`. Passing it to mvnw should be obsolete. We can rely on `maven.config` on CI.

.github/workflows/pr-build.yml
```diff
@@ -156,6 +156,15 @@ jobs:
           path: ~/.m2
           key: "${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}"
           restore-keys: ${{ runner.os }}-m2
+      # Restore-only, like the m2 cache above: entries written here would be
+      # readable only by this PR and would just churn the shared pool. The
```
kthoms, line 160: remove "this PR". Once merged this is confusing.

.github/workflows/release.yml
```diff
@@ -162,7 +162,14 @@ jobs:
         run: |
           PROJECT_ROOT=$(pwd)
           # Make a verification build. Also the deploy step will need the tomcat distro for building, but we don't want to deploy it.
+          # -Dmaven.build.cache.enabled=false: released artifacts are always
```
kthoms, line 165: see above - is implicit

PR #3515, opened 2026-08-19
fix(webapps-neo): Make the packaged web app usable and close its security defects

### LINKED ISSUE #3514: webapps-neo is unusable in the packaged distribution, plus security defects in its default configuration
Following the merge of #3296, a review of the packaged Run distribution — built, booted in four configurations and probed — found that **webapps-neo cannot be used at all in the shipped artifact**, plus several security defects in its default configuration.

Everything below was reproduced against a real `operaton-bpm-run` build, not read off the source.

## The web app does not load

**1. The Content-Security-Policy blocks the SPA's own bundle.** The page renders blank.

The policy carries a nonce together with `'strict-dynamic'`. `'strict-dynamic'` makes browsers ignore `'self'`, `https:` and `'unsafe-inline'` wherever it appears, so only scripts carrying that nonce may run. The neo `index.html` is served statically from the classpath, its single `<script src>` has no nonce attribute, and nothing substitutes one:

```
Loading the script 'http://localhost:8080/assets/index-*.js' violates the following
Content Security Policy directive: "script-src 'nonce-…' 'strict-dynamic' …".
Note that 'strict-dynamic' is present, so host-based allowlisting is disabled.
The action has been blocked.
```

The policy was inherited from the legacy webapp, where it is correct — `ProcessEnginesFilter` templates `index.html` and stamps the nonce onto its scripts. webapps-neo has no such filter and no reader for the nonce it generates.

This was not caught because the e2e and accessibility suites run against the Vite dev server, which never applies the header filter, and CI has no check that loads the packaged application in a browser.

**2. Login always fails with 403.** The SPA posts to `…/login/neo`, but `"neo"` was not among the application ids `AuthenticationUtil` probes for, so `authorizedApps` could never contain it and `doLogin`'s containment check rejected every login — correct credentials included.

`NeoApiNamespaceIT` asserted the endpoint answered any `4xx`, which a 403 satisfies, so nothing failed.

**3. A deployment with no administrator cannot be bootstrapped.** The legacy webapp creates the first admin through its setup resource; webapps-neo has no equivalent. Any database not seeded via `operaton.bpm.admin-user` — a production configuration, for instance — leaves the SPA with a login mask and no account to use on it.

## Security defects in the default configuration

**4. The CSP nonce was application-global.** Both webapps generated a per-response nonce and stored it on the `ServletContext`, so concurrent requests overwrote each other's. Measured with eight concurrent clients: the legacy webapp alone served a mismatched nonce on **29 of 60** page loads; with webapps-neo alongside it, **49 of 60**, because neo maps its header filter to `/*` and so rewrote the shared value on every request anywhere on the server. A mismatch blocks every inline script on the page.

The race is pre-existing; webapps-neo amplifies it.

**5. Rejected requests lost their security headers.** The header-security and session-cookie filters were registered after the filters that reject requests, so a 401 or 403 left the chain before either ran: refusals went out with no CSP, no `X-Content-Type-Options`, no HSTS, and the `JSESSIONID` minted on that request went out without `SameSite`.

**6. `/api/admin/**` was absent from `deniedPaths`,** where the legacy webapp denies it and re-allows specific shapes. Not exploitable as it stands — the admin application mounts only the authentication resource, whose endpoints must be anonymous anyway — but any resource added there later would have been unauthenticated with no code change and no failing test.

Adding the deny rule turned out not to be enough on its own, because of two further bugs in the rule evaluation: a denied path matching no allow entry was granted anonymously rather than refused, and a rule ending in `/` compiled to a regex without it and never matched, which had silently disabled the engine-list rule.

## Smaller items

- `operaton.bpm.webapp.csrf.target-origin` is unset in both shipped configurations, so the CSRF filter's same-origin check is inert and only the token applies. Undocumented.
- `docker-compose-full.yaml` and `docker-compose-deploy.yaml` build from `./operaton`, deleted during the #3296 review; neither stack has been buildable since.
- A failed login set a state the router had no branch for, so the page went blank instead of showing an error.

## Deliberately out of scope

- The pre-existing Run defaults: unauthenticated `/engine-rest`, `Access-Control-Allow-Origin: *`, the `demo:demo` account, open WADL and `/actuator/health`. These are long-standing product behaviour, documented in `SECURITY-BEST-PRACTICES.md`, and `--production` closes them.
- The rule-evaluation weaknesses under (6) exist in the legacy webapp's copy as well. Changing shared evaluation semantics there needs its own audit and regression pass, so only the webapps-neo copy is addressed here.
- No browser-level smoke test of the packaged distribution. Worth adding — it is what would have caught (1), and neither the Maven build nor CI exercises the packaged app in a browser.

lambdaschmied2:
Fixes #3514.

Makes webapps-neo actually usable in the packaged distribution and closes the security defects found reviewing #3296 against a real build.

Every fix below was verified against a running `operaton-bpm-run`, and the two that are only observable in a browser were verified in one.

## Commits

| | |
|---|---|
| `fix(webapps-neo): Stop the CSP blocking the SPA's own bundle` | The page rendered blank. `'strict-dynamic'` disables the `'self'` allowance, and the static `index.html` carries no nonce, so the browser refused the application bundle. Drop the nonce and `'strict-dynamic'` from the neo default; narrow the host allowlist to `'self'` in exchange. |
| `fix(webapps-neo): Let the SPA log in` | `"neo"` was not an application id the login check knew, so every login answered 403. |
| `feat(webapps-neo): Create the first administrator from the SPA` | Ports the legacy setup resource with its guard, adds an availability endpoint and a first-run screen. |
| `fix(webapps): Scope the CSP nonce to the request` | The per-response nonce lived on the `ServletContext` and was overwritten by concurrent requests. |
| `fix(webapps-neo): Decorate rejected requests too` | 401/403 responses carried no security headers and their session cookie no `SameSite`. |
| `fix(webapps-neo): Make the security filter rules actually deny` | `/api/admin/**` was not denied, and two evaluation bugs meant a deny rule did not deny. |
| `docs(distro): Explain the CSRF target-origin setting` | The same-origin check is inert unless configured; document it. |
| `chore(webapps-neo): Drop the two broken compose stacks` | Both build from a directory deleted during the #3296 review. |

## Verified against a running build

| | before | after |
|---|---|---|
| SPA loads in a browser | blank page, bundle blocked by CSP | dashboard renders, login works end to end |
| `POST …/login/neo` with `demo:demo` | `403` | `200`, `authorizedApps` includes `neo` |
| First-run setup on an empty database | endpoint absent (`404`) | create → `204`, closes itself, second attempt → `403`, new admin signs in |
| Legacy nonce matches its CSP header (8 concurrent clients) | 11/60 | **60/60** |
| Anonymous `GET /api/engine/engine/default/user` | `401`, no CSP/`nosniff`, cookie without `SameSite` | `401` with all three |
| Anonymous unlisted `/api/admin/**` | reachable | `401` |

## Tests

378 webapps-neo unit · 687 webapps unit · 57 starter integration · 588 frontend — all green, frontend lint clean.

New: four `UserAuthenticationResourceTest` cases pinning the login grant and refusal (three fail without the fix), `SecurityFilterRulesTest` mirroring the legacy rules test, `SetupResourceTest`, `RejectedRequestHeaderSecurityIT`, a CSP regression test, and frontend coverage for the setup screen and API.

`NeoApiNamespaceIT` accepted any `4xx` on the login endpoint, which a 403 satisfies — that is why the login blocker went unnoticed. It now excludes 403.

## Notes for review

- One commit touches the legacy webapp: the nonce fix needs `ProcessEnginesFilter`, the only reader, to take the value off the request. The race is pre-existing there (29/60 mismatches with no neo involved), so fixing it request-side repairs both rather than only de-amplifying.
- The rule-evaluation fixes are applied to the webapps-neo copy only. The legacy copy shares the behaviour, but changing it needs its own audit — noted in #3514.
- Pre-existing Run defaults are untouched: unauthenticated `/engine-rest`, wildcard CORS, `demo:demo`. `--production` already closes them.
- Neo's CSP is now deliberately different from legacy's, and stricter: no `https:` and no `'unsafe-inline'` in `script-src`, since the bundle is same-origin and there are no inline scripts.
- CI would not have caught the blank page: the e2e and accessibility suites run against the Vite dev server, which does not apply the header filter. A browser-level smoke test of the packaged app is worth adding separately.


lambdaschmied2:
@javahippie can you have a look, once the checks are through?

Issue #3537, opened 2026-08-22
Use zizmor to harden GitHub Actions


javahippie:
## What needs to be done?

I would like to incorporate [zizmor](https://docs.zizmor.sh/) as a GitHub action. This does static analysis for our pipelines and catches things like unpinned actions, template expansion, too broad credential lifecycle.

This would protect us from supply chain attacks or some targeted attacks (e.g. a pull request with malicious changes to our GH actions)

## Additional context

I would like to add it as a step to our pipelines. The first changes would be as a PoC, so we can check how much time this adds to the pipeline runs and if it's better suited for nightlies, or can be added at the beginning of all pipelines in Pull Requests, too

kthoms:
This is helpful and I am willing to work on these issues. Not sure if we need nightly reports. I think a weekly schedule and on demand manual dispatch is enough. This keeps the workflow simple.

Once we have hunted the findings down we can make this a PR check maybe.

javahippie:
@kthoms would be fine for me! 

If we enable this for PRs, I would integrate it into the PR build directly. If a contributor creates a PR, I would prefer the pipeline to be stopped by zizmore before anything else runs and can cause harm

kthoms:
I have resolved all issues and configured the workflow to execute only on changes to analyzed resources. Then it is fine to run on PRs.

PR #3538, opened 2026-08-22
chore: Enable static pipeline analysis with zizmor

javahippie:
Relates to #3537


.github/workflows/zizmor-check.yml
```diff
@@ -0,0 +1,26 @@
+name: GitHub Actions Security Analysis with zizmor 🌈
+
+on:
+  push:
+    branches: ["main", "release/**"]
```
kthoms, line 8: restrict to paths that zizmor actually checks. This avoids runs on unrelated code.

.github/workflows/zizmor-check.yml
```diff
@@ -0,0 +1,26 @@
...
+  pull_request:
+    branches: ["**"] # If the runtime is quick enough
+
+permissions: {}
```
kthoms: is this obsolete?

.github/zizmor/config.yml
```diff
@@ -0,0 +1 @@
+rules: {}
```
kthoms: obsolete?

.github/workflows/zizmor-check.yml
javahippie: I think so, we don't need special permissions right now.

.github/zizmor/config.yml
javahippie: This is for exclusions, and I'm pretty certain that we will need them at some point, so I included the file from the beginning for documentation purposes

.github/workflows/zizmor-check.yml
javahippie: Ah, you mean that zizmor is only started if files in `.github` were changed? That's a good idea

kthoms: I have updated this

.github/workflows/zizmor-check.yml
kthoms: removed

kthoms:
Test executions:
- [x] [auto-merge.yml](https://github.com/operaton/operaton/actions/runs/32804929808)
- [x] [build.yml](https://github.com/operaton/operaton/actions/runs/32805123073)
- [x] [integration-build.yml](https://github.com/operaton/operaton/actions/runs/32804805805)
- [x] [migration-test.yml](https://github.com/operaton/operaton/actions/runs/32805307470)
- [x] [slack-link-check.yml](https://github.com/operaton/operaton/actions/runs/32805245259)

PR #3554, opened 2026-08-25
fix(webapps): Correct broken documentation links in neo web apps

antonAtGit:
## What

Fixes three documentation links in `webapps-neo` that returned HTTP 404:

- **TaskForm.jsx** and **StartProcessList.jsx** — the "How to migrate" link
  for legacy embedded forms was missing the `docs/` path segment.
- **Processes.jsx** — the "How to deploy" link on the empty processes page
  still pointed at the removed `/manual/latest/…` install path.

## Fix

- add the missing `docs/` segment →
  `https://docs.operaton.org/docs/documentation/user-guide/task-forms/`
- repoint the deploy link →
  `https://docs.operaton.org/docs/documentation/user-guide/process-engine/deployments`

All three corrected URLs return HTTP 200.

related to #3553


lambdaschmied2 (approved):
All good, thx!

Issue #3557, opened 2026-08-25
webapps-neo: browser locale "de" requests non-existent /locales/de (404)


antonAtGit:
When the browser reports a language-only locale (e.g. `de`), the neo web apps
request `/locales/de/translation.json`, which does not exist — only
region-specific files (`de-DE`, `en-US`, …) are shipped. The 404 response is then
parsed as JSON and logged as an error on startup:

```
i18next::backendConnector: loading namespace translation for language de failed
failed parsing /locales/de/translation.json to json
```

## Steps to reproduce
1. Set the browser/OS language to German (`de`), or open the app with `?lng=de`.
2. Open the dev-tools console on startup → JSON parse error for
   `/locales/de/translation.json`.

## Cause
i18next has no mapping from the detected language-only code to the region file,
and `load` defaults to also requesting the language-only (`de`) and `dev`
variants.

I'd like to work on this.

lambdaschmied2:
Fixed with PR #3558

PR #3558, opened 2026-08-25
fix(webapps): Map browser locale to region translation file in neo

antonAtGit:
## What

Fixes the startup i18n error in `webapps-neo` when the browser reports a
language-only locale (e.g. `de`): the app requested
`/locales/de/translation.json`, which 404'd and was logged as a JSON parse
error. Only region-specific files (`de-DE`, `en-US`, …) are shipped.

## Fix (`src/helper/i18n.js`)

- **`convertDetectedLanguage`** maps the detected code to the region file:
  `de`, `de-AT` → `de-DE`; unknown codes fall through to `fallbackLng`.
- **`load: 'currentOnly'`** stops i18next from also requesting the language-only
  (`de`) and `dev` variants.
- The language→region map is derived from a single `SUPPORTED_LNGS` list.

Verified in the browser: `?lng=de` now loads `de-DE` with no 404 and no
`rejecting language code` warnings.

related to #3557

### DIFF
--- a/webapps-neo/frontend/src/helper/i18n.js
+++ b/webapps-neo/frontend/src/helper/i18n.js
@@ -8,6 +8,13 @@ import LanguageDetector from 'i18next-browser-languagedetector';
 // have a look at the Quick start guide
 // for passing in lng and translations on init
 
+// Translation files are region-specific (see /public/locales), but browsers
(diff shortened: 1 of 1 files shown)


lambdaschmied2:
Works on my machine as well compared to the previous version.

PR #3591, opened 2026-09-02
fix(webapps): Fix password related texts (German)

### LINKED ISSUE #3573: Admin webapp: German USERS_MY_PASSWORD drops the {{name}} placeholder and misleads, plus a typo
Two issues in the German locale of the Admin webapp, found while setting up Operaton 2.1.4 with a German UI.

`webapps/frontend/public/app/admin/locales/de.json`

## 1. `USERS_MY_PASSWORD` drops the `{{name}}` placeholder — the label says the opposite of what it means

| | |
|---|---|
| en | `{{name}}'s Password*:` |
| de | `Aktuelles Passwort*:` |

The placeholder is filled with the **logged-in administrator's** name, not the name of the user being edited — `webapps/frontend/ui/admin/client/scripts/pages/userEdit.js`:

```js
$scope.currentUserPassword = $translate.instant(
  'USERS_MY_PASSWORD',
  {name: $scope.$root.userFullName},
);
```

Without the placeholder, "Aktuelles Passwort" (literally "current password") reads as *the current password of the account I am editing*. That is the natural reading, and it is wrong: this field is the admin's own password, asked for as confirmation. German even has a separate, correctly translated string for the self-service case — `USERS_OLD_PASSWORD` = "Altes Passwort".

Why this is worth fixing rather than cosmetic: an administrator resetting a forgotten password for someone else will look for a password they cannot have. On the same page sit **"Benutzer löschen"** (delete user) and **"Benutzer entsperren"** (unlock user). Reaching for the wrong one of those is a plausible next step.

Suggested: `Ihr Passwort zur Bestätigung ({{name}})*:` — or any wording that keeps `{{name}}` and makes clear whose password is meant.

## 2. `USERS_PASSWORD_INVALID` has a typo

| | |
|---|---|
| en | `Password must have at least 8 characters.` |
| de | `Passwort muss mindestens achte Zeichen besitzen.` |

"achte" should be "acht". The same message also exists as `SETUP_USER_PASSWORD_LENGTH` ("Das Passwort muss mindestens aus acht Zeichen bestehen."), which is correct — so the two could simply be aligned.

Suggested: `Das Passwort muss aus mindestens acht Zeichen bestehen.`

---

Both are still present on `main` as of today. Happy to open a PR if that is useful — just say so.

For context: the German locale is otherwise complete and worked out of the box once `locales` was enabled in `config.js` (2530 entries for Cockpit, 482 Tasklist, 645 Admin, 171 Welcome — exactly matching the English counts). Thanks for shipping it.

kthoms:
fixes #3573


kthoms:
Backported the change for release 2.1.5

Issue #3594, opened 2026-09-02
webapps-neo: flesh out the Engine Metrics page with five dashboard widgets


antonAtGit:
### Context

The bundled `metrics` plugin currently renders three values: the engine version, process starts (12 mo) and flow nodes executed (12 mo). It works and it demonstrates the plugin API nicely, but as the app's "Metrics" page it is close to empty — a new user opening it learns almost nothing about their engine.

The engine already exposes everything needed to make that page useful. This proposes filling it in, entirely inside the existing bundled plugin.

### Proposal

Five widgets, all reading existing REST endpoints, no engine-side changes:

| Widget | Endpoint |
|---|---|
| Running vs. completed instances (donut) | `/process-instance/count` + `/history/process-instance/count?finished=true` |
| Process starts per day, last 30 days (chart) | `/metrics?name=root-process-instance-start&interval=86400` |
| Top processes by running instances, with incident counts | `/process-definition/statistics?incidents=true` |
| Top user tasks by volume | `/history/task/report?reportType=count&groupBy=taskName` |
| Failed-jobs health tile | `/job/count?withException=true` |

Plus a layout pass so the tiles and the existing values read as one dashboard instead of a list.

The daily chart uses `root-process-instance-start` deliberately, so call activities don't inflate the numbers.

<img width="2557" height="819" alt="Image" src="https://github.com/user-attachments/assets/303ee5bd-b5e9-4ce9-820b-660e4e64e2c1" />

### Scope

The diff stays inside `src/plugins/bundled/metrics/` — `plugin.jsx`, `metrics.css`, `plugin.test.jsx`. **No core file is touched:** no routing, no header, no `config.js`, no API layer. Translations stay in the plugin's own `translations` block (`en-US`, `de-DE`), same as today.

Nothing here is new API surface. It extends the plugin through the extension point the plugin system already provides, which is arguably also the point: the reference bundled plugin should look like something worth referencing.

### Contribution

Implemented and tested against an engine under synthetic load (the `dev-fixtures` load generator), with unit tests for the widgets.

PR #3595, opened 2026-09-02
feat(webapps-neo): Flesh out the Engine Metrics page with five dashboard widgets

antonAtGit:
Fills in the Engine Metrics page as proposed in #3594.

The bundled `metrics` plugin rendered three values — engine version, process starts (12 mo), flow nodes executed (12 mo). This adds five widgets on top of them, all reading REST endpoints the engine already exposes.

| Widget | Endpoint |
|---|---|
| Running vs. completed instances (donut) | `/process-instance/count` + `/history/process-instance/count?finished=true` |
| Process starts per day, last 30 days (chart) | `/metrics?name=root-process-instance-start&interval=86400` |
| Top processes by running instances, with incidents | `/process-definition/statistics?incidents=true` |
| Top user tasks by volume | `/history/task/report?reportType=count&groupBy=taskName` |
| Failed-jobs health tile | `/job/count?withException=true` |

Plus a layout pass so the tiles read as one dashboard rather than a list.

### Scope

```
src/plugins/bundled/metrics/plugin.jsx        +365
src/plugins/bundled/metrics/metrics.css       +335  -6
src/plugins/bundled/metrics/plugin.test.jsx   +220
```

The diff stays inside the bundled plugin. **No core file is touched** — no routing, no header, no `config.js`, no API layer — so nothing outside the metrics page can be affected. Translations stay in the plugin's own `translations` block (`en-US`, `de-DE`), as before. No new dependency.

Eight commits, one per widget plus the layout pass and a fix, so this can be reviewed or cherry-picked piecewise.

### Details worth flagging

- The daily chart queries `root-process-instance-start`, not `activity-instance-start`, so call activities do not inflate the numbers. Sparse buckets (the engine returns only days with activity) are filled to a gap-free 30-day series.
- Every widget handles its own empty state — no running instances, no completed tasks, no starts in the window — so a fresh engine renders sensibly instead of showing empty boxes.
- The donut's centre label is sized from its digit count, so a five- or six-digit total stays clear of the ring.

### Testing

`plugin.test.jsx` covers the widgets, the series builder and the donut label sizing. Full frontend suite passes (595 tests), ESLint is clean on the plugin, `vite build` succeeds.

Verified manually against an engine under synthetic load from `webapps-neo/frontend/dev-fixtures` (~7k running instances across 19 process definitions, incidents present), in light and dark mode.

### Open question

These are proposed as the default page content. The widgets are independent, so making them individually toggleable is straightforward if you would rather not expand the default view — happy to follow whichever you prefer.

related to #3594

### DIFF
--- a/webapps-neo/frontend/src/plugins/bundled/metrics/metrics.css
+++ b/webapps-neo/frontend/src/plugins/bundled/metrics/metrics.css
@@ -174,6 +174,11 @@
     white-space: nowrap;
   }
 
+  #content.metrics-page .ranking-sub {
+    color: var(--color-text-secondary, #888);
+    font-size: 0.85em;
+  }
+
   #content.metrics-page .ranking-bar-track {
     block-size: 0.6rem;
     background: var(--background-3, #eee);
--- a/webapps-neo/frontend/src/plugins/bundled/metrics/metrics.css
+++ b/webapps-neo/frontend/src/plugins/bundled/metrics/metrics.css
@@ -6,6 +6,16 @@
  * its styles predictably win without specificity wars and never leak.
  */
 @layer custom {
+  /* `main` carries no padding of its own; give the page the same breathing room
+     the other pages get (Home uses `p-3`), so content isn't flush to the edges. */
+  #content.metrics-page {
+    padding: var(--spacing-4, 1.5rem);
+  }
+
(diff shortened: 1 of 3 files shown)


lambdaschmied2 (approved):
lgtm

Issue #3593, opened 2026-09-02
webapps-neo: configurable CI branding (logo, colours, favicon) via runtime config


antonAtGit:
### Motivation

Operators want their own logo and colours in the web apps — a UI that looks like the organisation's own tool gets noticeably better internal adoption. Today that requires forking and rebuilding the frontend. This is a visual overlay only, not a white-label rebrand: same UI, same behaviour.

### Proposal

An optional `branding` block in the existing runtime config (`config.json`, with `VITE_BRANDING` as the dev fallback). No rebuild, no new dependency, no new endpoint.

```json
{
  "branding": {
    "appName": "Example Corp Workflows",
    "logoUrl": "/branding/logo.svg",
    "logoAlt": "Example Corp",
    "faviconUrl": "/branding/favicon.svg",
    "colors":     { "primary": "#0d4f8b", "link": "#0d4f8b", "danger": "#b3261e", "warning": "#9a6700", "success": "#1a7f37" },
    "colorsDark": { "primary": "#7cb3e8", "link": "#7cb3e8" },
    "hue": 212
  }
}
```

Every field is optional — missing or invalid values fall back to the Operaton default, so a broken block can never break the UI.

### How it works

- **Colours** — the stylesheet already uses CSS custom properties (`--hue`, `--color-primary`, …). Branding injects one `:root` override at boot, plus a `prefers-color-scheme: dark` block from `colorsDark`.
- **Logo** — `Header.jsx` and the login screen read `logoUrl` / `logoAlt`.
- **Favicon and title** — swapped at boot.

Values are validated before reaching the DOM or CSS (hex colours, relative or `http(s)` URLs only), so a config file cannot inject markup or stylesheet rules.

### Why config, and not the plugin system

The web apps have a plugin system, and branding was considered as a plugin first. We decided against
                                                                               - **Timing** — branding hirst paint. Plugins areloaded after boot, from `pluginsUrl`, which would mean a visible flash of the  default logo and colours
- **Availability** — the branding must work on a plain installation with no    `pluginsUrl` configured ang the base look-and-feeldepend on an optional subsystem inverts that.                                  - **Nature of the extensid the app with views andbehaviour. Branding changes the shell's global CSS tokens, the favicon and the title; there is no view t
- **Data, not code** — a plugin ships executable JS. Branding is a handful of
validated strings and numer surface for anoperator-supplied artefact and can be checked before it is applied.

The plugin system stays the right place for anything that adds functionality —
[...]

PR #3596, opened 2026-09-02
feat(webapps-neo): Config-driven CI branding (logo, colours, favicon)

antonAtGit:
Implements the configurable CI branding proposed in #3593.

An optional `branding` block in the runtime configuration white-labels the app per installation — no rebuild, no new dependency, no new endpoint. It rides along the `config.json` that is already fetched at boot, with `VITE_BRANDING` as the dev fallback.

```json
{
  "branding": {
    "appName": "Example Corp Workflows",
    "logoUrl": "/branding/logo.svg",
    "logoAlt": "Example Corp",
    "faviconUrl": "/branding/favicon.svg",
    "colors":     { "primary": "#0d4f8b", "link": "#0d4f8b", "danger": "#b3261e", "warning": "#9a6700", "success": "#1a7f37" },
    "colorsDark": { "primary": "#7cb3e8", "link": "#7cb3e8" },
    "hue": 212
  }
}
```

Every field is optional. A missing, empty or malformed value falls back to the current Operaton default, so a broken branding block renders exactly like today rather than breaking the UI.

### Scope

```
src/branding.js               +97   validation, CSS override, favicon and title
src/branding.test.js          +98
src/config.js                 +18   branding in from_env and from_document
src/index.jsx                  +7   apply at boot, login logo
src/components/Header.jsx      +7   header and mobile logo
src/css/style.css              +1
```

### How it works

- **Colours** — the stylesheet is already built on CSS custom properties (`--hue`, `--color-primary`, `--color-link`, …). Branding injects one `:root` override stylesheet at boot, plus a `@media (prefers-color-scheme: dark)` block from `colorsDark`, so light and dark stay separately tunable.
- **Logo** — `Header.jsx` and the login screen read `logoUrl` / `logoAlt`; the `aria-label` follows `appName` rather than staying hardcoded.
- **Favicon and title** — swapped at boot from `faviconUrl` / `appName`.
- The one-line `style.css` change moves the mobile logo's capitalisation from the markup (`OPERATON`) into `text-transform`, so a configured name keeps the same look without being shouted in the DOM.

### Validation

Values are checked before they reach the DOM or CSS: colours must be hex, URLs must be a relative path or `http(s)` — which rules out `javascript:` and `data:` — numbers must be finite, strings non-empty. Only valid declarations are emitted, so a config file cannot inject markup or arbitrary stylesheet rules.

### Why config rather than a plugin

The web apps have a plugin system, and this was considered as a plugin first. Branding has to be in place before first paint, and it has to work on an installation with no `pluginsUrl` configured; a plugin loads after boot, which would mean a flash of the default logo and colours on every page load. It also changes global CSS tokens, the favicon and the title rather than adding a view, and it is validated data instead of executable JavaScript. The reasoning is spelled out in #3593.

### Testing

`branding.test.js` covers the validation rules, the generated CSS for light and dark, and the fallbacks for missing and malformed blocks. Full frontend suite passes (598 tests), `vite build` succeeds, ESLint reports nothing new (the one remaining warning in `index.jsx` predates this branch).

Verified manually against real customer corporate identities, in light and dark mode.

related to #3593

### DIFF
--- a/webapps-neo/frontend/.env.development
+++ b/webapps-neo/frontend/.env.development
@@ -7,3 +7,9 @@ VITE_AUTH_MODE=basic # oauth | basic
 VITE_OAUTH_AUTHORITY=http://localhost:8888/realms/operaton
 VITE_OAUTH_CLIENT_ID=operaton-web-apps
 VITE_OAUTH_REDIRECT_URI=http://127.0.0.1:5173/
+
+# CI branding (logo, app name, favicon, colour tokens; see src/branding.js).
+# Quote the value. The hex colours contain "#", which dotenv otherwise reads as
+# the start of a comment and silently truncates the JSON — the block then fails
+# to parse and everything falls back to the Operaton defaults, with no error.
+#VITE_BRANDING='{"appName":"Acme BPM","logoUrl":"/branding/logo.svg","logoAlt":"Acme BPM","faviconUrl":"/branding/favicon.svg","colors":{"primary":"#0d4f8b","link":"#0d4f8b"},"colorsDark":{"primary":"#7cb3e8","link":"#7cb3e8"},"hue":212}'
--- a/webapps-neo/frontend/src/branding.js
+++ b/webapps-neo/frontend/src/branding.js
@@ -1,0 +1,97 @@
+/**
+ * branding.js
+ *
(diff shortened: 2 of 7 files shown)


lambdaschmied2:
Tested, worked. Thx!

PR #3644, opened 2026-09-09
fix(webapps-neo): Migrate every matching instance when none is selected

### LINKED ISSUE #3628: webapps-neo: Migration with no instances selected fails instead of migrating all matching instances
The migration page tells the user:

_Leave all unchecked to migrate every matching instance._

Doing exactly that fails, and nothing is migrated.

**Cause**

With no checkbox ticked and no filter set, the frontend sends both selectors as null:

// webapps-neo/frontend/src/pages/Migrations.jsx:227-231
selected_ids.length > 0 ? selected_ids : null,   // → null
has_query ? query_obj : null,                    // → null when no filter is set

The request body becomes {"processInstanceIds": null, "processInstanceQuery": null} (src/api/resources/migration.js:39-53). The engine collects an empty set and rejects it:

// engine/…/impl/migration/MigrateProcessInstanceCmd.java:97
ensureNotEmpty(BadUserRequestException.class,
    "Process instance ids cannot empty", "process instance ids", processInstanceIds);

The engine is right — it was given nothing. The hint in the UI is what's wrong: "nothing selected" is never translated into "all matching".

**Steps to reproduce**

1. Deploy a process definition in two versions, with at least one running instance of the old one.
2. Migrations → pick source and target → generate the mapping.
3. In "select instances", tick nothing and set no filter — what the hint suggests.
4. Execute.

**Expected**: all matching instances of the source version are migrated.
Actual: BadUserRequestException: Process instance ids cannot empty, nothing is migrated.

_Seen on 2.2.0-M3_

**Suggested fix**

Send a query instead of null when nothing is selected. execute already stamps the source definition onto every query:

// src/api/resources/migration.js:44-50
processInstanceQuery: { ...process_instance_query,
                        processDefinitionId: migration_plan.sourceProcessDefinitionId }

An empty query therefore means exactly "every instance of the source version". The rule would be: instances ticked → send the ids, query null; otherwise → send the query, even when empty.

Worth deciding alongside: migrating all instances is not reversible, so showing the affected count and asking for confirmation seems appropriate, and running it as a batch (executeAsync) is likely the normal case rather than the exception.

antonAtGit:
Fixes #3628.

## The defect

The page promises:

> Leave all unchecked to migrate every matching instance.

Doing exactly that sent both selectors as `null`:

```js
// src/pages/Migrations.jsx (before)
selected_ids.length > 0 ? selected_ids : null,   // → null
has_query ? query_obj : null,                    // → null when no filter is set
```

The engine collected an empty set from `{"processInstanceIds": null, "processInstanceQuery": null}`
and rejected the request:

```java
// engine/…/impl/migration/MigrateProcessInstanceCmd.java:97
ensureNotEmpty(BadUserRequestException.class,
    "Process instance ids cannot empty", "process instance ids", processInstanceIds);
```

The engine is right — it was given nothing to migrate. What was wrong is that the UI
never translated "nothing selected" into "every matching instance".

## The change

An empty query is exactly the set the hint describes, because `execute` already stamps
the source definition onto whatever query it is given:

```js
// src/api/resources/migration.js
processInstanceQuery: { ...process_instance_query,
                        processDefinitionId: migration_plan.sourceProcessDefinitionId }
```

So the query is now sent when nothing is ticked, and the promise holds.

**Nothing else changes.** The query builder stays additive to the checkboxes — the engine
unions ids and query, and a query that was actually built is still sent alongside a
selection, exactly as before. Only the combination that could not work is affected:

| ticked | query built | ids sent | query sent | changed |
| --- | --- | --- | --- | --- |
| yes | no | yes | no | no |
| yes | yes | yes | yes | no |
| no | yes | no | yes | no |
| no | no | no | **yes (empty)** | **yes** |

## Where the decision lives

It moved out of the submit handler into `migration_selectors` in `migration_helpers.js`.
It sat inline in the JSX where no test could reach it, which is presumably why the case
went unnoticed. Four tests now cover the table above.

## Worth deciding separately

Migrating *every* matching instance is not reversible, and on a busy definition it can be
a lot of them. Showing the affected count and asking for confirmation would fit here, and
running it as a batch (`executeAsync`) is arguably the normal case rather than the
exception. Both felt like product decisions rather than part of this fix, so they are not
in this PR — happy to follow up if you want them.

I have not checked whether `validate` has the same gap when nothing is selected.

## Testing

Full suite green (602), ESLint clean, Prettier clean on the touched files. Verified by
hand against a local engine: five instances on v1 of a definition, nothing ticked, no
filter — before the change the migration was rejected and nothing moved, after it all five
were migrated to v2. Ticking a subset still migrates only that subset.

### DIFF
--- a/webapps-neo/frontend/src/pages/Migrations.jsx
+++ b/webapps-neo/frontend/src/pages/Migrations.jsx
@@ -21,6 +21,7 @@ import {
   remove_query_row,
   update_query_row,
   build_query,
+  migration_selectors,
   add_query_params_abstract,
 } from "./migration_helpers.js";
 
@@ -225,12 +226,12 @@ const ProcessSelection = () => {
             query_obj = build_query(
               migration_state.process_instance_query.peek(),
             ),
-            has_query = Object.keys(query_obj).length > 0;
+            selectors = migration_selectors(selected_ids, query_obj);
           engine_rest.migration.execute(
             state,
             migration_plan,
-            selected_ids.length > 0 ? selected_ids : null,
-            has_query ? query_obj : null,
+            selectors.process_instance_ids,
+            selectors.process_instance_query,
             execute_form_data.peek().skip_custom_listeners,
             execute_form_data.peek().skip_io_mappings,
             execute_form_data.peek().async,
--- a/webapps-neo/frontend/src/pages/Migrations.test.js
+++ b/webapps-neo/frontend/src/pages/Migrations.test.js
@@ -8,6 +8,7 @@ import {
   update_variable,
   update_mapping,
   add_query_params_abstract,
+  migration_selectors,
 } from "./migration_helpers.js";
 
 // --- Tests ---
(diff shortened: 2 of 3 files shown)


lambdaschmied2 (approved):
lgtm

PR #3648, opened 2026-09-09
feat(webapps-neo): Assign a task to another user

### LINKED ISSUE #3629: webapps-neo: a task cannot be assigned to another user
There is no way to hand a task over to someone else. The dialog behind the assignee field offers three buttons, and all three act on the logged-in user:

// webapps-neo/frontend/src/pages/Tasks.jsx:905-935 (shortened)
<dialog id="set_assignee">
  {assignee_is_different && !assigned ? (
    <button onClick={() => engine_rest.task.assign_task(state, null, task.id)}>Reset assignee</button>
  ) : (user_is_assignee || claimed) && !unclaimed ? (
    <button onClick={() => engine_rest.task.unclaim_task(state, task.id)}>Unclaim</button>
  ) : (
    <button onClick={() => engine_rest.task.claim_task(state, task.id)}>Claim</button>
  )}
</dialog>

There is no input for a user id. Claim and unclaim always act on yourself; reset clears the field.

The API binding already exists — it is just never called with a user:

// webapps-neo/frontend/src/api/resources/task.js:91
const assign_task = (state, assignee, task_id) =>
  POST(`/task/${task_id}/assignee`, { userId: assignee }, state, state.api.task.assign_result);

It is exported, the assign_result signal exists (src/state.js:154), and the only call site in the frontend passes null. What's missing is the input, not the plumbing.

**Regression against the previous webapps**

The old Tasklist had an editable assignee field:

<span cam-widget-inline-field
      change="assign(varValue)"
      validate="isInvalidUser(this)"
      value="assignee.id">

You typed a user id, it was validated against the identity service, and the task was assigned to that person — alongside claim, unclaim and reset.

**Steps to reproduce**

1. Log in as a member of a candidate group that has an open task.
2. Open the task, click the Assignee field.

Expected: enter another user and hand the task over.
Actual: only claim, unclaim or reset — no input field.

_Seen on 2.2.0-M3_

**Suggested fix**

Add an input for the user id to the dialog and call assign_task with it. Worth deciding alongside:

- Free text with validation (as the old Tasklist did) or completion via GET /user?firstNameLike=…. A plain dropdown does not scale past a few users.
- Whether to restrict the choice to the task's candidate group. POST /task/{id}/assignee accepts any user id, so this is a product decision, not a constraint.
- Whether to offer delegation (POST /task/{id}/delegate) as well, which Camunda 7 had.
- Assigning requires UPDATE on the task; without it the button should not appear rather than fail on click.

antonAtGit:
Fixes #3629.

## The defect

The assignee dialog offered claim, unclaim and reset — all three act on the signed-in
user, so a task could not be handed to a colleague. In a support rota that is the normal
case: holidays, sick leave, shift changes, escalation to a named person. Without it the
only route was "the assignee unclaims, the target claims", which needs both people at the
keyboard at once and lets anyone else take the task in between.

The binding was already there — `assign_task` takes an assignee and `assign_result` exists
— the single call site just passed `null`. This adds the field that was missing.

## Validating the id

The id is checked before the task is handed over, against the same endpoint the previous
Tasklist used:

```
GET /user?id=<id>&maxResults=1
```

The engine cannot do this for us. It answers `204` to any string:

```
POST /task/{id}/assignee  {"userId":"no-such-user"}  →  204
assignee afterwards: 'no-such-user'
```

A typo therefore succeeds and leaves the task held by an account that does not exist: it
drops out of every list and only "reset assignee" brings it back.

An id the lookup does not find is refused with a message that names **both** reasons the
lookup can come back empty — no such user, or the caller may not see them. Identity queries
are filtered by `READ` on `User`, and the UI cannot tell the two apart, so a message
saying only "no such user" would be wrong wherever the directory is closed. The message
clears as soon as the field is edited again.

This carries over the previous Tasklist's behaviour together with its precondition:
installations whose users lack `READ` on `User` will have to grant it for assignment to
work, exactly as before.

## Feedback after every action

Every action in the dialog now re-reads the task and closes, so the card and the buttons
show the new state. They did not before. Three flags were meant to reflect the action just
taken:

```js
claim_result = state.api.task.claim_result.value?.data
claimed      = claim_result?.status === RESPONSE_STATE.SUCCESS
```

The signal carries the status beside `data`, not inside it — and since the engine answers
these calls with `204`, the helper resolves the body to the string `"No Content"`, so the
comparison was `"No Content".status === "success"`. All three were always false, and
nothing re-read the task either, so clicking claim or reset changed nothing on screen and
looked as though it had failed. With the task re-read the flags have nothing left to do
and are gone; `task.assignee` carries the truth.

## Layout

The dialog follows the shape the other dialogs in this file already use — heading, then
label/input pairs in the global two-column form grid, then the submit in a
`.button-group`. The heading sits **outside** the form on purpose: the global `form` rule
is that grid, so a heading placed inside becomes a grid item and lands beside the field.
The existing claim button moved into a `.button-group` of its own for the same reason, so
the two blocks line up instead of it sitting loose beside "close".

## Testing

Full suite green (604), ESLint clean. Nine tests cover the dialog, including: assigning to
a valid id, refusing an unknown one without calling `assign_task`, the message clearing on
edit, the task being re-read after an action, and *not* being re-read when the action
failed.

Walked through against a running engine:

| input | result |
| --- | --- |
| `no-such-user` | message at the field, no assignment, dialog stays open |
| edit the field | message disappears |
| `bob` | assigned, card shows `bob`, dialog closes |
| claim / reset | card updates, dialog closes, buttons reflect the new state |


lambdaschmied2:
@antonAtGit we got merge conflicts. Can you update this branch - the rest looks fine, and I would approve.

PR #3649, opened 2026-09-09
feat(webapps-neo): Show only the pages the user is authorized for

### LINKED ISSUE #3630: webapps-neo: Navigation ignores the user's authorizations — every menu entry is shown to everyone
**The primary navigation is a fixed list with no reference to the logged-in user's permissions:**

// webapps-neo/frontend/src/components/Header.jsx:22-30
const builtin_nav = [
  { href: "/tasks",       nameKey: "nav.tasks",       hotkey: "alt+shift+1" },
  { href: "/processes",   nameKey: "nav.processes",   hotkey: "alt+shift+2" },
  { href: "/decisions",   nameKey: "nav.decisions",   hotkey: "alt+shift+3" },
  { href: "/deployments", nameKey: "nav.deployments", hotkey: "alt+shift+4" },
  { href: "/batches",     nameKey: "nav.batches",     hotkey: "alt+shift+5" },
  { href: "/migrations",  nameKey: "nav.migrations",  hotkey: "alt+shift+6" },
  { href: "/admin",       nameKey: "nav.admin",       hotkey: "alt+shift+7" },
];

Every authenticated user is offered all seven entries, Admin included, regardless of their APPLICATION authorizations. Searching the frontend for authorizedApps, isAuthorized or any permission check in the navigation returns nothing.

The information is already there and simply unused — the login response carries it:

POST /api/admin/auth/user/default/login/neo
{"userId":"melder-bw","authorizedApps":["neo","tasklist","welcome"]}
{"userId":"agent",    "authorizedApps":["neo","tasklist","welcome","cockpit"]}

This is not a data leak. The engine still filters every query, so an unauthorized user reaching a page sees an empty list. It is a usability and trust problem: people are offered areas that are closed to them and land in empty screens with no explanation.

**Steps to reproduce**

1. Create a user without cockpit and admin access:

APPLICATION ACCESS → tasklist   (GRANT)
no ACCESS on cockpit, no ACCESS on admin
2. Set operaton.bpm.authorization.enabled: true and log in as that user.
3. The login response contains authorizedApps: ["neo","tasklist","welcome"].

Expected: the navigation shows Tasks.
Actual: it shows Tasks, Processes, Decisions, Deployments, Batches, Migrations and Admin. All are reachable; the pages behind them are empty.

_Seen on 2.2.0-M3_

**Suggested fix**

Filter the navigation against authorizedApps from the login response. The old webapps knew three applications, Neo has finer-grained pages, so a mapping is needed:

| Neo page | application |
| --- | --- |
| `/tasks` | `tasklist` |
| `/processes`, `/decisions`, `/deployments`, `/batches`, `/migrations` | `cockpit` |
| `/admin` | `admin` |

**_The following questions should be addressed and implemented separately in an enhancement ticket:_**

- Is application-level granularity sufficient? It is too coarse—access to the Cockpit also includes batches and migrations. Checks per resource (authorization/check?permissionName=READ&resourceName=…) would be more granular, but would require one request per page during login.
- Simply hiding the entry is not enough: When /admin is entered directly, a clear message should be displayed instead of a blank page; therefore, a route guard is needed here.
- When “authorization.enabled” = false, the server reports all applications, so navigation remains complete as before.
- A user who has access to nothing but the welcome page would see an empty navigation bar—it would be better to have a page that clearly explains this.
- `nav_entries()` adds every entry from the PAGE plugin to the built-in entries. Whether and how plugins can tie their visibility to a permission should be clarified at the same time, otherwise the same vulnerability will reoccur.

antonAtGit:
Fixes #3630.

## The defect

The primary navigation was a fixed list, so every signed-in user was offered all seven
entries, Admin included, whatever their `APPLICATION` authorizations. The pages behind them
are not a data leak — the engine filters every query — but people were sent into areas
closed to them and landed on empty screens with nothing explaining why.

The information was already there and unused: the sign-in response carries
`authorizedApps`.

## The change

`authorizedApps` is kept in state, restored on reload — the session survives a refresh, so
the navigation must not open up again — and cleared on sign-out.

Neo has finer-grained pages than the three applications the engine authorizes on, so
`helper/authorized_apps.js` holds that mapping in one place:

| page | application |
| --- | --- |
| `/tasks` | `tasklist` |
| `/processes`, `/decisions`, `/deployments`, `/batches`, `/migrations` | `cockpit` |
| `/admin` | `admin` |

Pages belonging to no application — `/`, `/account`, `/help` — stay available.

**Hiding an entry is not enough on its own**, since the address bar, a bookmark and the
keyboard shortcuts all reach a page directly. The routes are wrapped so an unauthorized
page says which application is missing instead of rendering empty. That wrapper is also why
the shortcuts are left alone: they route as before and meet the same answer, so there is
one place where this is decided.

When the server reports nothing — a backend configured elsewhere has no session to ask —
nothing is hidden. Hiding on a guess would be worse than showing, and with engine
authorization switched off the server reports every application anyway, so that needs no
separate case.

## Testing

Ten tests: the mapping and its fallbacks, the navigation filtered per user, and the guard
rendering the page or the message.

Walked through against a running engine with three users:

| user | `authorizedApps` | navigation |
| --- | --- | --- |
| a reporter | `neo, tasklist, welcome` | `/tasks` |
| a support agent | `+ cockpit` | `/tasks`, `/processes`, `/decisions`, `/deployments`, `/batches`, `/migrations` |
| admin | `+ admin` | all of them |

Typing a closed route directly: `/admin` as the agent and `/processes` as the reporter both
answer "No access to this area — your account is not authorized for the … application",
while `/tasks` opens normally for both.

Full suite green (610), ESLint clean.

### DIFF
--- a/webapps-neo/frontend/public/locales/de-DE/translation.json
+++ b/webapps-neo/frontend/public/locales/de-DE/translation.json
@@ -847,6 +847,11 @@
     "goto-hint-prefix": "Mit der \"Gehe zu\"-Funktion oben rechts oder durch Drücken von",
     "goto-hint-suffix": "können Sie schnell zu einer Ressource navigieren, wenn Sie deren ID kennen."
   },
+  "no-access": {
+    "title": "Kein Zugriff auf diesen Bereich",
+    "text": "Ihr Konto ist für die Anwendung {{app}} nicht freigeschaltet.",
+    "hint": "Ein Administrator muss Ihrer Gruppe den Zugriff darauf gewähren."
+  },
   "bpmn": {
     "zoom-in": "Vergrößern",
     "zoom-out": "Verkleinern",
--- a/webapps-neo/frontend/public/locales/en-US/translation.json
+++ b/webapps-neo/frontend/public/locales/en-US/translation.json
@@ -847,6 +847,11 @@
     "goto-hint-prefix": "Using the \"Go To\" function in the top right or pressing",
     "goto-hint-suffix": "you can quickly navigate to a resource if you know its ID."
   },
+  "no-access": {
+    "title": "No access to this area",
+    "text": "Your account is not authorized for the {{app}} application.",
+    "hint": "An administrator has to grant your group access to it."
+  },
   "bpmn": {
     "zoom-in": "Zoom in",
     "zoom-out": "Zoom out",
--- a/webapps-neo/frontend/src/api/resources/auth.js
+++ b/webapps-neo/frontend/src/api/resources/auth.js
@@ -50,6 +50,7 @@ const start_session = (state, username, password) => {
       // Once the session exists the password has no further use here, so make
       // sure nothing is left holding one.
       state.auth.credentials.value = { username: null, password: null };
+      state.auth.authorized_apps.value = result.authorizedApps ?? null;
       return result.userId ?? username;
     });
 };
@@ -61,7 +62,13 @@ const session_user = (state) => {
 
   return fetch(_url_auth(), { headers, credentials: "include" })
     .then((response) => (response.ok ? response.json() : null))
-    .then((result) => result?.userId ?? null)
+    .then((result) => {
+      // A reload restores the session from the cookie, so the applications have
+      // to be restored with it — otherwise the navigation would open up again
+      // after every refresh.
+      state.auth.authorized_apps.value = result?.authorizedApps ?? null;
+      return result?.userId ?? null;
+    })
     .catch(() => null);
 };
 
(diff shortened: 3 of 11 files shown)


javahippie:
Ran it locally, all green. Looks right to me, three things first:

The dashboard isn't guarded, so a tasklist-only user still lands on four cards and gets "No access" when they click one. Go To has the same problem, it hard-codes the page list a second time. Related: page_allowed is an exact match, so /admin/users slips through. Worth making it a prefix match now while there's only the one caller.

The auth.js changes have no assertions. You could delete the line in session_user and the suite would stay green. auth.test.js:213 is the one to extend.

PR #3660, opened 2026-09-10
fix(engine-rest): Let a task be sent back as it was received

### LINKED ISSUE #3659: webapp-neo & engine: Setting a due date fails for any task that uses a Operaton Form
PUT /task/{id} rejects the body that GET /task/{id} just produced:

400 Cannot construct instance of `org.operaton.bpm.engine.form.OperatonFormRef`
    (no Creators, like default constructor, exist)

Leaving the property out is no way around it, because the endpoint replaces the task rather than patching it:

500 ENGINE-03072 Cannot change tenantId of Task '…'.
    Current tenant id 'x', Tenant id to set 'null'

So a client has to send the whole task, and the whole task is refused. Setting a due date or a follow-up date is impossible.

**Which tasks are affected**

Only those whose operatonFormRef is non-null — tasks referencing an Operaton Form via camunda:formRef. With camunda:formKey the property is null and deserialises fine, which is why this has gone unnoticed: formKey is what most existing models use. It affects everyone following the path the new web apps require, since they render formRef forms and no longer render embedded HTML forms.

Measured against a running engine, same task, same body except one property:

| `operatonFormRef` in the body                        | response |
| ---------------------------------------------------- | -------- |
| `{"key":"approval","binding":"deployment",…}`         | 400      |
| `null`                                                | 204      |
| absent                                                | 204      |

**Steps to reproduce**

curl -su user:pw "$ENGINE/task/$TID" -o t.json           `→ task uses camunda:formRef`
curl -su user:pw -X PUT "$ENGINE/task/$TID" \
     -H 'Content-Type: application/json' -d @t.json
 `→ 400 Cannot construct instance of OperatonFormRef`

_Seen on 2.2.0-M3_

**Cause**

TaskDto.operatonFormRef is typed as an interface, so Jackson tries to instantiate it on input. It never needed to: the property is output-only in fact — TaskDto.updateTask(Task) copies eleven fields and not this one, and the engine's Task interface has only getOperatonFormRef(), no setter. The value comes from the process model.

**Suggested fix**

Mark the property read-only for Jackson, so what a client sends for it is dropped rather than failing the request. It is still reported on the way out.

Worth deciding alongside: the OpenAPI templates use TaskDto as the request body for PUT /task/{id} and POST /task/create and list this property there, so the documentation still presents it as writable. Expressing that would mean adding a readOnly flag to the lib.property macro, which no model currently uses.

**I have a fix ready and would like to take this one — please assign it to me.**

antonAtGit:
Fixes #3659.

## The defect

`PUT /task/{id}` rejects the body `GET /task/{id}` just produced:

```
400 Cannot construct instance of `org.operaton.bpm.engine.form.OperatonFormRef`
    (no Creators, like default constructor, exist)
```

Leaving the property out is no way around it, because the endpoint replaces the task rather
than patching it — drop `tenantId` along with it and the answer is

```
500 ENGINE-03072 Cannot change tenantId of Task '…'
```

So the whole task has to be sent, and the whole task is refused. Setting a due date or a
follow-up date is impossible for any task referencing an Operaton Form.

## Two commits

**`engine-rest`** — `TaskDto.operatonFormRef` is typed as an interface, so Jackson tried to
instantiate it on input. It never needed to: the property is output-only in fact.
`TaskDto.updateTask(Task)` copies eleven fields and not this one, and the engine's `Task`
interface has only `getOperatonFormRef()`, no setter — the value comes from the process
model. It is now marked read-only, so what a client sends for it is dropped instead of
failing the request. It is still reported on the way out.

**`webapps-neo`** — the failure was invisible, which is why it took a while to find.
`update_task` wrote the answer of the PUT into the signal holding the loaded task and
re-read the task regardless of the outcome, so the error was replaced by a fresh copy of
the unchanged task before anything could render it; both date dialogs closed
unconditionally. The PUT now has its own signal, the re-read happens only on success, the
result reaches the caller, and the dialogs close only when the change went through. The
timestamp was also built with a single-digit second — accepted by the engine, but not the
format it claims to be.

The second commit stands on its own: it makes any rejected change visible, whatever the
cause.

## Testing

`testPutUpdateTaskWithFormRef` sends a body carrying the reference and expects 204;
195 tests in `TaskRestServiceInteractionTest` green. 612 frontend tests green, three of them
new around `update_task`.

## Worth deciding separately

The OpenAPI templates use `TaskDto` as the request body for `PUT /task/{id}` and
`POST /task/create` and list this property there, so the documentation still presents it as
writable — sending it is now harmless rather than fatal, but it still does nothing. Saying
so properly means `"readOnly": true` in the schema, and the `lib.property` macro has no such
flag; no model in the repository uses one. That felt like your call rather than something to
fold into this fix.

### DIFF
--- a/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/dto/task/TaskDto.java
+++ b/engine-rest/engine-rest/src/main/java/org/operaton/bpm/engine/rest/dto/task/TaskDto.java
@@ -18,6 +18,8 @@ package org.operaton.bpm.engine.rest.dto.task;
 
 import java.util.Date;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import org.operaton.bpm.engine.BadUserRequestException;
 import org.operaton.bpm.engine.form.OperatonFormRef;
 import org.operaton.bpm.engine.rest.dto.converter.DelegationStateConverter;
@@ -47,6 +49,8 @@ public class TaskDto {
   private String caseDefinitionId;
   private boolean suspended;
   private String formKey;
+  /** Derived from the process model; {@link #updateTask(Task)} never reads it. */
+  @JsonProperty(access = JsonProperty.Access.READ_ONLY)
   private OperatonFormRef operatonFormRef;
   private String tenantId;
   /**
--- a/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/TaskRestServiceInteractionTest.java
+++ b/engine-rest/engine-rest/src/test/java/org/operaton/bpm/engine/rest/TaskRestServiceInteractionTest.java
@@ -3355,6 +3355,27 @@ public class TaskRestServiceInteractionTest extends
   }
 
   @Test
+  void testPutUpdateTaskWithFormRef() {
+    // PUT replaces the task, so a client sends back what GET returned — form reference included.
+    Map<String, Object> json = new HashMap<>();
+    json.put("name", "A Task");
+    json.put("operatonFormRef", Map.of("key", "aFormKey", "binding", "deployment"));
+
+    given()
+        .pathParam("id", EXAMPLE_TASK_ID)
+        .body(json)
+        .contentType(ContentType.JSON)
+        .header("accept", MediaType.APPLICATION_JSON)
+    .expect()
+        .statusCode(Status.NO_CONTENT.getStatusCode())
+    .when()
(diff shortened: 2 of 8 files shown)


kthoms (commented):
Looks good from my side.
@lambdaschmied2 Please check also.

lambdaschmied2 (approved):
Looks good on the front-end side as well.

PR #3708, opened 2026-09-18
fix(release): disable JReleaser issue-closing to avoid crash

kthoms:
## Summary
The 2.2.0-RC1 dry-run crashed in "Release with JReleaser":
```
[ERROR] JReleaser failed after 0.094 s
java.lang.NumberFormatException: For input string: "32805123073"
```

JReleaser's `release.github.issues` feature (auto-comments/closes issues referenced in commit messages, e.g. `closes #123`) scans commit messages for bare `#<digits>` references. Commit e2645b1df7's message contains `run #32805123073` (a GitHub Actions run ID quoted for context, not an issue reference), which overflows JReleaser's int-based issue-number parser and aborts the whole release instead of skipping the unparseable match.

That commit is permanent history on `main`, so every future release from `main` would hit this identically.

## Fix
Set `release.github.issues.enabled: false`. This turns off the auto-comment/auto-close-on-issue feature; it does not affect the changelog generation or artifact publishing.

## Evidence
Dry-run: https://github.com/operaton/operaton/actions/runs/35334347382

## Test plan
- [ ] Re-run `release.yml` with `dry_run=true` on `main` (qualifier `RC1`) and confirm `Release with JReleaser` completes.
- [ ] Consider filing an upstream JReleaser issue for the crash-on-unparseable-issue-number behavior (should skip/warn, not abort).


kthoms:
Filed upstream: https://github.com/jreleaser/jreleaser/issues/2172

Confirmed by reading the JReleaser v1.26.0 source that this workaround is correct — both `extractIssues` call sites (`ChangelogProvider.resolveChangelog` and `ChangelogGenerator.createChangelog:115`) are gated on `release.github.issues.enabled`, so disabling it avoids the crash entirely. We can re-enable once JReleaser guards the `Integer.valueOf` parse.

